PR #189, opened 2023-11-14
Content Security Policy (CSP)

apepper:
I'm not too happy with `devServerCspHeader`, but haven't yet found a better practical way to do this.


vite.config.ts
```diff
@@ -68,3 +73,18 @@ export default defineConfig(({ mode }) => {
     },
   }
 })
+
+function devServerCspHeader() {
+  const directives = JSON.parse(JSON.stringify(headersCsp))
```
dcsaszar, line 110: * [ ] How about https://developer.mozilla.org/en-US/docs/Web/API/structuredClone?

dcsaszar (commented):
Looks good from a functional point of view. There is some clarification needed, see comments.

From a readability point of view, there's a lot of room for improvement. One bigger issue is that the different parts are mostly separated by commits (which I was able to follow after reading the history twice and doing my own research in between).
If a reader looks at the final outcome, it's hard to understand which part is doing what and why, and how they are supposed to connect. Ideas:
* We could inline `src/_headersCsp` in the `vite.config.js`. This way we have less indirection and it becomes clearer this is *not* part of the app source.
* If the headers were duplicated and written plainly in `_headers` and `vite.config.js`, it would be super easy to understand, but a bit harder to maintain consistency.
* We could make `_headers` the source of truth for `vite.config.js`, but I have no idea how much infrastructure (i.e. overhead) this requires. Unfortunately this approach would no longer work as soon as we need dynamic hashes.

apepper:
REVIEW ([95f65ba](https://github.com/Scrivito/scrivito-portal-app/pull/189/commits/95f65ba105285ad4044c67929b04e410e1fe29dc))

PR #197, opened 2023-11-27
Add a note, that bs-primary and bs-secondary are also used by packages

apepper:


aea-JR (approved):
Great, thank you.

agessler (approved):
I think this what you will expect. All the CSS is public - it's your portal app.
We use Bootstrap Markup and CSS. If packages will not use this it would be confusing.

I think every Scrivito Website out there is Bootstrap flavoured. So if you use a different Framework have to to care for everything.

When a package uses a BS variable it would be the right way. The only dependency is that you use at least BS v5.0 in your app.

PR #215, opened 2023-12-22
Weekly Dependabot ignores major updates

dcsaszar:
FYI If I understand the docs correctly we could also configure Dependabot to inform us about major updates independently from the weekly check by configuring [`groups.update-types`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) instead of ignoring them globally.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,5 +5,8 @@ updates:
     groups:
       update:
         patterns: ['*']
+    ignore:
+      - dependency-name: '*'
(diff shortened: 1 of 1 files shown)


apepper (approved):
LGTM ([4b52ec7](https://github.com/Scrivito/scrivito-portal-app/pull/215/commits/4b52ec7e80439090c6f65bc8e157219107f9dcea))

Looks promising. Thanks for looking this up! Let's give it a try.

PR #218, opened 2024-01-02
Add content dump script

dcsaszar:


src/dumpContent.ts
```diff
@@ -0,0 +1,112 @@
...
+    `https://api.scrivito.com/tenants/${INSTANCE_ID}/${apiPath}`,
+    { body, headers, method: body ? 'put' : 'get' },
+  )
+  if (response.status === 200) return response.json()
+
+  console.log(`\nHTTP status ${response.status}, retrying...\n`)
+  const sleep = promisify(setTimeout)
+  await sleep(2000)
+  return fetchJson(apiPath, body)
+}
```
apepper, line 112: I would prefer, if we could use `unstable_JrRestApi` from scrivito with `configure(tenantId, clientId, clientSecret)` instead of a custom implementation using "old school" `SCRIVITO_API_KEY`.

Is there something, that speaks against this suggestion?

src/dumpContent.ts
```diff
@@ -0,0 +1,112 @@
...
+import { promisify } from 'util'
+
+type ObjData = { _id: string }
+type SearchData = { continuation?: string; objs: ObjData[] }
+type BlobsData = { private_access: { get: { url: string } } }
+
+const DIRECTORY = 'contentDump'
+
+const API_KEY = process.env.SCRIVITO_API_KEY
+const INSTANCE_ID = process.env.SCRIVITO_TENANT || ''
```
apepper, line 12: It would be great if `dotenv` works here as well.

src/dumpContent.ts
```diff
@@ -0,0 +1,112 @@
+#!npx vite-node
+import fs from 'fs'
+import { promisify } from 'util'
+
+type ObjData = { _id: string }
+type SearchData = { continuation?: string; objs: ObjData[] }
+type BlobsData = { private_access: { get: { url: string } } }
+
+const DIRECTORY = 'contentDump'
```
apepper, line 9: First I would place this in `src/` since it will (most likely with the next PR) be used by the import script that is included in the SPA.

src/dumpContent.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+async function dumpBinary(id: string) {
+  const binary = await fetchJson<BlobsData>(`blobs/${encodeURIComponent(id)}`)
+  const url = binary.private_access.get.url
+  const response = await fetch(url)
+  if (response.status !== 200) throw new Error(`Failed to fetch ${url}`)
+  const blob = await response.blob()
+  fs.writeFileSync(
+    `${DIRECTORY}/binary/${encodeURIComponent(id)}`,
```
apepper, line 81: How about naming this folder `binaries` and the other folder `objs`?

src/dumpContent.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+async function dumpObjAndBinaries(objData: ObjData) {
+  await forEachBinaryId(objData, dumpBinary)
+  dumpObj(objData)
+}
+
+async function forEachBinaryId(
+  data: unknown,
+  handler: (binaryId: string) => Promise<void>,
```
apepper, line 57: Why is this function needed as a callback? Why not call the `dumpBinary` function directly?

src/dumpContent.ts
```diff
@@ -0,0 +1,112 @@
...
+) {
+  if (!data || typeof data !== 'object') return
+  if (isBinaryData(data)) await handler(data[1].id)
+  else for (const d of Object.values(data)) await forEachBinaryId(d, handler)
```
apepper, line 61: When can this case happen?

src/dumpContent.ts
```diff
@@ -0,0 +1,112 @@
...
+  let continuation
+
+  do {
+    const data: SearchData = await fetchJson<SearchData>(
+      `workspaces/published/objs/search`,
+      JSON.stringify({
+        continuation,
+        include_objs: true,
+        options: { site_aware: true },
+        size: 10,
```
apepper, line 40: Given the speed of this script is not 100% important: we could use `size: 1` to fetch one obj at a time. With that the printed indicator `.` would match the number of objs.

This would also allow to (maybe later) add a progress bar.

apepper (changes requested):
FEEDBACK ([1ea389b](https://github.com/Scrivito/scrivito-portal-app/pull/218/commits/1ea389b0f6c1d9fedc0fe71dafebb4fa1949e3f4))

Very nice 🚀! Good start.

I did not do a full in-depth review, but already looked at it to give some more directions - see comments.

src/dumpContent.ts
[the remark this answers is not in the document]
dcsaszar: Leftover from before refactoring

src/dumpContent.ts
dcsaszar: Yes it doesn't work yet for accessing raw binary content. See https://infopark.slack.com/archives/C6BU7G38F/p1701421365735489?thread_ts=1683206386.104579&cid=C6BU7G38F

src/dumpContent.ts
```diff
@@ -0,0 +1,112 @@
...
+async function dumpContent() {
+  let continuation
+
+  do {
+    const data: SearchData = await fetchJson<SearchData>(
+      `workspaces/published/objs/search`,
+      JSON.stringify({
+        continuation,
+        include_objs: true,
+        options: { site_aware: true },
```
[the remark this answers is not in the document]
dcsaszar: Yes, it's basically `Obj.onAllSites()`

src/dumpContent.ts
dcsaszar: I thought about it but realized that dumping would always be done from the Golden Master. Using dotenv would require to change the local config before dumping. Do we really want that? Or are you thinking of adding another local config or differently named env variables?

src/dumpContent.ts
dcsaszar: The import script is happy with this path. My local prototype can already restore from there.
Would you make it part of `src` nevertheless?

src/dumpContent.ts
dcsaszar: Thanks to GitHub I'm not 100% sure which case you mean. This function is a recursive object walker, so there can be
1. Leafs (e.g. strings)
2. Binary data attributes
3. Nested objects (e.g. widgets)

src/dumpContent.ts
dcsaszar: BTW a similar situation is the case for the restore part. AFAICS the needed APIs are not yet available with the editor token. An API key is currently required for restoring, too.

dcsaszar:
Implemented feedback and started discussions.

src/dumpContent.ts
apepper: No, thats ok. Let's leave it in `contentDump` then.

src/dumpContent.ts
apepper: Sounds good. Let's keep it like it is.

src/dumpContent.ts
apepper: A differently named env variable sounds good. This script will only be executed by maintainers of the project, e.g. you or me 😸. I would still prefer, if I can store the sensitive information in `.env.local` or similar.

src/dumpContent.ts
apepper: For the restore part it becomes even more important, to not work with `SCRIVITO_API_KEY`. A "regular" Scrivito Editor should be able to execute the restore without a `SCRIVITO_API_KEY`.

I escalated this to https://infopark.slack.com/archives/G011EJEGKD5/p1704362075767339 .

apepper: For the dumpContent part I could live with the `SCRIVITO_API_KEY` for now. But please refactor `fetchJson` to have the same signature as `unstable_JrRestApi.fetch`, so it becomes easier to replace it here, once the backend is able to use `unstable_JrRestApi`.

src/dumpContent.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+function isBinaryData(data: unknown): data is ['binary', { id: string }] {
+  return (
+    Array.isArray(data) &&
+    data.length === 2 &&
+    data[0] === 'binary' &&
+    typeof data[1] === 'object' &&
+    data[1].id
+  )
```
apepper, line 69: I also suggest to make it more robust:

```suggestion
  if (!Array.isArray(data)) return false

  const [attributeType, attributeValue] = data as unknown[]
  return (
    attributeType === 'binary' &&
    isObject(attributeValue) &&
    typeof (attributeValue as { id: unknown }).id === 'string'
  )
```

Otherwise it might throw a TypeError:

```js
isBinaryData(['binary', null])
// => Uncaught TypeError: Cannot read properties of null (reading 'id')
```

And also returns "truish" for other types as well:

```js
isBinaryData(['binary', { id: { notA: 'string, but an object' } }])
// => {notA: 'string, but an object'}
```

src/dumpContent.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+async function dumpBinary(id: string) {
```
apepper, line 72: I suggest to rename `id` to `binaryId`.

src/dumpContent.ts
```diff
@@ -0,0 +1,110 @@
...
+  const binary = await fetchJson<BlobsData>(`blobs/${encodeURIComponent(id)}`)
+  const url = binary.private_access.get.url
+  const response = await fetch(url)
+  if (response.status !== 200) throw new Error(`Failed to fetch ${url}`)
+  const blob = await response.blob()
+  fs.writeFileSync(
+    `${DIRECTORY}/binaries/${encodeURIComponent(id)}`,
```
apepper, line 79: Can you make `${DIRECTORY}/binaries` and `${DIRECTORY}/objs` DRY? Currently it is used in two different places.

src/dumpContent.ts
```diff
@@ -0,0 +1,110 @@
...
+
+const DIRECTORY = 'contentDump'
+
+const API_KEY = process.env.SCRIVITO_API_KEY
+const INSTANCE_ID = process.env.SCRIVITO_TENANT
+
+if (INSTANCE_ID && API_KEY) {
+  clearDump()
+  await dumpContent()
+  console.log('\n✅ Dump complete.')
```
apepper, line 17: I think it would be beneficial if we have an obj/binary counter in the result message. E.g. something like `✅ Dump complete (112 objs and 37 binaries).`.

src/dumpContent.ts
```diff
@@ -0,0 +1,110 @@
...
+async function dumpObjAndBinaries(objData: ObjData) {
+  await dumpBinaries(objData)
+  dumpObj(objData)
+}
+
+async function dumpBinaries(data: unknown) {
+  if (!data || typeof data !== 'object') return
+  if (isBinaryData(data)) await dumpBinary(data[1].id)
+  else for (const d of Object.values(data)) await dumpBinaries(d)
+}
```
apepper, line 60: I have a hard time to understand this function. Now - after longer reading of the actual code and data structure of the objs - I think we should make it more explicit. I would explicitly handle `_widget_pool` and avoid recursion. Something like this:

```ts
// earlier in the file:
type ObjData = { _id: string; _widget_pool: { [key: string]: object } }

async function dumpBinaries(objData: ObjData) {
  const { _widget_pool, ...otherAttributes } = objData

  for (const widget of Object.values(_widget_pool)) {
    for (const widgetAttribute of Object.values(widget)) {
      await dumpBinaryAttribute(widgetAttribute)
    }
  }

  for (const objAttribute of Object.values(otherAttributes)) {
    await dumpBinaryAttribute(objAttribute)
  }
}

async function dumpBinaryAttribute(data: unknown) {
  if (!isBinaryAttribute(data)) return
  return dumpBinary(data[1].id)
}
```

WDYT?

apepper (changes requested):
FEEDBACK ([8ab91bf](https://github.com/Scrivito/scrivito-portal-app/pull/218/commits/8ab91bf9e086c37f1f58438b512d58f33c785fa1))

Nice progress! I now did a thorough review (see my comments). I also checked with the backend team. They claim, that all the used routes are now able to be used with IAM token.

So with that:

* [ ]  According to https://infopark.slack.com/archives/G011EJEGKD5/p1704362075767339 all of the used routes should work with IAM tokens. Please use `unstable_JrRestApi`. If something is not working, please clarify directly with the backend team.

src/dumpContent.ts
dcsaszar: Binary attributes in widgets are very rare. Having different code paths for widget and obj attributes may lead to a situation where the widget code path is never taken. Since we have no tests, this may lead to bugs that stay undetected for very long.
I admit that expecting (searching for) binary attributes anywhere is a simplification but I don't think it's an oversimplification since the very specific type signature can not happen by chance.

I'll give it a try anyway.

dcsaszar:
> Please use unstable_JrRestApi.

Since `unstable_JrRestApi` doesn't send the special header we can't use it.

src/dumpContent.ts
```diff
@@ -60,12 +60,13 @@ async function dumpBinaries(data: unknown) {
 }
 
 function isBinaryAttribute(data: unknown): data is ['binary', { id: string }] {
+  if (!Array.isArray(data)) return false
```
apepper, line 84: Strange, that TypeScript uses the following signature: `isArray(arg: any): arg is any[];`. I would have expected `arg is unknown[];`. Anyway your code works 😸.

src/dumpContent.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+async function dumpObjAndBinaries(objData: ObjData) {
+  await dumpBinaries(objData)
+  dumpObj(objData)
+}
+
+async function dumpBinaries(data: ObjData | WidgetData) {
+  for (const value of Object.values(data))
+    if (isBinaryAttribute(value)) await dumpBinary(value[1].id)
```
apepper, line 77:
```suggestion
  for (const value of Object.values(data)) {
    if (isBinaryAttribute(value)) await dumpBinary(value[1].id)
  }
```

src/dumpContent.ts
```diff
@@ -0,0 +1,161 @@
...
+async function fetchJson<T>(
+  apiPath: string,
+  options: { data?: Record<string, unknown>; method?: string } = {},
+): Promise<T> {
+  process.stdout.write('.')
+
+  if (!apiToken) apiToken = await fetchIamToken()
+
+  const response = await fetch(
+    `https://api.scrivito.com/tenants/${INSTANCE_ID}/${apiPath}`,
```
apepper, line 124: Isn't there a `api.justrelate.com/scrivito` route for this as well?

src/dumpContent.ts
```diff
@@ -0,0 +1,161 @@
...
+    {
+      body: options.data ? JSON.stringify(options.data) : undefined,
+      headers: {
+        Authorization: `Bearer ${apiToken}`,
+        'Content-Type': 'application/json',
+        'Scrivito-Access-As': 'editor',
```
apepper, line 130: This looks strange. Why do we need to claim to be editor? Is this the "special header" that is missing in order for us to use `unstable_JrRestApi` directly?

apepper (changes requested):
FEEDBACK ([f4d7502](https://github.com/Scrivito/scrivito-portal-app/pull/218/commits/f4d75020273287ee356ac7866f96748997f70aa4))

Very nice 🚀!

There are only some minor comments and one bigger question regarding `unstable_JrRestApi` and its headers.

src/dumpContent.ts
dcsaszar: Yes, see the discussion:
https://infopark.slack.com/archives/G011EJEGKD5/p1704365149566769?thread_ts=1704362075.767339&cid=G011EJEGKD5

Also: https://infopark.slack.com/archives/C04TN8LJA/p1701704213052029?thread_ts=1701184869.334039&cid=C04TN8LJA

src/dumpContent.ts
dcsaszar: There is, but currently it has a bug which makes it fail for blobs. 😿

apepper: I see. So for now this is "good enough".

src/dumpContent.ts
apepper: I talked with Kris and `'Scrivito-Access-As': 'editor'` is more or less public API. I also filed a wish for `unstable_JrRestApi` to allow additional `headers`.

PR #221, opened 2024-01-03
Fix dependabot.yml

dcsaszar:
As we already group patch and minor updates in `update-patch-minor` the remaining updates can only be major ones, so no need to ignore anything.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,11 +10,3 @@ updates:
           - 'patch'
     schedule:
       interval: 'weekly'
-  - package-ecosystem: 'npm'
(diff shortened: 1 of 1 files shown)


apepper:
But will the remaining updates be created by dependabot? So far I have not seen one "only major" update.

apepper (approved):
LGTM ([c9f1268](https://github.com/Scrivito/scrivito-portal-app/pull/221/commits/c9f126812cacccb891e58e2e31e9821bdab3be84))

But I'm not sure, if major one-package PRs will now be done by dependabot. Lets see.

PR #227, opened 2024-01-04
DataListWidget: Allow to show footer

apepper:
Adds an option to show a footer for a data list widget. This allows to e.g. add a data form to push items to the data list widget list.

### DIFF
--- a/src/Widgets/DataListWidget/DataListWidgetClass.ts
+++ b/src/Widgets/DataListWidget/DataListWidgetClass.ts
@@ -4,7 +4,9 @@ export const DataListWidget = provideWidgetClass('DataListWidget', {
   attributes: {
     content: 'widgetlist',
     data: 'datalocator',
+    footer: 'widgetlist',
     nothingFound: 'widgetlist',
     nrOfColumns: ['enum', { values: ['1', '2', '3', '4', '5', '6'] }],
+    showFooter: 'boolean',
   },
 })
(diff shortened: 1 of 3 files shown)


dcsaszar (commented):
Code LGTM, but:

### Discussion

We use this approach in the console prototype but I am unhappy with it.

* Recently we introduced details pages where widgets on a page inherit the data class of a page.
* This means that most widgets (including a data list widget and a data form widget) don't need a data selector at all.
* This in turn means that it's already trivial for editors to build a data list with a form in a flexible manner, e.g.:
```
 [ Data list widget ]
 [ Data form widget ]
```
Which only leaves the edge case that a data list uses a different data locator from the page default. If needed, this could be implemented taking advantage of nested data scopes and structure widgets:
```
  [ Group widget with data locator]
    [ Data list widget ]
    [ Data form widget ]
  [ End group widget ]
```

This has some advantages:

* Editors are flexible in the layout of list + form (on top, under, left, right; throw in a descriptive text, etc.)
* We don't clutter the WYSIWYG with "Editor notes"

WDYT?
Could the use case you had in mind be build like this?

apepper:
How does the data list read out the datalocator from the obj? So far I only found code in the `DataListWidetComponent` of the console to read out the datalocator from the widget.

dcsaszar:
> How does the data list read out the datalocator from the obj? So far I only found code in the `DataListWidetComponent` of the console to read out the datalocator from the widget.

Yes, so far we haven't implemented this. I can imagine a React.Context for passing down the data locator. Candidates for providing the context: layout, page, group widget.

apepper:
I like your proposed idea. A "Group widget" with a datalocator e.g. for TicketMessage would work. But I see this as an overall refactoring, that we should do once we switch `DataListWidget` to derive the current data from the stack.

PR #228, opened 2024-01-05
DataLabelWidget: Add "datetime" option

apepper:
For now it only supports "relative" dates. More datetime options will probably come in the future.

For dates less then a month ago it will show a relative date. For dates older then a month it will show day and month. For all other dates it will show day, month and year.

### DIFF
--- a/src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
+++ b/src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
@@ -52,17 +52,26 @@ const AttributeValue = connect(function AttributeValue({
   attributeValue: unknown
   showAs: string | null
 }) {
-  if (showAs === 'currency') return formatCurrency(attributeValue, currency)
-  if (showAs === 'datetime') return formatDatetime(attributeValue)
+  if (showAs === 'currency') {
+    return <Currency value={attributeValue} currency={currency} />
+  }
 
-  return formatText(attributeValue)
+  if (showAs === 'datetime') return <Datetime value={attributeValue} />
+
+  return <Text value={attributeValue} />
 })
 
-function formatText(value: unknown) {
+function Text({ value }: { value: unknown }) {
   return value ? value.toString() : 'N/A'
 }
 
-function formatCurrency(value: unknown, currency: string | null) {
+function Currency({
+  value,
+  currency,
+}: {
+  value: unknown
+  currency: string | null
+}) {
   if (value === null) return 'N/A'
 
   const number = Number(value)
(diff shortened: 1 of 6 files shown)


src/Widgets/DataLabelWidget/RelativeDate.tsx
```diff
@@ -0,0 +1,57 @@
+import { formatDistance } from 'date-fns'
```
dcsaszar, line 1: * [ ] FYI I see that we have a lot of date logic in this module anyway. So we could also use the JS built-in relative helpers instead of adding a dependency.

See https://github.com/infopark/scrivito.com/pull/1823 for an example.

src/Widgets/DataLabelWidget/DataLabelWidgetEditingConfig.ts
```diff
@@ -13,6 +13,7 @@ provideEditingConfig(DataLabelWidget, {
       values: [
         { value: 'text', title: 'Text' },
         { value: 'currency', title: 'Currency' },
+        { value: 'datetime', title: 'Date time' },
```
dcsaszar, line 16: * [ ] I don't think "date time" is a common term. How about "Date/time" or just "Date"?

src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
```diff
@@ -73,3 +75,12 @@ function formatCurrency(value: unknown, currency: string | null) {
 
   return formatter.format(number)
 }
+
+function formatDatetime(value: unknown) {
+  if (value === null) return 'N/A'
+
+  const date = new Date(value as string)
```
dcsaszar, line 82: * [ ] Can editors break this by just playing around with settings in the UI or is value guaranteed to be a string? If we don't know for sure we should do a type check instead of casting.

src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
```diff
@@ -52,6 +53,7 @@ const AttributeValue = connect(function AttributeValue({
   showAs: string | null
 }) {
   if (showAs === 'currency') return formatCurrency(attributeValue, currency)
+  if (showAs === 'datetime') return formatDatetime(attributeValue)
```
dcsaszar, line 56: * [ ] FYI all of these functions look like helpers but are actually components and could be written as such.

```
if (showAs === 'datetime') return <Datetime value={attributeValue} />
```

dcsaszar (changes requested):
Looks good. See comments for minor stuff.

src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
apepper: To my understanding `new Date(value)` can take any kind of `value` and return an invalid date (that is the check on line 83 `if (Number.isNaN(date.getTime())) return 'N/A'`). But I'll make this code more explicit to no longer need the cast.

apepper:
REVIEW ([7daf996](https://github.com/Scrivito/scrivito-portal-app/pull/228/commits/7daf99609c3c635bfd25ec006331d6c48d05d4f8))

PR #229, opened 2024-01-08
Less TODOs

agessler:
z-0 deleted + @include transition translated to native CSS + inline styles deleted for buttons, sections + if btn empty it displays message

### DIFF
--- a/src/Widgets/ButtonWidget/ButtonWidgetComponent.tsx
+++ b/src/Widgets/ButtonWidget/ButtonWidgetComponent.tsx
@@ -23,7 +23,6 @@ provideComponent(ButtonWidget, ({ widget }) => {
       <LinkTag
         to={target}
         className={buttonClassNames.join(' ')}
-        style={{ minHeight: '40px', minWidth: '35px' }} // TODO: Check this by a designer
       >
         {!title && isInPlaceEditingActive()
           ? 'Provide the button link and text in the widget properties.'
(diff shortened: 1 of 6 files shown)


src/Widgets/ButtonWidget/ButtonWidgetComponent.tsx
```diff
@@ -20,11 +20,7 @@ provideComponent(ButtonWidget, ({ widget }) => {
 
   return (
     <WidgetTag className={alignmentClassNameWithBlock(widget.get('alignment'))}>
-      <LinkTag
-        to={target}
-        className={buttonClassNames.join(' ')}
-        style={{ minHeight: '40px', minWidth: '35px' }} // TODO: Check this by a designer
```
apepper, line 26: A button without a label now more or less collapses: 
Screenshot 2024-01-11 at 11 22 06

Please ensure, that it has at least a min height that is equal to the height with a text.

This is also important during loading - Scrivito will first render an "empty" button before rendering the button with the label, once the target obj is loaded.

apepper (changes requested):
FEEDBACK ([84bdec9](https://github.com/Scrivito/scrivito-portal-app/pull/229/commits/84bdec94561c00a7272d7e0388e140a13fc63a4d))

Nice start 🚀! Please see my comment for open issues.

apepper (approved):
LGTM ([08e19a2](https://github.com/Scrivito/scrivito-portal-app/pull/229/commits/08e19a253d063b8f8855a904277d6296cb7e9a23))

Cool stuff!

PR #230, opened 2024-01-08
provideLocalStorageDataClass with initial content + Orders from local storage

apepper:
The data used for "local storage" is directly taken from the current response of PisaSales with example content.

Needs new content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://improved-providelocalstorage.scrivito-portal-app.pages.dev/orders-02d09ad2dd7d2332?_scrivito_workspace_id=je305f894b0a1bb4&_scrivito_display_mode=view


src/utils/provideLocalStorageDataClass.ts
```diff
@@ -91,3 +91,33 @@ function isStringObject(item: unknown): item is Record<string, unknown> {
...
+): [string, unknown][] {
+  if (order.length === 0) return items
+
+  return [...items].sort((a, b) => {
+    const itemA = a[1]
+    const itemB = b[1]
+    if (!isObject(itemA) || !isObject(itemB)) return 0
+
+    for (const [attr, ascOrDesc] of order) {
+      const valueA = (itemA as Record<string, unknown>)[attr]
```
dcsaszar, line 143: * [ ] Suspicious cast. `object` suddenly becomes a more special type. We should use a better type guard than `isObject` and get rid of the cast. `isStringObject` is close, we probably need something like `isStringNumberObject`.

src/utils/provideLocalStorageDataClass.ts
```diff
@@ -91,3 +91,33 @@ function isStringObject(item: unknown): item is Record<string, unknown> {
...
   return Object.keys(item).every((key) => typeof key === 'string')
 }
+
+function sortItems(
+  items: [string, unknown][],
+  order: Array<[string, 'asc' | 'desc']>,
+): [string, unknown][] {
+  if (order.length === 0) return items
+
+  return [...items].sort((a, b) => {
```
dcsaszar, line 137: * [ ] Since we're using lodash anyway, I wonder if we could just use `orderBy` instead of rolling our own? It accepts nested field paths, direction, and multiple criteria.

src/utils/provideLocalStorageDataClass.ts
```diff
@@ -61,10 +69,38 @@ export function provideLocalStorageDataClass(className: string) {
...
+        const initialRecord: Record<string, Record<string, unknown>> = {}
+        initialContent.forEach((item) => {
+          const id = item._id
+
+          if (Object.keys(initialRecord).includes(id)) {
+            throw new Error(`Duplicate _id ${id} in initialContent!`)
+          }
+
+          initialRecord[id] = item
+        })
```
dcsaszar, line 91: * [ ] This is a safety check for our own data. I don't see why we would need that. If we have an error in the initialContent we can just fix it.

src/utils/provideLocalStorageDataClass.ts
```diff
@@ -61,10 +69,38 @@ export function provideLocalStorageDataClass(className: string) {
     },
   })
 
+  async function initializeContent(
+    initialContent: { _id: string; [key: string]: unknown }[],
+  ) {
+    const initializedKey = `${recordKey}-initialized`
+    const initializedValue = `yes - already initialized with ${await sha1(
+      JSON.stringify(initialContent),
+    )}`
```
dcsaszar, line 78: * [ ] Given the needed local storage is proportional to the number of records anyway, we could simplify the code by using `JSON.stringify(initialContent)` directly as the `initializedValue`.

dcsaszar (changes requested):
Looks good in general. I'd like to simplify the code given this is not meant for production anyway. See comments.

src/utils/provideLocalStorageDataClass.ts
apepper: I would prefer to keep it - since it does not at much complexity to the code.

apepper:
REVIEW ([0c89b32](https://github.com/Scrivito/scrivito-portal-app/pull/230/commits/0c89b32e2ac2a852eaa482788f14e0d725698360))

src/utils/provideLocalStorageDataClass.ts
dcsaszar: After the latest refactoring this is the most complex (for example it uses exception driven control flow) but least useful (see above) part of the module.
* [ ] FYI If we really need it we could make it more readable by extracting a well-named helper method.

dcsaszar (approved):
Cool stuff.

PR #239, opened 2024-01-12
bootstrap-icons 1.11.3 + Provide all bootstrap-icons also as SVGs

apepper:


dcsaszar (commented):
I see copying the files is mentioned under "External image" in the docs.
But since vite has static asset handling by default (https://vitejs.dev/guide/assets#importing-asset-as-url) how about `npm i bootstrap-icons` and:

```js
import apple from 'bootstrap-icons/icons/apple.svg'

function Apple() {
  return <img src={apple} />
}
```

apepper:
Our designer is no big fan of taking bootstrap directly from npm. That's why we added a copy to the project.

dcsaszar:
Without a concrete reason why the designer prefers the icons in the src instead of the node_modules I can't really judge the "big fan" aspect here.

For developers the advantages would be:
* Smaller repo (currently src is 3.2M instead of 12M)
* Ability to review updates (currently GitHub just crashes)

I see that the `design` folder already has symlinks. With a symlink for the icons there would be virtually no difference for the designer in the way they can work.

As long as the designer doesn't plan to do "dirty stuff" with the icons folder I don't see why they shouldn't be a fan of the package approach. If they plan to do "dirty stuff" this may become hard to review for maintainers, as we typically don't touch vendor folders.

apepper:
FYI: https://github.com/Scrivito/scrivito-portal-app/pull/244

PR #244, opened 2024-01-16
npm based bootstrap-icons + version 1.11.3

apepper:
See https://github.com/Scrivito/scrivito-portal-app/pull/239#issuecomment-1891797897 for motivation. `npm run start-design` continues to just work.

Closes #239.


agessler:
I don't provide SVGs for these Bootstrap icon set. If somebody wants this feel free to do.

apepper:
> I don't provide SVGs for these Bootstrap icon set. If somebody wants this feel free to do.

There is no need. The package already also provides all svg files for every icon:

Screenshot 2024-01-16 at 14 41 18

PR #251, opened 2024-01-23
Add "_id" attributes to many data classes

apepper:


apepper:
FYI: Needs Scrivito 1.41.0 (not yet released).

dcsaszar (approved):
Looks good in general. My only question is
* [ ] Which criteria do we use to decide if an ID should be part of the editing config or not?

I'd like to avoid overloading the UI with details editors may never need.

apepper:
> Which criteria do we use to decide if an ID should be part of the editing config or not?

Roughly: If one could want to filter for this attribute (e.g. look for the user of the current ticket). But since this can't be completely known for all cases I would say: Add `_id` Attributes for all data classes, except for n:m relation ship data classes.

PR #289, opened 2024-02-07
Support general marketing consent and subscriptions

dcsaszar:
The naming is taken from the Neoletter UI Profile/Subscriptions tab.

First go. The `fetchMySubscribedTopicIds` helper is duplicated. Ideas for if and where to share code between data connectors are welcome.

Demo content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dave-neoletter-subscriptions.scrivito-portal-app.pages.dev/subscriptions-027be0e907d46d8e?_scrivito_workspace_id=s204359a88b63a0b&_scrivito_display_mode=view (FYI The icons need an adjustment to allow for boolean values but I consider this out of scope for this PR.)

Neoletter topics: https://edit.neoletter.com/i/d0a154d76edf2a7bd991fc658e700a1d/topics


apepper (commented):
FEEDBACK ([398a115](https://github.com/Scrivito/scrivito-portal-app/pull/289/commits/398a1151a6c180647aa19c234b7a1b6495f2c8fc))

I like the overall idea how you have structured the two endpoints into two other data classes/items.

But we definitely need to work on the UI for this: About about we introduce a `DataFormToggleWidget` that allows to toggle two different states right away (similar to the boolean toggle. This would combine the "data icon", "data form boolean" and "data form button submit" into one.

If you like the idea can you please coordinate with Alex G. to get render-designs and a widget thumbnail for that from him? If it helps, you could submit this `DataFormToggleWidget` as an independent PR. Or if you like you can include it in this PR.

Also it would be good if the content is more explainatory to the visitor. Maybe you can get together with Andreas to improve the wordings of the specific subscriptions page and also add a description for each broad section (e.g. what does a "General marketing consent" consent actually allow).

dcsaszar:
> Also it would be good if the content is more explainatory to the visitor. Maybe you can get together with Andreas to improve the wordings of the specific subscriptions page and also add a description for each broad section (e.g. what does a "General marketing consent" consent actually allow).

FYI I copied the wording 1:1 from
* the examples on https://docs.neoletter.com/topics-259cbad42938db68
* the Neoletter UI

So, if we improve it, we should ideally also improve it in the Neoletter UI and in the docs. CC @aviebke

apepper:
> So, if we improve it, we should ideally also improve it in the Neoletter UI and in the docs.
> 
> image

@dcsaszar Where can I find this UI?

dcsaszar:
> @dcsaszar Where can I find this UI?

https://edit.neoletter.com/i/d0a154d76edf2a7bd991fc658e700a1d/profiles/437d3335797849af99462f0b7a7c1a5c?tab=subscriptions

src/Data/Subscription/SubscriptionDataClass.ts
```diff
@@ -0,0 +1,97 @@
...
+          ),
+        ) as { results: Topic[] }
+      ).results
+
+      const mySubscribedTopicIds = await fetchMySubscribedTopicIds()
+
+      return {
+        results: myTopics.map(({ id, description, title }) => ({
+          id,
+          isConsentGiven: mySubscribedTopicIds.includes(id),
```
apepper, line 30: If we don't convert `status` to a boolean, but to `status` `"revoked"`/`"given"` instead, we would have not need to implement boolean handling into `DataIconWidget` etc.

apepper: From the perspective of the end-user there is no big different between unknown and revoked. How about we introduce the artificial status `"given"`/`"notGiven"` to avoid the word "revoke"?

apepper: Or `"subscribed"`/`"unsubscribed"`.

dcsaszar:
@apepper REVIEW

* The conditional widget from 07623d2 is now unused
* The style is the Bootstrap default and can later be adjusted
* We got some explanatory input for the general marketing consent but haven't found time to update the content yet (CC @aviebke)

image

aviebke:
I was talking to Jan today about how the "General marketing consent" works, explained it in the docs, and I am now waiting for him to digest what I've written and give feedback. Afterwards, we can extract from this text what we think is a useful addition to the Neoletter UI or the portal.

Feel free to take a look at the [working copy](https://edit-beta.scrivito.com/6d226c03f32a8ea4e8ae8e5cbe6c6e2c~https://docs.neoletter.com/profiles-c4fea38cff3bc2be?_scrivito_workspace_id=obd7e08187b64219&_scrivito_display_mode=view#general-marketing-consent) already at this stage.

src/Widgets/DataConditionWidget/DataConditionWidgetClass.ts
```diff
@@ -0,0 +1,9 @@
+import { provideWidgetClass } from 'scrivito'
+
+export const DataConditionWidget = provideWidgetClass('DataConditionWidget', {
```
apepper, line 3: Please extract (or remove) `DataConditionWidget` from this PR, since it is no longer in use.

src/Widgets/DataFormBooleanWidget/DataFormBooleanWidgetEditingConfig.ts
```diff
@@ -10,21 +10,47 @@ provideEditingConfig(DataFormBooleanWidget, {
...
+      description: 'Default: boolean true',
+    },
     defaultValue: {
       title: 'Default value',
       description:
         'Only relevant for a create form. An update form will utilize the existing boolean value.',
     },
     helpText: { title: 'Help text' },
+    style: {
+      title: 'Style',
```
apepper, line 28: Please add:

```ts
  description: 'Default: Checkbox'
```

src/Widgets/DataFormBooleanWidget/DataFormBooleanWidgetEditingConfig.ts
```diff
@@ -10,21 +10,47 @@ provideEditingConfig(DataFormBooleanWidget, {
...
+      values: [
+        { value: 'check', title: 'Checkbox' },
+        { value: 'switch', title: 'Switch' },
+      ],
+    },
+    submitOnChange: { title: 'Submit on change?' },
```
apepper, line 34: How about the title `'Immediately submit changed value?'`? Also a description should be added, that this will not apply for new items, only for updating existing items.

src/Widgets/DataFormBooleanWidget/DataFormBooleanWidgetComponent.tsx
```diff
@@ -15,27 +15,40 @@ provideComponent(DataFormBooleanWidget, ({ widget }) => {
   const labelOptions: { htmlFor?: string } = {}
   if (!isInPlaceEditingActive()) labelOptions.htmlFor = id
 
+  const className =
```
apepper, line 18: I suggest to rename `className` to `style`.

src/Data/Subscription/SubscriptionDataClass.ts
```diff
@@ -0,0 +1,93 @@
...
+          ),
+        ) as { results: Topic[] }
+      ).results
+
+      const mySubscribedTopicIds = await fetchMySubscribedTopicIds()
+
+      return {
+        results: myTopics.map(({ id, description, title }) => ({
+          id,
+          consent: mySubscribedTopicIds.includes(id) ? 'given' : 'notGiven',
```
apepper, line 30: Since we are now showing/updating the value using the `DataFormBooleanWidget` we can switch back to a real boolean here:

```suggestion
          consent: mySubscribedTopicIds.includes(id)
```

This would also make the `dataForNo` and `dataForYes` option in this PR obsolete.

Please also adjust the `consent` flag in the `GeneralMarketingConsent` accordingly.

apepper (changes requested):
FEEDBACK ([61c0b71](https://github.com/Scrivito/scrivito-portal-app/pull/289/commits/61c0b710aa9a76bcf89bf7406ccb82b772f53286))

We're moving in a good direction!

Please see my comments regarding the code.

Content wise please improve the "General marketing consent" section (e.g. with improved wording by Andreas). The goal would be to have a working copy that could be published directly once this PR is approved and merged.

src/Widgets/DataFormBooleanWidget/DataFormBooleanWidgetComponent.tsx
```diff
@@ -15,27 +15,40 @@ provideComponent(DataFormBooleanWidget, ({ widget }) => {
...
+    widget.get('style') === 'switch' ? 'form-check form-switch' : 'form-check'
+
+  const submitOnChange = widget.get('submitOnChange')
```
krishan, line 21: Why is `submitOnChange` a property of the `BooleanWidget`?
IMHO it should be a property of the `FormWidget`.

Because having a form where some fields are "submitOnChange" and others are not would be confusing.

Let's keep it simple for the editor and say "a form either submits on change, or i does not".

Or am I missing something?

src/Widgets/DataFormBooleanWidget/DataFormBooleanWidgetClass.ts
```diff
@@ -5,10 +5,14 @@ export const DataFormBooleanWidget = provideWidgetClass(
   {
     attributes: {
       attributeName: 'string',
+      dataForNo: 'string',
+      dataForYes: 'string',
```
krishan, line 9: Why do we need something complicated like `dataForNo` or `dataForYes`?

Does Neoletter really have such a complex API? Isn't it a simple boolean?

krishan:
General question regarding the approach:

You seem to be working on a intricate upgrade for the boolean toggle, which introduces a lot of complexity into the model (`submitOnChange`, `dataForYes` and such).

Why not just use a simple button ("subscribe" or "unsubscribe")?

src/Widgets/DataFormBooleanWidget/DataFormBooleanWidgetClass.ts
krishan: Generally, if something is called ` BooleanWidgetClass`, shouldn't it use booleans?

If the neoletter API needs "special words" instead of booleans, that would be the responsiblity of the data connection.

apepper: I think this was an intermediate step. Please see https://github.com/Scrivito/scrivito-portal-app/pull/289#discussion_r1492238961 .

src/Widgets/DataFormBooleanWidget/DataFormBooleanWidgetComponent.tsx
apepper: Good idea: My initial idea was, that only a `DataFormBooleanWidget` without a container could be used. But that would duplicate the save logic to `DataFormBooleanWidget`.

So yeah, let's move `submitOnChange` up to the `DataFormContainerWidget`.

apepper:
> You seem to be working on a intricate upgrade for the boolean toggle, which introduces a lot of complexity into the model (`submitOnChange`, `dataForYes` and such).
> 
> Why not just use a simple button ("subscribe" or "unsubscribe")?

A `subscribe`/`unsubscribe` button was Daves initial implementation (see https://github.com/Scrivito/scrivito-portal-app/pull/289#issuecomment-1936141494). But it relied on a `DataConditionWidget` that original also relied on a string representation of a boolean value. One could change this boolean value to a string like `'given'`/`'notGiven'` to avoid the "string representation of boolean value".

But to me a boolean-toggle seems to be more "state of the art" for subscriptions, that's why I ask Dave to go in this direction instead.

krishan:
> it relied on a DataConditionWidget that original also relied on a string representation of a boolean value

Side Note: As you know, my plan is to allow for DataLocators to refer to attributes (see: https://github.com/infopark/scrivito_js/issues/10255 ). AFAICS this would also help us here: If the `DataConditionWidget` finds that its datalocator points to an attribute of type boolean, it just uses that. 

So that would have solve this problem here, right?

apepper:
> Side Note: As you know, my plan is to allow for DataLocators to refer to attributes (see: [infopark/scrivito_js#10255](https://github.com/infopark/scrivito_js/issues/10255) ). AFAICS this would also help us here: If the `DataConditionWidget` finds that its datalocator points to an attribute of type boolean, it just uses that.
> 
> So that would have solve this problem here, right?

With the implementation of `DataConditionWidget` as it was proposed by Dave this would solve the problem here. But I'm not yet 100% sure, if the proposed `DataConditionWidget` would work well for any other non-boolean attribute.

src/Data/GeneralMarketingConsent/GeneralMarketingConsentEditingConfig.ts
```diff
@@ -0,0 +1,9 @@
+import { provideEditingConfig } from 'scrivito'
+import { GeneralMarketingConsent } from './GeneralMarketingConsentDataItem'
+
+provideEditingConfig(GeneralMarketingConsent, {
+  title: 'General marketing consent',
+  attributes: {
+    consent: { title: 'Consent' },
```
apepper, line 7:
```suggestion
    isConsentGiven: { title: 'Is consent given?' },
```

src/Data/Subscription/SubscriptionEditingConfig.ts
```diff
@@ -0,0 +1,11 @@
+import { provideEditingConfig } from 'scrivito'
+import { Subscription } from './SubscriptionDataClass'
+
+provideEditingConfig(Subscription, {
+  title: 'Subscription',
+  attributes: {
+    title: { title: 'Title' },
+    description: { title: 'Description' },
+    consent: { title: 'Consent' },
```
apepper, line 9:
```suggestion
    isConsentGiven: { title: 'Is consent given?' },
```

src/Widgets/DataFormBooleanWidget/DataFormBooleanWidgetComponent.tsx
```diff
@@ -12,17 +12,21 @@ provideComponent(DataFormBooleanWidget, ({ widget }) => {
...
-      ? attributeValue
-      : widget.get('defaultValue')
+  const defaultChecked = !!(attributeValue ?? widget.get('defaultValue'))
 
   return (
-    <div className="mb-3" key={[id, attributeName, defaultChecked].join('-')}>
+    <div
+      className={['mb-3', 'form-check']
+        .concat(style === 'switch' ? 'form-switch' : [])
+        .join(' ')}
```
apepper, line 24: How about this?

```ts
const classNames = ['mb-3', 'form-check']
if (widget.get('style') === 'switch') { classNames.push('form-switch') }

// ...
className={classNames.join(' ')}
```

apepper (changes requested):
FEEDBACK ([c0ce476](https://github.com/Scrivito/scrivito-portal-app/pull/289/commits/c0ce476a57c4b8ba586dbf17d7a4edf61d8d30ec))

Nice 🚀!

Almost there. Only some minor details can be improved (see comments).

dcsaszar:
REVIEW

The content for "MAILINGS NOT LINKED TO A NEWSLETTER TOPIC" has been approved by Andreas.

apepper (approved):
LGTM ([a72058f](https://github.com/Scrivito/scrivito-portal-app/pull/289/commits/a72058facef15c4a07f52d1347f4aebceb7699e2))

Very nice 🚀!

PR #296, opened 2024-02-13
Run dependabot only once a month

apepper:
Once a week is too much noise for me.

I did not find a good way to delay "major" updates for e.g. a week or a month. Thoughts?


dcsaszar (approved):
LGTM let's just try and see

PR #297, opened 2024-02-13
Updated pisa API data

apepper:
Needed content changes: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://new-pisa-data.scrivito-portal-app.pages.dev/my-tynacoon-59274260f223b65f?_scrivito_workspace_id=i4c55a37bc5458f8&_scrivito_display_mode=view


apepper:
Please focus your review on the editing config and the content changes in the working copy. Did I cover all places, that changes from a "localized only" value to an enum value? Please also keep in mind, that the

src/utils/provideLocalStorageDataClass.ts
```diff
@@ -36,7 +36,11 @@ export function provideLocalStorageDataClass(
             : items.filter((item) =>
                 Object.entries(filters).every(
                   ([filterAttribute, filterValue]) =>
-                    item[filterAttribute] === filterValue,
+                    item[filterAttribute] === filterValue ||
+                    (filterValue === 'true' &&
+                      item[filterAttribute] === true) ||
+                    (filterValue === 'false' &&
+                      item[filterAttribute] === false),
```
dcsaszar, line 43: * [ ] FYI Alternatively we could simplify/generalize the check with something like `item[filterAttribute]?.toString() === filterValue`

src/Data/Ticket/TicketEditingConfig.ts
```diff
@@ -8,11 +8,15 @@ provideEditingConfig(Ticket, {
     description: { title: 'Description' },
     number: { title: 'Ticket number' },
     referenceNumber: { title: 'Reference number' },
-    type: { title: 'Type' },
-    status: { title: 'Status' },
+    type: { title: 'Type (enum)' },
```
dcsaszar, line 11: * [ ] FYI the term "enum" is developer specific and editors may fail to understand it. To a certain extent this also applies to "localized". Ideas:
  * Type (code), Type (human readable)
  * Type code, Type description

src/Data/Ticket/TicketDataClass.ts
```diff
@@ -8,7 +8,15 @@ export const Ticket = provideLocalStorageDataClass('Ticket', {
 
     return {
       ...newData,
-      status: newData.status || 'captured',
+      type:
+        newData.type || newData.typeLocalized === 'Question'
+          ? 'PSA_SVC_CAL'
+          : newData.typeLocalized === 'Claim'
+            ? 'PSA_SVC_CPL'
+            : 'PSA_SVC_TRB',
```
dcsaszar, line 16: * [ ] FYI complex frontend logic a at least a yellow, if not a red flag. Every consumer has to agree on the same logic instead of "just consuming" the backend. This may introduce subtle errors which cannot be fixed by just fixing the backend. Should we put a comment like "can be removed when the backend is fixed" here?

src/Data/Order/OrderEditingConfig.ts
```diff
@@ -6,12 +6,20 @@ provideEditingConfig(Order, {
   attributes: {
     keyword: { title: 'Keyword' },
     number: { title: 'Number' },
-    type: { title: 'Type' },
-    status: { title: 'Status' },
+    type: { title: 'Type (enum)' },
+    typeLocalized: { title: 'Type (localized)' },
+    status: { title: 'Status (emum)' },
```
dcsaszar, line 11: * [ ] Typo emum -> enum (but see other comment)

src/Data/Order/OrderEditingConfig.ts
```diff
@@ -6,12 +6,20 @@ provideEditingConfig(Order, {
...
+    statusLocalized: { title: 'Status (localized' },
```
dcsaszar, line 12: * [ ] Typo missing parenthesis
* [ ] Same issue is present elsewhere

src/Data/Quote/QuoteDataClass.ts
```diff
@@ -6,272 +6,350 @@ export const Quote = provideLocalStorageDataClass('Quote', {
...
       customer: 'Maier Schmiertechnik GmbH (München)',
       commercialAgent: '052601BEBCEC39C8E040A8C00D0107AC',
       technicalAgent: 'F87BDC400E41D630E030A8C00D01158A',
       salesPartner: '',
-      status: 'quote lost',
-      mainStatus: 'Quote lost',
+      status: 'PSA_PRO_QUO_LST',
+      statusLocalized: 'quote lost',
+      mainStatus: 'PSA_SAP_MAI_QUO_LST',
+      mainStatusLocalized: 'Quote lost',
```
dcsaszar, line 315: * [ ] FYI lower/upper case inconsistency (was already present before)

dcsaszar (changes requested):
See comments. Content looks good to me, but I only checked the main lists and one item each and don't really know what to watch out for.

src/Data/Ticket/TicketDataClass.ts
apepper: This code is only intended as a quickfix for the localstorage to work. With a real backend you would not use such code.

src/Data/Quote/QuoteDataClass.ts
apepper: I'm not sure, what you mean. Do you men `mainStatus` is wrongly cases? Should it be `mainstatus` instead?

apepper:
> Content looks good to me, but I only checked the main lists and one item each and don't really know what to watch out for.

The main change is for all "enum" attributes, that most (but not all) places in the content should used `xLocalized` instead of `x`. So a regular visitor should not see something like `PSA_QUO_STY_GEN_OVR` or `PSA_INF`.

src/utils/provideLocalStorageDataClass.ts
apepper: I would like to keep this code specific for boolean attributes.

src/Data/Quote/QuoteDataClass.ts
dcsaszar: statusLocalized: '**q**uote lost',
mainStatusLocalized: '**Q**uote lost',

dcsaszar (approved):

PR #302, opened 2024-02-14
FIX for underline, selection, text-multiline, comment, btn

agessler:
under a link only in useful area, selection color update, text-multiline fix, comment avatar size fix, search result btn color fix

### DIFF
--- a/src/assets/stylesheets/index.scss
+++ b/src/assets/stylesheets/index.scss
@@ -100,6 +100,7 @@
   color: var(--bs-btn-hover-color);
   background-color: var(--bs-btn-hover-bg);
   border-color: var(--bs-btn-hover-border-color);
+  text-decoration: none;
 }
 .btn-check + .btn:hover {
   color: var(--bs-btn-color);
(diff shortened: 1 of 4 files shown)


src/assets/stylesheets/index.scss
```diff
@@ -405,26 +406,36 @@ div {
 
 a {
   color: var(--bs-primary);
-  text-decoration: none;
+  text-decoration: underline;
```
apepper, line 409: I added "Text with a link" widgets on several places in https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://agessler-css-fixes.scrivito-portal-app.pages.dev/colors?_scrivito_workspace_id=q6a733838df01858&_scrivito_display_mode=view . Some combinations still don't show an underline (e.g. a card within a card). Please investigate.

apepper (changes requested):
FEEDBACK ([16e3f86](https://github.com/Scrivito/scrivito-portal-app/pull/302/commits/16e3f867573bffc88c721399e8a6d2863c6232fa))

Nice 🚀!

But there are still some non-underlined links (see comments).

src/assets/stylesheets/index.scss
agessler: Its not easy. We are using card nearly everywhere. In some cases we don't want underline in person cards in "real" content it could be nice. Not easy to see codewise when what to do.

apepper: I think the fix to the issues is to change the following

https://github.com/Scrivito/scrivito-portal-app/blob/16e3f867573bffc88c721399e8a6d2863c6232fa/src/assets/stylesheets/index.scss#L428

to `.card > a {`

So with that, the "underline of a card, that is a link" is only limited to the top link and not all inner links as well.

apepper: Please try this out and update the PR if it works.

apepper: Will be addressed in a follow up PR.

PR #307, opened 2024-02-19
Neoletter profile wins

apepper:
The data for the current user should come mainly from the neoletter profile service. Only the e-mail and profile picture should be taken from IAM. Only very limited information should be taken from Pisa (actual code for that in a different branch).

Needs content changes: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://neoletter-profile-wins.scrivito-portal-app.pages.dev/my-profile-0e84d1e469f7cfac?current_user_id=0&_scrivito_workspace_id=u6ea13b19f0b6fc9&_scrivito_display_mode=view

### DIFF
--- a/src/Data/CurrentUser/CurrentUserDataItem.ts
+++ b/src/Data/CurrentUser/CurrentUserDataItem.ts
@@ -6,53 +6,73 @@ import {
   unstable_JrRestApi,
 } from 'scrivito'
 import personCircle from '../../assets/images/person-circle.svg'
+import { ensureString } from '../../utils/ensureString'
 
 export const CurrentUser = provideDataItem('CurrentUser', {
   async get() {
     const user = await load(currentUser)
     if (!user) return {}
 
-    const rawNeoletterData = await unstable_JrRestApi.get(neoletterProfileUrl())
-
-    let company: string | undefined
-    let phoneNumber: string | undefined
-    let salutation: string | undefined
-    const pisaUserId: string | undefined = 'F87BDC400E41D630E030A8C00D01158A'
-    const picture: string = user.picture() || personCircle
-
-    if (isNeoletterData(rawNeoletterData)) {
-      company = rawNeoletterData.company || ''
-      phoneNumber = rawNeoletterData.phone_number || ''
-      salutation = rawNeoletterData.salutation || ''
+    const neoletterProfile = await unstable_JrRestApi.get(neoletterProfileUrl())
+    if (!isNeoletterData(neoletterProfile)) {
+      throw new Error('Neoletter data is not in the expected format')
     }
 
+    // TODO: read data from Pisa
+    const pisaUserId: string = 'F87BDC400E41D630E030A8C00D01158A'
+    const salesUserId: string = '052601BEBCEC39C8E040A8C00D0107AC'
+    const serviceUserId: string = 'D456ACF6FF405922E030A8C02A010C68'
+
     return {
-      company,
       email: user.email(),
+      picture: ensureString(user.picture()) || personCircle,
       jrUserId: user.id(),
+
       pisaUserId,
-      name: user.name(),
-      phoneNumber,
-      picture,
-      salesUserId: '052601BEBCEC39C8E040A8C00D0107AC', // TODO: replace with real Pisa data
-      salutation,
-      serviceUserId: 'D456ACF6FF405922E030A8C02A010C68', // TODO: replace with real Pisa data
+      salesUserId,
+      serviceUserId,
+
+      // TODO: Remove user.name() fallback once the name is always available in Neoletter
+      name: ensureString(neoletterProfile.name) || user.name(),
+      company: ensureString(neoletterProfile.company),
+      familyName: ensureString(neoletterProfile.family_name),
+      givenName: ensureString(neoletterProfile.given_name),
+      phoneNumber: ensureString(neoletterProfile.phone_number),
+      salutation: ensureString(neoletterProfile.salutation),
     }
   },
   async update(params) {
-    const { company, phoneNumber, salutation, ...otherArgs } = params
+    const {
+      company,
+      familyName,
+      givenName,
+      name,
+      phoneNumber,
+      salutation,
+      ...otherArgs
+    } = params
     if (Object.keys(otherArgs).length > 0) {
       throw new Error(`Unknown keys - ${Object.keys(otherArgs)}`)
     }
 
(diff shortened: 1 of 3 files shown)


src/Data/CurrentUser/CurrentUserDataItem.ts
```diff
@@ -6,53 +6,73 @@ import {
...
+
       pisaUserId,
-      name: user.name(),
-      phoneNumber,
-      picture,
-      salesUserId: '052601BEBCEC39C8E040A8C00D0107AC', // TODO: replace with real Pisa data
-      salutation,
-      serviceUserId: 'D456ACF6FF405922E030A8C02A010C68', // TODO: replace with real Pisa data
+      salesUserId,
+      serviceUserId,
```
dcsaszar, line 33: * [ ] salesUserId and serviceUserId are not part of provideEditingConfig(CurrentUser). It's not clear why they are here.

src/Data/User/UserDataClass.ts
```diff
@@ -1,8 +1,25 @@
 import { provideLocalStorageDataClass } from '../../utils/provideLocalStorageDataClass'
 import richterImage from './FakeBinaries/richter.jpg'
 import braschauImage from './FakeBinaries/braschau.jpg'
+import { CurrentUser } from '../CurrentUser/CurrentUserDataItem'
```
dcsaszar, line 4: * [ ] FYI depending on the Scrivito data item is reusing one abstraction layer too many, which makes the code harder to understand and less isolated (the dependency on `Scrivito.load` was the yellow flag for me).

src/Data/User/UserDataClass.ts
```diff
@@ -1,8 +1,25 @@
...
+import { load } from 'scrivito'
 
 export const User = provideLocalStorageDataClass('User', {
+  postProcessData: async (data) => {
+    const pisaUserId = await load(() => CurrentUser.get('pisaUserId'))
+    if (data._id !== pisaUserId) return data
```
dcsaszar, line 10: * [ ] FYI Only processing the user without also processing the list of users looks like there's a piece missing. Shouldn't we also inject the user to the index? Maybe an explanatory comment is missing?

dcsaszar (changes requested):
Looks good. See comments.

src/Data/User/UserDataClass.ts
apepper: `postProcessData` is used for all `index`, `get` and `update` functions. See https://github.com/Scrivito/scrivito-portal-app/pull/305

src/Data/User/UserDataClass.ts
apepper: What do you suggest? Extract the `get` method of `CurrentUserDataItem` and call that method directly?

src/Data/CurrentUser/CurrentUserDataItem.ts
apepper: Good catch! They are currently hard-coded in `PersonCardWidget` but we might switch this over to a datalocator.

src/Data/User/UserDataClass.ts
dcsaszar: Yes, that sounds good.

src/Data/User/UserDataClass.ts
dcsaszar: Yes, but I read the code above as if it post processes a GET for user with a certain ID, without post processing the INDEX to make sure the user is present. If my observation is correct, my question is: how does the system learn about the user?

src/Data/User/UserDataClass.ts
apepper: > Extract the `get` method of `CurrentUserDataItem` and call that method directly?

If we do that we would have to fetch the current user information on every single user call - for get but also for index. If we use `CurrentUserDataItem` instead we benefit from Scrivitos caching.

apepper:
REVIEW ([0085dd4](https://github.com/Scrivito/scrivito-portal-app/pull/307/commits/0085dd42888e82b5bc98a4be3fe56b8a4d79f92d))

src/Data/User/UserDataClass.ts
dcsaszar: > If we use CurrentUserDataItem instead we benefit from Scrivitos caching.

I see. So we trade a drawback (depending on Scrivito.load) for an advantage (profit from Scrivito caching). Worth a comment?

dcsaszar (approved):

src/Data/User/UserDataClass.ts
apepper: > If my observation is correct, my question is: how does the system learn about the user?

Pisa knows about the current user - otherwise you would not be able to communicate with Pisa. So Pisas `whoami` endpoint will provide you with a pisa user ID. The `user` endpoint of pisa understands this user ID and returns the data for that user.

apepper: But Pisa will take care of this. No need to inject it manually.

src/Data/User/UserDataClass.ts
dcsaszar: How about:

```
// Use the data item cached by Scrivito instead of connecting to the backend directly
```

src/Data/User/UserDataClass.ts
dcsaszar: I still don't get it. The users are provided using `provideLocalStorageDataClass`. Pisa is not involved in generating the list of users.

apepper: `provideLocalStorageDataClass('User'` is only meant as a stand-in for a real pisa instance, so we can develop fairly independent of an actual instance running.

The fake user data has this:

https://github.com/Scrivito/scrivito-portal-app/blob/effb95937f70c97975e626fc2d6aaa2e75e382ef/src/Data/User/UserDataClass.ts#L71-L78

And the partially fake current user info has this:

https://github.com/Scrivito/scrivito-portal-app/blob/effb95937f70c97975e626fc2d6aaa2e75e382ef/src/Data/CurrentUser/CurrentUserDataItem.ts#L23

With that "Markus Neumann" becomes the currently logged in user with the information provided by IAM and Neoletter.

dcsaszar: Thank! The hardcoded `F87BDC400E41D630E030A8C00D01158A` was the missing link. I had actually eyeballed the fake users expecting to find the ID there but overlooked it. A comment (or a constant) would have helped.

PR #310, opened 2024-02-20
[design] Fix links in nested card widgets

apepper:
As discussed in https://github.com/Scrivito/scrivito-portal-app/pull/302#discussion_r1492509298.

See it live on https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://fix-nested-links.scrivito-portal-app.pages.dev/colors?_scrivito_workspace_id=q6a733838df01858&_scrivito_display_mode=view

### DIFF
--- a/src/assets/stylesheets/index.scss
+++ b/src/assets/stylesheets/index.scss
@@ -425,7 +425,7 @@ nav a,
 .nav a,
 .navbar a,
 .nav-bordered a,
-.card a {
+.card > a {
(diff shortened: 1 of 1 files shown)


agessler (approved):
We can do it this way. The point was is when you use a card with avarage content you might need a underline of a text link. The code before adresse just when a link is used as wrapper - in use at internal area data lists.

PR #319, opened 2024-03-01
DataEmptyWidget

apepper:
Live preview: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dataemptywidget.scrivito-portal-app.pages.dev/my-tynacoon?_scrivito_workspace_id=h67a0c6846936371&_scrivito_display_mode=view

We want to move into a direction where the "DataListWidget" no longer renders an "empty case". Instead this should be handled in a different Widget - the `DataEmptyWidget`. This is the first (but not final) step in this direction.


src/Widgets/DataEmptyWidget/DataEmptyWidgetComponent.tsx
```diff
@@ -0,0 +1,50 @@
...
+        ) : (
+          <button
+            className="btn btn-sm btn-outline-secondary"
+            onClick={() => setShowAnyway(true)}
+          >
+            Show content anyway
+          </button>
+        )}
+      </EditorNoteOrNull>
+      {showAnyway ? <ContentTag content={widget} attribute="content" /> : null}
```
dcsaszar, line 47: * [ ] FYI I'm not convinced the "show anyway" UX is what we want. Given that
  * The message + button take up more space than a typical "no items" message, so it often *increases* UI clutter instead of reducing it
  * It requires switching to preview for WYSIWYG anyway
  
  we could simply always show the content in edit mode. The conditional nature could be highlighted by one or more of
  * opacity
  * outline
  * background
  * widget label

src/Widgets/DataEmptyWidget/DataEmptyWidgetEditingConfig.ts
```diff
@@ -0,0 +1,12 @@
+import { provideEditingConfig } from 'scrivito'
+import { DataEmptyWidget } from './DataEmptyWidgetClass'
+import Thumbnail from './thumbnail.svg'
+import { TextWidget } from '../TextWidget/TextWidgetClass'
+
+provideEditingConfig(DataEmptyWidget, {
+  title: 'Data Empty',
```
dcsaszar, line 7: * [ ] FYI It would be great to have a more self-explanatory title. Sth. like "If no data", "Missing data fallback", "Fallback".

dcsaszar (changes requested):
I like the general idea. See comments. Also:

* [ ] If I configure a data source to return quantity: 0 or a non-matching filter, the fallback is not shown.

src/Components/EditorNoteOrNull.tsx
```diff
@@ -0,0 +1,15 @@
+import { isInPlaceEditingActive } from 'scrivito'
+
+/** Shows `children` nicely formatted in edit mode, `null` overwise */
+export function EditorNoteOrNull({ children }: { children: React.ReactNode }) {
```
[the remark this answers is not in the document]
apepper: FYI: I took `EditorNote`.

src/Widgets/DataFormContainerWidget/DataFormContainerWidgetComponent.tsx
```diff
@@ -64,6 +70,7 @@ provideComponent(DataFormContainerWidget, ({ widget }) => {
         await dataItem.update(attributes)
         await toastAndRedirect(dataItem)
       } else {
+        if (!dataScope) throw new Error('No data scope found!')
```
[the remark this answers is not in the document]
apepper: This error is mainly there to satisfy TypeScript. The function `onSubmit` is always defined, but the component code only call it, if `dataItem` or `dataScope` is present.

src/Widgets/DataEmptyWidget/DataEmptyWidgetComponent.tsx
apepper: Nice idea with the `opacity`.

dcsaszar (approved):

PR #324, opened 2024-03-05
DataColumnListWidget

apepper:
Extracts `nrOfColumns` option of `DataListWidget` into `DataColumnListWidget`. Motivation: There can be potentially many different ways to render a datalist. Instead of a "one-widget-to-rule-them-all" each list widget should be specialized.

thumbnail

Live preview: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://datacolumnlistwidget.scrivito-portal-app.pages.dev/my-tynacoon?_scrivito_workspace_id=o307f510c3f42ae9&_scrivito_display_mode=view


dcsaszar (approved):
Code-wise LGTM.

> Instead of a "one-widget-to-rule-them-all" each list widget should be specialized.

I can't review this part since we haven't specified the boundaries between "it should be an option" and "it should be a widget". In general I'm opposed to having two widgets for stylistic options the editor should be able to play with. The current state of the two widgets here falls into this category. But I don't know about the future plans, so I guess I don't see the full picture.

PR #328, opened 2024-03-07
DataLabelWidget: Add "link" format

apepper:
Motivation: The upcoming "event" will have a `url` field. With this option the URL becomes clickable.

### DIFF
--- a/src/Widgets/DataLabelWidget/DataLabelWidgetClass.ts
+++ b/src/Widgets/DataLabelWidget/DataLabelWidgetClass.ts
@@ -5,7 +5,7 @@ export const DataLabelWidget = provideWidgetClass('DataLabelWidget', {
     label: 'string',
     attributeName: 'string',
     details: 'string',
-    showAs: ['enum', { values: ['text', 'currency', 'datetime', 'url'] }],
+    showAs: ['enum', { values: ['text', 'currency', 'datetime', 'link'] }],
     valueSize: [
       'enum',
       {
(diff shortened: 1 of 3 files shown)


src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
```diff
@@ -84,3 +85,13 @@ function Datetime({ value }: { value: unknown }) {
 
   return <RelativeDate date={date} />
 }
+
+function URL({ value }: { value: unknown }) {
```
dcsaszar, line 89: * [ ] FYI For suffixes we use camel instead of all uppercase: `*Id` not `*ID`, `*Url` not `*URL`. Thus `Url` would be the more consistent casing here.

src/Widgets/DataLabelWidget/DataLabelWidgetClass.ts
```diff
@@ -5,7 +5,7 @@ export const DataLabelWidget = provideWidgetClass('DataLabelWidget', {
     label: 'string',
     attributeName: 'string',
     details: 'string',
-    showAs: ['enum', { values: ['text', 'currency', 'datetime'] }],
+    showAs: ['enum', { values: ['text', 'currency', 'datetime', 'url'] }],
```
dcsaszar, line 8: * [ ] FYI It's not just rendered as a URL but as a link with the URL as linked text. Maybe the more important "show-as" aspect is `link` not `url`?

dcsaszar (approved):
Nice, see comment for some alternatives.

PR #329, opened 2024-03-11
Enum data localization (pt 1: quotes)

dcsaszar:
Successor of #322

Working copy: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dave-new-enum-localization.scrivito-portal-app.pages.dev/quotes?_scrivito_workspace_id=hfaa80b41e53c154&_scrivito_display_mode=diff


src/utils/valuesConfig.ts
```diff
@@ -0,0 +1,59 @@
+import { DataClass, DataItem } from 'scrivito'
+import { ensureString } from './ensureString'
+
+const values: Partial<Record<string, Partial<Record<string, string[]>>>> = {}
+const localizers: Partial<
+  Record<string, Partial<Record<string, Partial<Record<string, string>>>>>
+> = {}
```
apepper, line 7: This works, but is hard to understand. Could you maybe add more intermediate type/interfaces to help to improve understanding?

src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
```diff
@@ -60,7 +61,11 @@ const AttributeValue = connect(function AttributeValue({
   if (showAs === 'currency') return <Currency value={attributeValue} />
   if (showAs === 'datetime') return <Datetime value={attributeValue} />
 
-  return <Text value={attributeValue} />
+  if (!dataItem) return <Text value={attributeValue} />
```
apepper, line 64: Let's move the `!dataItem` check to the top of the function and return `'N/A'` right away.

src/utils/valuesConfig.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  // We can use `value!` b/c its type depends on typeof data
+  return localizers[data.name()]?.[attributeName]?.[value!] || value!
+}
+
+function isDataItem(item: unknown): item is DataItem {
+  if (!item || typeof item !== 'object') return false
+  return typeof (item as DataItem).id === 'function'
+}
```
apepper, line 59: This code looks quite clever and I have a hard time to understand it.

As far as I can see, the currently only caller, that is with `dataItem` already read out the value previously. With that I suggest to simplify this function to only the `dataClass` case (including using named attributes and using present tense):

```suggestion
export function localizeAttributeValue({
  dataClass,
  attributeName,
  attributeValue,
}: {
  dataClass: DataClass
  attributeName: string
  attributeValue: string
}): string {
  return (
    localizers[dataClass.name()]?.[attributeName]?.[attributeValue] ||
    attributeValue
  )
}
```

Usage in `DataLabelWidgetComponent`:

```ts
  const value = localizeAttributeValue({
    dataClass: dataItem.dataClass(),
    attributeName,
    attributeValue: ensureString(attributeValue),
  })

  return <Text value={value} />
```

src/utils/valuesConfig.ts
```diff
@@ -0,0 +1,59 @@
+import { DataClass, DataItem } from 'scrivito'
```
apepper, line 1: I would rename this file to `dataValuesConfig.ts` to make it clearer, which values are ment.

src/utils/valuesConfig.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (isDataItem(data)) {
+    return localizedAttributeValue(
+      data.dataClass(),
+      attributeName,
+      ensureString(data.get(attributeName)),
+    )
+  }
+
+  // We can use `value!` b/c its type depends on typeof data
+  return localizers[data.name()]?.[attributeName]?.[value!] || value!
```
apepper, line 53: The `|| value!` has the consquence, that a localizer, that is explicitly returning `''` would still return the original enum value. E.g. for the `termsOfDelivery: 'RFC_TOD_FH'` this function would return `'RFC_TOD_FH'` instead of `''`, even though it is configured to be `''`.

I think we should use `?? value!` instead.

apepper (changes requested):
FEEDBACK ([3d2c9c9](https://github.com/Scrivito/scrivito-portal-app/pull/329/commits/3d2c9c9e956c68a18e027741f595a6f99d2b004a))

Nice 🚀. Really nice progress compared to the first version!

Please see my comments to take this PR even further.

src/utils/valuesConfig.ts
dcsaszar: FYI If we'd enable `noUncheckedIndexedAccess` we could drop the `Partial` part. But maybe this makes other areas of the codebase harder to maintain.

src/Data/Quote/QuoteValuesConfig.ts
```diff
@@ -0,0 +1,161 @@
...
+    RFC_TOD_FAS: 'free alongside ship',
+    RFC_TOD_FCA: 'free carrier',
+    RFC_TOD_FH: '',
+    RFC_TOD_FOB: 'free on board',
+    RFC_TOD_UN: '',
+  },
+  termsOfPayment: {
+    PSA_TOP_10: 'cash payable within 14 days',
+    PSA_TOP_20: 'cash',
+    PSA_TOP_50: 'by C.O.D.',
```
apepper, line 144: `PSA_TOP_40` seems to be missing from the list. Could you check, if there are maybe other values missing as well?

apepper (commented):
FEEDBACK ([deaa2b0](https://github.com/Scrivito/scrivito-portal-app/pull/329/commits/deaa2b0f7c3893c2b5ac69d236aa95f3d36c41c8))

Very nice 🚀! Almost there.

Please see my comment.

src/Data/Quote/QuoteValuesConfig.ts
dcsaszar: You misread: `PSI_TOP_40` line 150

apepper: I can see this value on the following URL: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dave-new-enum-localization.scrivito-portal-app.pages.dev/quote-details-0092ff67875b69f8?quote_id=4E1FEAAC57E7DFC7E040007F010032F5&_scrivito_workspace_id=hfaa80b41e53c154&_scrivito_display_mode=diff

Screenshot 2024-03-18 at 09 39 32

Please investigate.

apepper: The value `PSA_TOP_40` is clearly in use, e.g. on https://github.com/Scrivito/scrivito-portal-app/pull/329/files#diff-a0bc8e033a6d719f20539e57301bc559e7a181e8729ac9262efdd95795af5d15R21.

dcsaszar: Clarified. The value `PSA_TOP_40` is invalid. I fixed the data. REVIEW

apepper (approved):
LGTM ([eeec4c3](https://github.com/Scrivito/scrivito-portal-app/pull/329/commits/eeec4c39c1412b8579afc1203692f7b70258eb5f))

Nice 🚀!

Some "translations" look a bit strange (e.g. `[00] ERP Payment target`), but this is more a PisaSales content issue.

PR #331, opened 2024-03-13
DataLabelWidget: More date options

apepper:
Allows to format a date value as "Date and time" and "Only date". "Relative" was previously already available.

Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://datalabelwidget-date.scrivito-portal-app.pages.dev/my-tynacoon?_scrivito_workspace_id=mc2dbbd39b44e23a&_scrivito_display_mode=view

### DIFF
--- a/src/Widgets/DataLabelWidget/DataLabelWidgetEditingConfig.ts
+++ b/src/Widgets/DataLabelWidget/DataLabelWidgetEditingConfig.ts
@@ -1,6 +1,7 @@
 import { provideEditingConfig } from 'scrivito'
 import { DataLabelWidget } from './DataLabelWidgetClass'
 import Thumbnail from './thumbnail.svg'
+import { isString } from 'lodash-es'
 
 provideEditingConfig(DataLabelWidget, {
   title: 'Data Label',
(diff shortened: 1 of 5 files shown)


src/Widgets/DataLabelWidget/DataLabelWidgetEditingConfig.ts
```diff
@@ -17,11 +18,27 @@ provideEditingConfig(DataLabelWidget, {
...
   },
-  properties: ['attributeName', 'showAs', 'valueSize', 'marginBottom'],
+  properties: (widget) =>
+    [
+      'attributeName',
+      'showAs',
+      widget.get('showAs') === 'datetime' ? 'datetimeFormat' : null,
+      'valueSize',
+      'marginBottom',
+    ].filter(isString),
```
dcsaszar, line 41: * [ ] FYI I'm skeptical about adding dependencies for trivial stuff like `].filter(p => p !== null),`

src/utils/formatDate.ts
```diff
@@ -0,0 +1,39 @@
+const dateTimeFormatLocale = 'en-GB'
+
+export function formatDateTime(date: Date) {
```
dcsaszar, line 3: * [ ] FYI Refactoring idea: Export a type for available formats and provide a single helper function.

```ts
export type DateFormat = 'dateTime' | 'fullDateTime' // | etc
export function formatDate(date: Date, format: DateFormat)
```

dcsaszar (approved):
LGTM

src/Widgets/DataLabelWidget/DataLabelWidgetEditingConfig.ts
apepper: I already tried several things, e.g.

```ts
  properties: (widget): string[] =>
    [
      'attributeName',
      'showAs',
      widget.get('showAs') === 'datetime' ? 'datetimeFormat' : null,
      'valueSize',
      'marginBottom',
    ].filter((p) => p !== null),
```

or
```ts
  properties: (widget): string[] =>
    [
      'attributeName',
      'showAs',
      widget.get('showAs') === 'datetime' ? 'datetimeFormat' : null,
      'valueSize',
      'marginBottom',
    ].filter((p) => typeof p === 'string'),
```

All of these still print:

```
Type '(string | null)[]' is not assignable to type 'string[]'.
Type 'string | null' is not assignable to type 'string'.
Type 'null' is not assignable to type 'string'.
```

With `isString` (or probably some other guard function) it works.

dcsaszar: I see, TS stuff.
The cleanest I can think of right now is an inline type guard:
```ts
].filter((p): p is string => typeof p === 'string'),
```

apepper: => https://github.com/Scrivito/scrivito-portal-app/pull/333

PR #332, opened 2024-03-13
Events

apepper:
Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://events.scrivito-portal-app.pages.dev/my-tynacoon?_scrivito_workspace_id=vc5d7cb7ea655b19&_scrivito_display_mode=view

PR #338, opened 2024-03-18
DataFormContainer: Work properly inside of links

apepper:
E.g. clicking on the submit button propagates to an upper link, if the propagation is not stopped.

Follow-up of https://github.com/Scrivito/scrivito-portal-app/pull/332#discussion_r1527621580.

Example Content (**not to be published!**): https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dataformcontainer-bugfix.scrivito-portal-app.pages.dev/events-fb302b23c46fea41?_scrivito_workspace_id=v667419e005c44b6&_scrivito_display_mode=view

### DIFF
--- a/src/Widgets/DataFormContainerWidget/DataFormContainerWidgetComponent.tsx
+++ b/src/Widgets/DataFormContainerWidget/DataFormContainerWidgetComponent.tsx
@@ -44,6 +44,9 @@ provideComponent(DataFormContainerWidget, ({ widget }) => {
         onSubmit={onSubmit}
         onReset={onReset}
         className={isSubmitting ? 'form-loading' : ''}
+        onClick={(e) => {
+          e.stopPropagation()
+        }}
       >
         <InPlaceEditingOff>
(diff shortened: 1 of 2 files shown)


src/Widgets/DataFormContainerWidget/DataFormContainerWidgetComponent.tsx
```diff
@@ -44,6 +44,9 @@ provideComponent(DataFormContainerWidget, ({ widget }) => {
         onSubmit={onSubmit}
         onReset={onReset}
         className={isSubmitting ? 'form-loading' : ''}
+        onClick={(e) => {
```
dcsaszar, line 47: * [ ] I'm surprised stopping propagation for all clicks on the form works as intended when clicking on text inside the form. However, it doesn't stop the click propagation of form input fields. So it appears to apply to buttons only. How about making the intended effect clearer by stopping the propagation on the button component?

dcsaszar (changes requested):
Looks good but can be simplified.

PR #341, opened 2024-03-18
Remove Headline from DataPersonCardWidget

apepper:
It's better to leave this to a regular HeadlineWidget with all its options.

Needs content - please also check the event details page: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://simpler-datapersoncardwidget.scrivito-portal-app.pages.dev/my-tynacoon?_scrivito_workspace_id=p81a8ca32bdb5135&_scrivito_display_mode=view


dcsaszar (changes requested):
Code approved.
Content changes:
* [ ] Event details page: The white box underneath the event with additional contact details made sense and looked good. The new design has a white box in a white box, which looks off. Could we restore some version of the previous event details page?

apepper:
> Could we restore some version of the previous event details page?

I reverted the changes to the events details page.

PR #348, opened 2024-03-22
Use protected content inside portal

markus-codewrangler:
- Port [RedirectObj from example app](https://github.com/Scrivito/scrivito_example_app_js/tree/master/src/Objs/Redirect) to portal app.
- Extend RedirectObj in portal app with "ensureUserIsLoggedIn" boolean flag - enforces a log-in before redirecting
- Create a "RedirectObj" pointing to "My Tynacoon". After that change all "My Tynacoon" pages to "protected content" and remove "requireUserLogin" flag

[WC with site visibility adjustments](https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~http://localhost:8080/my-tynacoon?_scrivito_workspace_id=k11311614e5c25b5&_scrivito_display_mode=diff)


src/Objs/Redirect/RedirectComponent.tsx
```diff
@@ -18,21 +18,17 @@ class RedirectComponent extends React.Component {
         window.location.replace(url)
       }
     })
-  }
-
-  render() {
-    const link = this.props.page.get('link')
+  }, [page])
```
dcsaszar, line 22: * [ ] FYI we could move the assignment to `link` up and use `link` for the dependency instead of `page`. Maybe even use `url` as it best explains the actual effect to the reader.

src/Objs/Redirect/RedirectComponent.tsx
```diff
@@ -1,19 +1,20 @@
...
+      const url = link && urlFor(link)
 
       return { link, url }
     }).then(({ link, url }) => {
-      if (!link) return
+      if (!link || !url) return
 
       if (link.isExternal()) {
-        window.top.location.replace(url)
+        window.top?.location.replace(url)
```
dcsaszar, line 17: * [ ] Not sure under which circumstances `window.top` is `null`. Maybe falling back to the `else` block is the better strategy?

src/Objs/Redirect/RedirectObjClass.ts
```diff
@@ -4,5 +4,6 @@ export const Redirect = provideObjClass('Redirect', {
   attributes: {
     title: 'string',
     link: 'link',
+    ensureUserIsLoggedIn: 'boolean',
```
dcsaszar, line 7: * [ ] FYI naming: Looking at other booleans we prefer `require*` for switches like this (we're not fully consistent though). How about `requireLogin` or `loginRequired`?

src/Objs/Redirect/RedirectEditingConfig.ts
```diff
@@ -13,8 +13,13 @@ provideEditingConfig('Redirect', {
     link: {
       title: 'Link',
     },
+    ensureUserIsLoggedIn: {
+      title: 'Ensure user is logged in',
```
dcsaszar, line 17: * [ ] FYI we typically end boolean titles with a question mark.

src/Objs/Redirect/RedirectComponent.tsx
```diff
@@ -21,6 +21,8 @@ provideComponent(Redirect, ({ page }) => {
     })
   }, [page])
 
+  if (page.get('ensureUserIsLoggedIn')) ensureUserIsLoggedIn()
```
dcsaszar, line 24: * [ ] From the method name I'd assume this call has a side effect, so we should put it in an effect. 
* [ ] We must make sure not to create a race between `ensureUserIsLoggedIn()` and `location.replace`, because a reload will undo whatever `ensureUserIsLoggedIn` triggers. I think a fix would include branching depending on `isUserLoggedIn()`.

dcsaszar (changes requested):
Most parts and content changes look good. See comments.

src/Objs/Redirect/RedirectComponent.tsx
markus-codewrangler: I cleaned it up for good order.

FYI this is what Copilot has to say (not checked):

> In a typical browser environment, window.top should never be null because even if the window isn't part of a frameset, window.top will simply return a reference to the window itself.
> 
> However, TypeScript is designed to be safe and assumes that any property of the window object could potentially be null. This is because TypeScript can be used in a variety of environments, not just browsers, and in some of these environments, the window object or its properties might not be available.
> 
> In practice, if you're running your code in a browser environment and you're not doing anything unusual with the window object, you can safely assume that window.top will never be null.

src/Objs/Redirect/RedirectComponent.tsx
markus-codewrangler: I moved the assignments up. Now we have 3 dependencies instead. If you have ideas for improvement, let me know 🙂

src/Objs/Page/PageLayoutComponent.tsx
```diff
@@ -1,23 +1,6 @@
...
-      return (
-        <div className="text-center">
-          <div className="loading-placeholder" />
-        </div>
-      )
-    }
-  }
-
+provideLayoutComponent(Page, () => {
   return <CurrentPage />
```
apepper, line 5: Given that this LayoutComponent now only renders the current page, we can remove the `PageLayoutComponent` completely.

src/Objs/Redirect/RedirectObjClass.ts
```diff
@@ -0,0 +1,9 @@
+import { provideObjClass } from 'scrivito'
+
+export const Redirect = provideObjClass('Redirect', {
```
apepper, line 3: Please add `Redirct` to `BLACKLIST_OBJ_CLASSES` - so that it will not show up in search results.

src/Objs/Redirect/RedirectComponent.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { InPlaceEditingPlaceholder } from '../../Components/InPlaceEditingPlaceholder'
+import { Redirect } from './RedirectObjClass'
+
+provideComponent(Redirect, ({ page }) => {
+  const requireLogin = page.get('requireLogin')
+  const link = page.get('link')
+  const url = link && urlFor(link)
+
+  React.useEffect(() => {
+    if (requireLogin && !isUserLoggedIn) {
```
apepper, line 17: If I'm not mistaken, this is a function, not a constant:

```suggestion
    if (requireLogin && !isUserLoggedIn()) {
```

src/Objs/Redirect/RedirectComponent.tsx
```diff
@@ -0,0 +1,40 @@
...
+  provideComponent,
+  urlFor,
+} from 'scrivito'
+import { InPlaceEditingPlaceholder } from '../../Components/InPlaceEditingPlaceholder'
+import { Redirect } from './RedirectObjClass'
+
+provideComponent(Redirect, ({ page }) => {
+  const requireLogin = page.get('requireLogin')
+  const link = page.get('link')
+  const url = link && urlFor(link)
```
apepper, line 14: I think it would be better to move this to `useEffect` and use load instead. Otherwise the effect might be trigger multiple times, if a newer version of the URL (e.g. with a full slug, after loading the actual obj) is available.

src/Objs/Redirect/RedirectComponent.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  if (!link) {
+    return (
+      <InPlaceEditingPlaceholder center>
+        Select a link in the redirect page properties.
+      </InPlaceEditingPlaceholder>
+    )
+  }
+
+  return null
```
apepper, line 39: I think it would be good to show a note for `isEditorLoggedIn()` and _not_ redirect - similar to the `Dropdown` obj component (e.g. https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://markus-protects-content.scrivito-portal-app.pages.dev/about-us-9db36de70dd4d157?_scrivito_workspace_id=k11311614e5c25b5&_scrivito_display_mode=view).

apepper (changes requested):
FEEDBACK ([bf54ede](https://github.com/Scrivito/scrivito-portal-app/pull/348/commits/bf54edeef020bcdb47c5501b7a23359ccbd57166))

Not a full review, but some comments to improve this further.

src/Objs/Redirect/RedirectComponent.tsx
markus-codewrangler: Ok. But I think it doesn't make a big difference compared to `InPlaceEditingPlaceholder` - apart from the nicer format. Is that where you're coming from?

markus-codewrangler:
Please review, whoever will continue to review this 😁

src/Objs/Redirect/RedirectComponent.tsx
dcsaszar: After the changes in 87364a7 this is still not optimal. In the UI the redirects should never be followed but always be informative only.

src/Objs/Redirect/RedirectComponent.tsx
```diff
@@ -11,22 +12,23 @@ import { Redirect } from './RedirectObjClass'
...
-  const url = link && urlFor(link)
 
   React.useEffect(() => {
     if (requireLogin && !isUserLoggedIn()) {
       ensureUserIsLoggedIn()
       return
     }
 
-    if (!link || !url) return
+    load(() => link && urlFor(link)).then((url) => {
```
dcsaszar, line 22: * [ ] FYI For simpler code  the `!link` condition could be a guard at the beginning of the effect callback

src/Objs/Redirect/RedirectComponent.tsx
```diff
@@ -30,11 +30,19 @@ provideComponent(Redirect, ({ page }) => {
...
     return (
-      <InPlaceEditingPlaceholder center>
-        Select a link in the redirect page properties.
-      </InPlaceEditingPlaceholder>
+      <div className="container d-flex h-100">
+        <div className="alert alert-warning d-flex m-auto">
+          <i className="bi bi-exclamation-circle bi-2x" aria-hidden="true"></i>
+          <div className="my-auto mx-2">
+            This obj never renders any content but triggers a redirect to the
+            specified target.
```
dcsaszar, line 40: * [ ] FYI Depending on which behavior we agree on, instead of "specified target" we could render a link to the target. This way, the editor can simulate the redirect by clicking.

src/Objs/Page/PageLayoutComponent.tsx
dcsaszar, line 1: * [ ] AFAICS this is unrelated to the PR. Also this change will probably break landing pages outside the hierarchy. Revert?

dcsaszar (changes requested):
Converging rapidly. See comments.

src/Objs/Page/PageLayoutComponent.tsx
markus-codewrangler: what is "this"?

src/Objs/Redirect/RedirectComponent.tsx
markus-codewrangler: Please do the fine-tuning later. This PR is just the basic implementation.

src/Objs/Page/PageLayoutComponent.tsx
dcsaszar: GitHub magic 😿 
image

dcsaszar: Looks like this is triggered by https://github.com/Scrivito/scrivito-portal-app/pull/348#discussion_r1559182375, so OK

dcsaszar (approved):
All clear from my side.

src/Objs/SubnavigationOverview/SubnavigationOverviewEditingConfig.ts
```diff
@@ -32,6 +27,6 @@ provideEditingConfig(SubnavigationOverview, {
     'hideInNavigation',
     'linkIcon',
     'topBannerBackground',
-    'requireUserLogin',
+    '',
```
apepper, line 30: Please remove this.

src/Objs/Redirect/RedirectComponent.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <div className="alert alert-warning d-flex m-auto">
+          <i className="bi bi-exclamation-circle bi-2x" aria-hidden="true"></i>
+          <div className="my-auto mx-2">
+            This obj never renders any content but triggers a redirect to the
+            target specified in the page properties.
+          </div>
+        </div>
+      </div>
+    )
+  }
```
apepper, line 47: Sorry for the rollback: But we need the redirect also for editors. Otherwise one will always "land" on the redirect page, if one clicks "My Tynacoon".

So please remove the `isEditorLoggedIn()` specific code from this component.

apepper (changes requested):
FEEDBACK ([05a50e4](https://github.com/Scrivito/scrivito-portal-app/pull/348/commits/05a50e4705463541b1cab5c92b3a138871bcdef5))

Getting closer 🚀!

Sadly the "editor note" turns out to be not that good of an idea - sorry about that (see comments).

Could you also...
* [ ] merge the current `main` into this branch?
* [ ] Update the content of the working copy? Not all pages in the english "My Tynacoon" are protected and all german content is not yet changes to the new style of redirect.

src/Objs/Redirect/RedirectComponent.tsx
markus-codewrangler: I now took the best of both solutions 🙂

markus-codewrangler:
> merge the current main into this branch?

✅ – I still cannot check any checkboxes in this project without sufficient permissions. Why don't have all Scrivito SDK team members the usual developer permissions on this repo - same as in the scrivito_js repo?? 🙄

> Update the content of the working copy? Not all pages in the english "My Tynacoon" are protected

- Which exact pages are you talking about (I don't have time for "manual" diffing)
- Did these not-yet-protected EN pages already exist when this PR was created? (I don't think so)

> and all german content is not yet changes to the new style of redirect.

As mentioned before, this PR is [just the basic implementation](https://github.com/Scrivito/scrivito-portal-app/pull/348#discussion_r1561226327), so please [take care of this within the portal-app team](https://github.com/Scrivito/scrivito-portal-app/pull/348#issuecomment-2049976331).

This PR deviated from the usual process in multiple ways:

- was open for well over a week before it was reviewed
- was reviewed by two different people
- had contradicting/evolving review comments
- contains several iterations of (changing!) requirements stated in review comments (which have **not** been part of the original requirements in the dev plan)
- the german content in question has been published 2 weeks(!!) after this PR was created AFAICS
- the german content is still being implemented/evolving AFAIU (so if a new german page would be published after I adapted the working copy but before you review my changes , will this then become an infinite loop/race situation?)

I know that **some** of this cannot be avoided in a PR sometimes, **but all of this in one PR is too much**!

I have volunteered to make a basic implementation of this story (since I had worked substantially on the External Users story anyway) as a "visiting" developer to the portal-app team. I'm not part of the internal coordination between Alex, Dave and Kris, so I don't have the latest information how the project requirements are evolving or what decisions are made (so there is an ongoing danger that requirements keep evolving without my knowledge while this PR drags on...). Also I have more than enough to do with my other (not-portal-app) stories.

So if there are no more **critical** issues with this PR, please merge this now and take care of any clean-up/cosmetic tasks (e.g. changing EN or DE page settings/redirects) in follow-ups yourselves. Thanks!

apepper (approved):
LGTM ([5ba1bc7](https://github.com/Scrivito/scrivito-portal-app/pull/348/commits/5ba1bc715d79b7da9b18e89bfc29f0bcd7ed6c4e))

Thank you so much for your work. Much appreciated 🚀!

PR #352, opened 2024-04-04
Better in-place editing feedback for uppercased design elements

dcsaszar:
Based on an internal discussion here: https://infopark.slack.com/archives/G01F5J9817Z/p1712222151294679?thread_ts=1712220545.316979&cid=G01F5J9817Z

### DIFF
--- a/src/assets/stylesheets/scrivitoEditing.scss
+++ b/src/assets/stylesheets/scrivitoEditing.scss
@@ -1,0 +1,7 @@
+// --------------------------------------------------------------- //
+// In-place editing styles adaptions
+// --------------------------------------------------------------- //
+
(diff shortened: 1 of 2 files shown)


apepper (approved):
LGTM ([bdf944c](https://github.com/Scrivito/scrivito-portal-app/pull/352/commits/bdf944cf093ce31e8187684fa7820e863ef94b66))

I like it. A pragmatic way.

Did you check all the contents, if the casing is correct in edit mode?

dcsaszar:
> Did you check all the contents, if the casing is correct in edit mode?

No, I didn't. But I also think if we do have issues we'll run into them anyway at some point. Same as with typos and similar stuff.

PR #354, opened 2024-04-04
Multi-language basics

dcsaszar:
Assumptions/constraints:
* Each site has a unique homepage `_language`
* If no site matches the current URL, the first site by language code is the fallback site
* The language prefix is always used, even if there is only one site
* The language detection ignores a multi-tenancy path prefix, no matter if multi-tenant or not

I.e. `<origin>/<tenantId>/<lang><path>` or `<origin>/<lang><path>`

### DIFF
--- a/src/config/scrivito.ts
+++ b/src/config/scrivito.ts
@@ -1,5 +1,5 @@
 import { configure } from 'scrivito'
-import { getTenantFromEnv } from './scrivitoSites'
+import { getTenantFromEnv, baseUrlForSite, siteForUrl } from './scrivitoSites'
 
 export function configureScrivito() {
   const tenant = getTenantFromEnv()
@@ -8,9 +8,11 @@ export function configureScrivito() {
   const config: Parameters<typeof configure>[0] = {
     adoptUi: true,
     autoConvertAttributes: true,
+    baseUrlForSite,
     optimizedWidgetLoading: true,
     strictSearchOperators: true,
     contentTagsForEmptyAttributes: false,
+    siteForUrl,
     tenant,
     // @ts-expect-error // TODO: Remove later on
     unstable: {
@@ -24,7 +26,6 @@ export function configureScrivito() {
 
   if (!import.meta.env.SCRIVITO_TENANT) {
     // Multitenancy mode
-    config.routingBasePath = `/${tenant}`
     config.extensionsUrl = `/_scrivito_extensions.html?tenantId=${tenant}`
   }
 
--- a/src/config/scrivito.ts
+++ b/src/config/scrivito.ts
@@ -1,13 +1,17 @@
 import { configure } from 'scrivito'
+import { getTenantFromEnv } from './scrivitoSites'
 
 export function configureScrivito() {
+  const tenant = getTenantFromEnv()
+  if (!tenant) return
+
   const config: Parameters<typeof configure>[0] = {
     adoptUi: true,
     autoConvertAttributes: true,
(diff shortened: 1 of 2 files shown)


src/config/scrivitoSites.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (language) urlParts.push(language)
+
+  return urlParts.join('/')
+}
+
+export function siteForUrl(
+  url: string,
+): { baseUrl: string; siteId: string } | undefined {
+  const language = /\b\/([0-9a-f]{32}\/)?(?<lang>[a-z]{2})([?/]|$)/.exec(url)
+    ?.groups?.lang
```
apepper, line 24: Are language codes always 2 chars long?

apepper (commented):
FEEDBACK ([be09e9e](https://github.com/Scrivito/scrivito-portal-app/pull/354/commits/be09e9ede17706e1d461507f3bd6075d78eddf07))

Nice 🚀!

I left some comments how this can be improved even more.

src/config/scrivitoSites.ts
dcsaszar: Simple ISO language codes: yes.

apepper: So the question is: Does Scrivito always use "Simple ISO language codes" for the "_language" attribute?

dcsaszar: Non-custom languages are two-letter. The UI accepts any custom code as long as it consists of letters and dashes. In order to support this, we'd need to get all languages first and check them individually instead of guessing from the URL.

Another assumption the code currently makes is that there is only one website with language versions. The UI allows to create multiple sites with multiple versions though.

src/config/scrivito.ts
```diff
@@ -20,35 +24,6 @@ export function configureScrivito() {
 
   if (!import.meta.env.SCRIVITO_TENANT) {
```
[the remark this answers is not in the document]
dcsaszar: #358

src/config/scrivitoSites.ts
apepper: > Another assumption the code currently makes is that there is only one website with language versions. The UI allows to create multiple sites with multiple versions though.

We could add a validation to the `Homepage` obj class, that there are not two site roots with the same language. If someone would like to create multiple sites with the same language, the code would need to be adjusted anyway.

src/config/scrivitoSites.ts
```diff
@@ -35,6 +35,14 @@ export function siteForUrl(
   if (baseUrl) return { baseUrl, siteId }
 }
 
+export async function ensureSiteIsPresent() {
+  if ((await load(currentSiteId)) === null) {
+    navigateTo(() =>
+      Obj.onAllSites().where('_path', 'equals', '/').order('_language').first(),
```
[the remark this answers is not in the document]
dcsaszar: #359

src/config/scrivitoSites.ts
```diff
@@ -0,0 +1,33 @@
+const location = typeof window !== 'undefined' ? window.location : undefined
+
+export function getTenantFromEnv(): string | undefined {
```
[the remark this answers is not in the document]
apepper: Do you mean the `VITE_MULTITENANCY_FALLBACK_SCRIVITO_TENANT` code-path? It is not in use and can actually be deleted. Not sure if I will ever reactive it.

PR #356, opened 2024-04-11
Make address widget labels editable

dcsaszar:
https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dave-german.scrivito-portal-app.pages.dev/de?_scrivito_workspace_id=p5bcc7f596b13012&_scrivito_display_mode=diff

PR #357, opened 2024-04-11
Make search-related labels and headlines editable

dcsaszar:
Same working copy as #356 

https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dave-german-search.scrivito-portal-app.pages.dev/de?_scrivito_workspace_id=p5bcc7f596b13012&_scrivito_display_mode=diff

### DIFF
--- a/src/Widgets/SearchResultsWidget/SearchResultsWidgetComponent.tsx
+++ b/src/Widgets/SearchResultsWidget/SearchResultsWidgetComponent.tsx
@@ -1,4 +1,6 @@
 import {
+  connect,
+  ContentTag,
   currentPageParams,
   ImageTag,
   InPlaceEditingOff,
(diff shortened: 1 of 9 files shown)


src/Widgets/SearchResultsWidget/SearchResultsWidgetComponent.tsx
```diff
@@ -98,9 +111,22 @@ provideComponent(SearchResultsWidget, ({ widget }) => {
...
+  const attribute = attributes[totalCount] || 'resultsHeadline'
 
-  return `${totalCount} search results`
-}
+  return (
+    <ContentTag
+      tag="span"
+      content={widget}
+      attribute={attribute}
+      dataContext={{ count: totalCount.toString() }}
```
krishan, line 129: @apepper @dcsaszar I just stumbled over this code 😨 

Passing hand-made "value objects" into `dataContext` is a deprecated, legacy API that was never released publicly. The SDK only keep it around because Neoletter still uses it. Please don't use it in the Portal. It's scheduled to be removed as soon as Neoletter no longer needs it.

Please remove this here as soon as possible, so that this API does not proliferate in any forks.

apepper: Addressed in https://github.com/Scrivito/scrivito-portal-app/pull/506.

PR #358, opened 2024-04-12
Multitenancy refactoring

dcsaszar:
https://github.com/Scrivito/scrivito-portal-app/pull/354#discussion_r1560916113

### DIFF
--- a/src/config/scrivito.ts
+++ b/src/config/scrivito.ts
@@ -2,10 +2,9 @@ import { configure } from 'scrivito'
 import {
   baseUrlForSite,
   ensureSiteIsPresent,
-  getTenantFromEnv,
   siteForUrl,
 } from './scrivitoSites'
-import { isMultitenancyEnabled } from './scrivitoTenants'
+import { getTenantFromEnv, isMultitenancyEnabled } from './scrivitoTenants'
 
 export function configureScrivito() {
   const tenant = getTenantFromEnv()
--- a/src/config/scrivito.ts
+++ b/src/config/scrivito.ts
@@ -4,10 +4,10 @@ import {
   ensureSiteIsPresent,
   siteForUrl,
 } from './scrivitoSites'
-import { getTenantFromEnv, isMultitenancyEnabled } from './scrivitoTenants'
+import { scrivitoTenantId, isMultitenancyEnabled } from './scrivitoTenants'
 
 export function configureScrivito() {
-  const tenant = getTenantFromEnv()
+  const tenant = scrivitoTenantId()
   if (!tenant) return
 
   const config: Parameters<typeof configure>[0] = {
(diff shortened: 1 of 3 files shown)


src/config/scrivitoTenants.ts
```diff
@@ -0,0 +1,36 @@
+export function isMultitenancyEnabled(): boolean {
+  return !import.meta.env.SCRIVITO_TENANT
+}
+
+const location = typeof window !== 'undefined' ? window.location : undefined
+
+export function scrivitoTenantId(): string | undefined {
```
apepper, line 7: I think it would be better, if we switch to 
```suggestion
export function scrivitoTenantId(): string {
```

and throw, if no tenantId could be calculated. This would help to avoid "if (!tenant)" code ouside of `scrivitoTenantId()`.

src/config/scrivito.ts
```diff
@@ -29,8 +29,7 @@ export function configureScrivito() {
     },
   }
 
-  if (!import.meta.env.SCRIVITO_TENANT) {
-    // Multitenancy mode
+  if (isMultitenancyEnabled()) {
     config.extensionsUrl = `/_scrivito_extensions.html?tenantId=${tenant}`
   }
```
apepper, line 34: I think its is safe to always set `extensionsUrl` to `/_scrivito_extensions.html?tenantId=${tenant}` regardless of `isMultitenancyEnabled()`. With that, we no longer need to have a `const config`, but could call `configure()` directly.

apepper (commented):
FEEDBACK ([07b03f8](https://github.com/Scrivito/scrivito-portal-app/pull/358/commits/07b03f8fbfc7469484d6129ee871e5902a20e9f7))

Nice 🚀!

Please see my comments for further improvements.

src/config/scrivitoTenants.ts
apepper: See also discussion in https://github.com/Scrivito/scrivito-portal-app/pull/354#discussion_r1560918091.

src/config/scrivito.ts
```diff
@@ -2,22 +2,22 @@ import { configure } from 'scrivito'
...
   siteForUrl,
 } from './scrivitoSites'
+import { scrivitoTenantId } from './scrivitoTenants'
 
 export function configureScrivito() {
-  const tenant = getTenantFromEnv()
-  if (!tenant) return
+  const tenant = scrivitoTenantId()
 
   const config: Parameters<typeof configure>[0] = {
```
apepper, line 12: Could you please inline `config`?

apepper (commented):
FEEDBACK ([975bc3f](https://github.com/Scrivito/scrivito-portal-app/pull/358/commits/975bc3face29d52a7a8192fa3da315df6ec4def3))

Very nice 🚀!

Almost there. Only some cosmetics left.

dcsaszar:
Sorry for typos 🙈

PR #359, opened 2024-04-12
Root path falls back to preferred language site

dcsaszar:
https://github.com/Scrivito/scrivito-portal-app/pull/354#discussion_r1560927997


src/config/scrivitoSites.ts
```diff
@@ -37,12 +37,27 @@ export function siteForUrl(
 
 export async function ensureSiteIsPresent() {
   if ((await load(currentSiteId)) === null) {
-    navigateTo(() =>
-      Obj.onAllSites().where('_path', 'equals', '/').order('_language').first(),
-    )
+    navigateTo(() => {
```
apepper, line 40: FYI: Conflicts. Please resolve.

src/config/scrivitoSites.ts
```diff
@@ -37,12 +37,27 @@ export function siteForUrl(
...
+      const sites = Obj.onAllSites().where('_path', 'equals', '/').toArray()
```
apepper, line 41: I'm not sure how `navigateTo` treats loadable content. Can you maybe explain?

apepper (commented):
FEEDBACK ([6a9347f](https://github.com/Scrivito/scrivito-portal-app/pull/359/commits/6a9347f1572de8e434f1d0ca6a2aad8340d2d7e7))

Already looks really nice. I only have some questions (see comments).

src/config/scrivitoSites.ts
dcsaszar: https://docs.scrivito.com/js-sdk/navigateTo#remarks

> Scrivito loads data asynchronously from the CMS backend. To make sure your target object data is available for navigation, pass in target as a function.

This means that navigation will happen only after all requested content has been fetched from the backend.
In this case these are all existing root objects.

src/config/scrivitoSites.ts
```diff
@@ -37,8 +37,55 @@ export function siteForUrl(
...
```
apepper, line 41: Does this also include neoletter content - if available?

src/config/scrivitoSites.ts
```diff
@@ -37,8 +37,55 @@ export function siteForUrl(
...
+        location.replace(`${location.origin}/${fallbackScrivitoTenant}`)
+        return
+      }
+    }
+
+    throw new Error('Could not determine tenant!')
   }
+
+  return tenant
 }
```
apepper, line 91: Where does this come from? Is this still needed?

src/config/scrivitoSites.ts
apepper: Thank your for the clarification!

apepper (commented):
FEEDBACK ([7ddbf1b](https://github.com/Scrivito/scrivito-portal-app/pull/359/commits/7ddbf1b7e39eb9d886ea680d1599c913519db22f))

See comments.

src/config/scrivitoSites.ts
dcsaszar: Yes, currently all multi-site related code assumes all sites are on the same origin and their content is suited for the Portal App. In order to support non-portal sites we'd two things need to be done:

* Update the multi-site config to properly handle "non-portal" sites, e.g., Neoletter content or Example App content
* In the `ensureSiteIsPresent` helper only consider base URLs that match the pattern `<current_origin>/<site_language>`

src/config/scrivitoSites.ts
dcsaszar: No! Somehow the main merge failed to properly remove it.

src/config/scrivitoSites.ts
apepper: Could you takle this in a follow up PR?

dcsaszar: Part two, sure!
Any idea if and how we could implement part 1?:
> to properly handle "non-portal" sites

apepper: For me it would be "good enough" to filter out known other site names such as Neoletter. Not a generic "there can be endless many sites" (yet).

PR #361, opened 2024-04-12
Add language switch

dcsaszar:


src/Widgets/NavigationWidget/SubComponents/LanguageSwitch.tsx
```diff
@@ -0,0 +1,87 @@
...
+  InPlaceEditingOff,
+  LinkTag,
+  Obj,
+  connect,
+  currentPage,
+  currentSiteId,
+} from 'scrivito'
+import { HomepageInstance } from '../../../Objs/Homepage/HomepageObjClass'
+
+export const LanguageSwitch = connect(function LanguageSwitch() {
```
apepper, line 13: High-level feedback: It would be nice, if the LanguageSwitch disappears, if only one site exists (so it is easy to convert the "multi-language portal" to a "single language portal").

E.g. in the working-copy https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dave-language-switch.scrivito-portal-app.pages.dev/en?_scrivito_workspace_id=j539ba0e70061273&_scrivito_display_mode=view

dcsaszar: * [x] f839952

src/Widgets/NavigationWidget/SubComponents/LanguageSwitch.tsx
```diff
@@ -0,0 +1,89 @@
...
+      const pageVersion = siteId && currentPage()?.versionOnSite(siteId)
+      const language = site.language() || 'en'
+      return {
+        label: displayName(language),
+        language,
+        version: pageVersion,
+        root: site as HomepageInstance,
+      }
+    })
+    .sort((a, b) => a.label.localeCompare(b.label))
```
apepper, line 28: `localeCompare` also allows to set a `locales` as a second parameter. Not sure, if we should use the current language of the site for that. What do you think?

src/Widgets/NavigationWidget/SubComponents/LanguageSwitch.tsx
```diff
@@ -0,0 +1,89 @@
...
+  if (versions.length < 2) return null
+
+  const activeSite = (
+    versions.find(({ root }) => root.siteId() === currentSiteId()) ||
+    versions[0]
+  )?.root
+
+  if (!activeSite) return null
+
+  const title = <LanguageLabel root={activeSite} showIconOnly />
```
apepper, line 39: I would prefer, if the `showIconOnly` option gets removed and an `ImageTag` is used instead. It has the drawback, that `attribute="siteLanguageIcon"` and `className="img-flag"` are duplicated. But the code-flow is probably easier to understand for readers.

apepper (commented):
FEEDBACK ([f839952](https://github.com/Scrivito/scrivito-portal-app/pull/361/commits/f83995206b0126fcf31730a9437ad48218d8df7f))

Looks really nice. I only found smaller issues.

src/Widgets/NavigationWidget/SubComponents/LanguageSwitch.tsx
dcsaszar: If you look at the code after this refactoring we have quite some duplication. I tend to unify CurrentLanguageLabel and LanguageLabel again, adding a prop like `labelClassName`. WDYT?

apepper: Sounds good to me.

apepper (commented):
FEEDBACK ([7d59e0e](https://github.com/Scrivito/scrivito-portal-app/pull/361/commits/7d59e0e4befa62f281d824c9067ee62cef4d7a4c))

Very nice 🚀! Only one small refactoring left (as you suggested).

apepper (approved):
LGTM ([be9b516](https://github.com/Scrivito/scrivito-portal-app/pull/361/commits/be9b516fc6de2990c7a505cad26f0a8b8317a526))

Very nice 🚀!

PR #365, opened 2024-04-16
Neoletter form widgets package integration

aea-JR:
Content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://neoletter-form-widgets-packa.scrivito-portal-app.pages.dev/de/kontakt-e6f438cec91e8b70?_scrivito_workspace_id=o6020ce728752aa0&_scrivito_display_mode=view


package.json
```diff
@@ -23,6 +23,7 @@
     "react-helmet-async": "^2.0.4",
     "react-toastify": "^10.0.5",
     "scrivito": "^1.41.0-rc1",
+    "scrivito-neoletter-form-widgets": "^0.0.11",
```
apepper, line 26: Please keep an eye on the resulting file size. With your PR `dist/assets/index-<some hex>.js` grew from `763.94 kB │ gzip: 241.79 kB` to `1,196.95 kB │ gzip: 387.00 kB`.

The size of the `index.js` file plays a critical rule how fast the app can be loaded and render content - e.g. for Lighthouse scores. A 50% increase is not desirable.

aea-JR: I will try to reduce the bundle size, most probably this is caused because of the BS-Icons.

apepper: Just as a rough benchmark: If you don't include neoletter widgets at all (not from source and not from package) the index.js filesize is: `757.41 kB │ gzip: 240.47 kB`.

So ideally the package could be as small as `6kB | gzip 1kB`. Auf cause there is some boilerplate code by Webpack/Vite, so reaching so low will be hard.

apepper: So overall a filesize-budget of less then 20 kB for the package would be ideal.

aea-JR: Current file size without the package: `762.71 kB │ gzip: 241.65 kB`
Current file size with latest package changes: `795.48 kB │ gzip: 250.29 kB`. 
Difference between them:  `32.77kB | gzip: 8.64kB`
The initial file size with the package was `1,196.95 kB │ gzip: 387.00 kB` as you already mentioned.

apepper: I measure `743.04 kB | gzip: 236.14 kB` for the current `main` (see https://app.circleci.com/pipelines/github/Scrivito/scrivito-portal-app/1049/workflows/616a1405-0b08-466a-96cc-b3f71c62aa2b/jobs/2924?invite=true#step-103-5049_96).

With that this PR adds `26.63 kB | gzip: 7.37 kB`. Very nice improvement 🚀!

apepper (approved):
LGTM ([4f553b7](https://github.com/Scrivito/scrivito-portal-app/pull/365/commits/4f553b7e8a648e7612504660ebf45f8fc9545dbd))

Wow, it works great 🚀!

Thank you @aea-JR for your dedicated work over such a long time!

PR #366, opened 2024-04-17
Load more button

apepper:
Content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://loadmorebutton.scrivito-portal-app.pages.dev/en/tickets?_scrivito_workspace_id=sff80da3d282ccf3&_scrivito_display_mode=diff


src/Widgets/DataLoadMoreButtonWidget/DataLoadMoreButtonWidgetComponent.tsx
```diff
@@ -0,0 +1,55 @@
...
+  if (buttonColor) classNames.push(buttonColor)
+
+  const alignmentClassName = alignmentClassNameWithBlock(
+    widget.get('alignment'),
+  )
+
+  if (!moreAvailable) {
+    if (!isInPlaceEditingActive() && !isComparisonActive()) return null
+
+    classNames.push('disabled')
```
dcsaszar, line 36: * [ ] As the only thing we want to indicate is that the button is conditional `disabled` is misleading.
* [ ] FYI I see that we can't go full WYSIWYG here but how about this: Make the button behave as intended for all pages but page 1.

Alternatively we could make a greyed-out paging button the default behavior except for cases when there's only one page. A disabled button on the last page could be the better UX as it more clearly indicates "you have successfully reached the last page".

src/Widgets/DataLoadMoreButtonWidget/DataLoadMoreButtonWidgetComponent.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <button className={classNames.join(' ')}>{widget.get('title')}</button>
+      </WidgetTag>
+    )
+  }
+
+  return (
+    <WidgetTag className={alignmentClassName}>
+      <button
+        className={classNames.join(' ')}
+        onClick={() => setLimit(limit + 10)}
```
dcsaszar, line 49: * [ ] FYI One more argument for using a paging context. If the editor configures a locator with, e.g., max 3 it will be surprising to see as many as 13 after "show more". If the editor configures 100 as little as 110 will be surprising. In the end we'll probably have a configurable more-batch size, but using the initial page-size could still be a sane default.

src/Widgets/DataLoadMoreButtonWidget/DataLoadMoreButtonWidgetEditingConfig.ts
```diff
@@ -0,0 +1,51 @@
...
+    buttonSize: 'medium',
+  },
+  validations: [
+    [
+      'title',
+      (title) => {
+        if (!title) return 'Please provide a title'
+      },
+    ],
+  ],
```
dcsaszar, line 50: * [ ] Instead of a validation I'd prefer an initial content. The purpose of the button is pretty clear so a sane default would make the editor's life way easier. The look after clearing the text explains that a title should be set, no validation is needed. But I'm also fine with having both, initial content and a validation.

src/Widgets/DataLoadMoreButtonWidget/thumbnail.svg
dcsaszar, line 1: * [ ] FYI seeing this icon I immediately thought: "I'd like to have a load-more button like this in the app!".

dcsaszar (changes requested):
Nice. See comments for feedback. Some is conceptual in nature.

Regarding the content:
* [x] FYI I'd suggest to use the button style and size we use for the search results (`btn btn-outline-primary`). This would keep the overall and in-page look and feel more consistent.
* [x] FYI the German label on the search is currently "Weitere Ergebnisse laden". Not sure how consistent we'd like to be and which (Mehr/Weitere/Weitere X) is best.

src/Widgets/DataLoadMoreButtonWidget/DataLoadMoreButtonWidgetComponent.tsx
apepper: I removed `classNames.push('disabled')` but kept the grey-out stuff. The button will disappear if there is not more to load in visitor mode.

apepper:
REVIEW ([9440a2d](https://github.com/Scrivito/scrivito-portal-app/pull/366/commits/9440a2d875d5da81f45c159ffa81a424917f0160))

src/Widgets/DataLoadMoreButtonWidget/DataLoadMoreButtonWidgetComponent.tsx
```diff
@@ -44,10 +39,7 @@ provideComponent(DataLoadMoreButtonWidget, ({ widget }) => {
 
   return (
     <WidgetTag className={alignmentClassName}>
-      <button
-        className={classNames.join(' ')}
-        onClick={() => setLimit(limit + 10)}
-      >
+      <button className={classNames.join(' ')} onClick={() => loadMore()}>
```
dcsaszar, line 41: * [ ] FYI could be simplified to `<button className={classNames.join(' ')} onClick={loadMore}>`

dcsaszar (approved):
LGTM

PR #367, opened 2024-04-18
Use protected content inside portal - follow up

apepper:
Follow up of https://github.com/Scrivito/scrivito-portal-app/pull/348.

Content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://protects-content-follow-up.scrivito-portal-app.pages.dev/en?_scrivito_workspace_id=kaac5c379b1b683f&_scrivito_display_mode=view

### DIFF
--- a/src/Objs/Homepage/HomepageEditingConfig.ts
+++ b/src/Objs/Homepage/HomepageEditingConfig.ts
@@ -26,6 +26,9 @@ provideEditingConfig(Homepage, {
     sitePortalOnlyMode: { title: 'Use portal-only mode?' },
     siteCartPage: { title: 'Location of cart page' },
     sitePortalOverviewPage: { title: 'Location of portal overview page' },
+    sitePortalOverviewFallbackPage: {
+      title: 'Location of portal overview page (redirect fallback)',
+    },
     siteSearchResultsPage: {
       title: 'Location of search results page',
     },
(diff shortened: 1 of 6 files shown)


src/Objs/Homepage/HomepageEditingConfig.ts
```diff
@@ -26,6 +26,9 @@ provideEditingConfig(Homepage, {
     sitePortalOnlyMode: { title: 'Use portal-only mode?' },
     siteCartPage: { title: 'Location of cart page' },
     sitePortalOverviewPage: { title: 'Location of portal overview page' },
+    sitePortalOverviewFallbackPage: {
+      title: 'Location of portal overview page (redirect fallback)',
```
dcsaszar, line 30: * [ ] FYI A better explanation would give some details about the nature of the fallback, e.g., "Portal overview page for logged-out visitors".
* [ ] FYI How about we save editors from the hassle by making the "active" state redirect aware? I.e. the actual path of `sitePortalOverviewPage` follows the redirect. Such a helpers is a good idea in general b/c it can also be used as the preferred link target in navigations, saving roundtrips and helping crawlers of prerendered content.

dcsaszar (approved):
LGTM but see comment for potentially better editor UX.

PR #368, opened 2024-04-18
Use protected content inside portal - follow up 2

apepper:
Follow up of https://github.com/Scrivito/scrivito-portal-app/pull/348.

Closes #367.

Idea by @dcsaszar in https://github.com/Scrivito/scrivito-portal-app/pull/367#discussion_r1570342507

### DIFF
--- a/src/Objs/Redirect/RedirectComponent.tsx
+++ b/src/Objs/Redirect/RedirectComponent.tsx
@@ -14,12 +14,12 @@ provideComponent(Redirect, ({ page }) => {
   const link = page.get('link')
 
   React.useEffect(() => {
-    if (!link) return
-
     if (requireLogin && !isUserLoggedIn()) {
       ensureUserIsLoggedIn()
       return
     }
+
+    if (!link) return
 
     load(() => urlFor(link)).then((url) => {
       if (!url) return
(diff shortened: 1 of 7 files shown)


src/Objs/Redirect/RedirectComponent.tsx
```diff
@@ -14,13 +14,13 @@ provideComponent(Redirect, ({ page }) => {
...
   React.useEffect(() => {
-    if (!link) return
-
     if (requireLogin && !isUserLoggedIn()) {
       ensureUserIsLoggedIn()
       return
     }
 
+    if (!link) return
+
```
dcsaszar, line 23: * [ ] FYI This change will trigger `ensureUserIsLoggedIn()` twice. I liked the code before more, because it only triggered it once after the link target was loaded. Why is this important, maybe add a comment?:
> Bugfix: RedirectComponent ensure log in before link check

src/Widgets/NavigationWidget/SubComponents/MetaNavigation.tsx
```diff
@@ -26,7 +27,12 @@ export const MetaNavigation = connect(function MetaNavigation({
     )
   }
 
-  const sitePortalOverviewPage = root.get('sitePortalOverviewPage')
+  const rawSitePortalOverviewPage = root.get('sitePortalOverviewPage')
+  // TODO: Remove workaround, once #10699 is available
+  const sitePortalOverviewPage =
+    (isRedirect(rawSitePortalOverviewPage) &&
+      rawSitePortalOverviewPage.get('link')?.obj()) ||
+    rawSitePortalOverviewPage
```
dcsaszar, line 35: * [ ] FYI Instead of repeating the same code three times, how about a dedicated (`getSitePortalOverviewPage`) or generic (`followRedirect(root.get('sitePortalOverviewPage'))`) helper?

dcsaszar (approved):
See comments.

src/Objs/Redirect/RedirectComponent.tsx
apepper: Because a not logged in user with have `link === null` since the user can't read the protected content.

dcsaszar (approved):

PR #373, opened 2024-04-22
Incorporate pisa - part 1

apepper:
Basic version: As soon as `env.PISA_URL` is set, pisa is used.

Extended version: For our preview-deployments and "main" a special `env.ENABLE_PISA_ONLY_ON_PISA_BRANCH` is introduced - in order to have pisa only running on https://pisa.scrivito-portal-app.pages.dev. This env is already set for this project in Cloudflare pages.

Of cause more modules will follow, once we agree on the basic mechanism.


src/Data/pisaClient.ts
```diff
@@ -0,0 +1,14 @@
+import { createApiClient } from 'scrivito'
+import { ensureString } from '../utils/ensureString'
+
+export function pisaUrl(): string {
```
dcsaszar, line 4: * [ ] Currently no external caller uses `pisaUrl()`. Remove `export`?

vite.config.ts
```diff
@@ -15,6 +15,9 @@ export default defineConfig(({ mode }) => {
   const outDir = 'dist'
 
   const PISA_URL = env.PISA_URL
+  const enablePisa = env.ENABLE_PISA_ONLY_ON_PISA_BRANCH
+    ? env.CF_PAGES_BRANCH === 'pisa' && !!PISA_URL
+    : !!PISA_URL
```
dcsaszar, line 20: * [ ] FYI the `enablePisa` logic could be moved to the server setup, e.g., to the "build command" option. This way, we'd be able to hide this very specific condition from the app config.

See https://stackoverflow.com/a/2440950

.circleci/config.yml
```diff
@@ -7,6 +7,8 @@ jobs:
         type: string
     docker:
       - image: cimg/node:18.17
+        environment:
+          PISA_URL: 'some-fake-pisa-url'
```
dcsaszar, line 11: * [ ] Even with the commit message it's unclear what this fake setting is about. I personally never looked at any filesize in the CI environment. Maybe we don't need this or there should be a helpful comment in the config.yml?

dcsaszar (changes requested):
Looks good so far.

vite.config.ts
apepper: Not sure how this would look like.

apepper:
REVIEW ([295a3fc](https://github.com/Scrivito/scrivito-portal-app/pull/373/commits/295a3fcb78c5c33e2ccb4ef71b2a0b26f5aceb2e))

dcsaszar (commented):
See comments.

vite.config.ts
apepper: I went with:
```bash
PISA_URL=$([ "$CF_PAGES_BRANCH" = "pisa" ] && echo $PISA_URL) npm run build
```

.circleci/config.yml
```diff
@@ -8,6 +8,9 @@ jobs:
     docker:
       - image: cimg/node:18.17
         environment:
+          # With a PISA_URL environment variable the build will not contain fake LocalStorage
+          # data classes. Given that it's easier to see the actual package size,
+          # without artificial fake data.
```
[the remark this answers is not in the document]
apepper: I removed the comment and manually added the env `PISA_URL` to our circle CI project.

At least I'm interested in seeing the file-size evolve. Other projects can set this up however they like.

dcsaszar (approved):
LGTM

vite.config.ts
dcsaszar: FYI Another option that came to my mind is a 2nd Pages setup building only the main branch with a different env. This one-time setup would save us from permanently having to duplicate the main branch.

apepper: As far as I know Cloudflare Pages does not allow to create two different applications, that link to the same git repository.

dcsaszar: I see. Very unfortunate.

PR #381, opened 2024-04-23
Incorporate pisa - part 3

apepper:


src/Data/DataIndexResponse.ts
```diff
@@ -0,0 +1,7 @@
+import { RawItem } from './RawItem'
+
+export interface DataIndexResponse {
+  results: RawItem[]
+  count?: number
+  continuation?: string
+}
```
dcsaszar, line 7: * [ ] FYI As long as there are no plans to add non-type exports to things like `DataIndexResponse` and `RawItem` a `Data/types.ts` could make more sense.

src/Data/toClientParams.ts
```diff
@@ -0,0 +1,41 @@
...
+    _order: params.order().length
+      ? params
+          .order()
+          .map((o) => o.join('.'))
+          .join(',')
+      : undefined,
+    _limit: params.limit().toString(),
+    _search: params.search() || undefined,
+    _count: params.includeCount()
+      ? params.includeCount().toString()
```
dcsaszar, line 26: * [ ] FYI `params.includeCount().toString()` is a very elaborate way to say `true`

src/Data/toClientParams.ts
```diff
@@ -0,0 +1,41 @@
...
+type IndexCallback = NonNullable<DataConnection['index']>
+type IndexParams = Parameters<IndexCallback>[0]
+
+interface OperatorSpec {
+  operator: 'equals' | 'notEquals'
+  opCode: 'eq' | 'neq'
+  value: string
+}
+
+export function toClientParams(params: IndexParams) {
```
dcsaszar, line 13: * [ ] Missing a return type.

src/Data/toClientParams.ts
```diff
@@ -0,0 +1,41 @@
...
+    ...toClientFilterParam(params.filters()),
+    _continuation: params.continuation(),
+    _order: params.order().length
+      ? params
+          .order()
+          .map((o) => o.join('.'))
+          .join(',')
+      : undefined,
+    _limit: params.limit().toString(),
+    _search: params.search() || undefined,
```
dcsaszar, line 24: * [ ] FYI Not sure about the `|| undefined`. Looks like a safeguard against accidental empty-string searches, but don't we hide potential application errors with this?

src/Data/toClientParams.ts
```diff
@@ -0,0 +1,41 @@
...
+      ? params.includeCount().toString()
+      : undefined,
+  }
+}
+
+function toClientFilterParam(filters: Record<string, OperatorSpec>) {
+  const params: Record<string, string> = {}
+
+  Object.keys(filters).forEach((name) => {
+    const { opCode, value } = filters[name] as OperatorSpec
```
dcsaszar, line 35: * [ ] FYI We can get rid of casting thanks to `Object.entries`

```ts
Object.entries(filters).forEach(([name, { opCode, value }]) => {
```

src/utils/convertBlobAttributes.ts
```diff
@@ -23,20 +23,23 @@ async function blobToBinary(blob: Blob | File): Promise<{
   dataBase64: string
   filename: string
 
-  // TODO: Remove for pisa
-  contentLength: number
-  contentType: string
-  _id: string
+  contentLength?: number // only needed for "localStorage" fake
+  contentType?: string // only needed for "localStorage" fake
+  _id?: string // only needed for "localStorage" fake
```
dcsaszar, line 28: * [ ] FYI Currently a single headline comment would do.
* [ ] FYI I'd not call it "fake" b/c there are valid use cases for local storage.

src/utils/pisaDataBinaryToUrl.ts
```diff
@@ -0,0 +1,31 @@
+import { pisaClient, pisaUrl } from '../Data/pisaClient'
+import { FullDataBinary } from './dataBinaryToUrl'
+
+export async function pisaDataBinaryToUrl(binary: FullDataBinary) {
```
dcsaszar, line 4: * [ ] Missing return type. However, as long as dataBinaryToUrl is the only importing module, we could move it there as a private helper.

vite.config.ts
```diff
@@ -16,7 +16,7 @@ export default defineConfig(({ mode }) => {
 
   const PISA_URL = env.PISA_URL
   const enablePisa = !!PISA_URL
-  if (enablePisa) console.log('Pisa included in the build.')
+  if (enablePisa) console.log('Pisa is enabled.')
```
dcsaszar, line 19: * [ ] FYI I realized that in practice this message is noisy. Personally I'd like to get rid of it.

dcsaszar (approved):
Nice. Approved b/c mandatory feedback is about specifying return types, which could also be handled in a follow-up.

src/Data/toClientParams.ts
apepper: Why is this needed? The derived types look fine to me.

src/Data/toClientParams.ts
apepper: This code is inspired by the SDK.

src/utils/pisaDataBinaryToUrl.ts
apepper: Same as https://github.com/Scrivito/scrivito-portal-app/pull/381#discussion_r1578102346,

vite.config.ts
apepper: Let's see, if I remove it eventually.

src/Data/toClientParams.ts
dcsaszar: > This code is inspired by the SDK.

I don't understand. If the SDK would autoconvert `''` to `undefined` we wouldn't need this the fallback here at all.

apepper: It's inspired by `js/src/scrivito_sdk/data_integration/create_standard_api_connection.ts#L109`. This is actually the behavior of `provideDataClass` with `restApi`.

PR #391, opened 2024-05-03
Accept-Language for some pisa data classes.

apepper:
Plus: Language switch reloads the app.

### DIFF
--- a/src/utils/currentLanguage.ts
+++ b/src/utils/currentLanguage.ts
@@ -5,6 +5,6 @@ export function getCurrentLanguage(): undefined | string {
   return currentSiteLanguage === null ? 'en' : currentSiteLanguage
 }
 
-export function acceptLanguageHeader() {
+export function languageHeaders() {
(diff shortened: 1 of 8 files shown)


src/utils/currentLanguage.ts
```diff
@@ -4,3 +4,7 @@ export function getCurrentLanguage(): undefined | string {
   const currentSiteLanguage = Obj.root()?.language()
   return currentSiteLanguage === null ? 'en' : currentSiteLanguage
 }
+
+export function acceptLanguageHeader() {
```
dcsaszar, line 8: * [ ] FYI it's a misnomer as we return an object, which indicates plural. The red flag is code like: `headers: acceptLanguageHeader()`. Ideas:
  * rename to sth. like `languageHeaders`
  * return only the value `getCurrentLanguage() ?? 'en'`

dcsaszar (approved):
LGTM. Just some naming feedback.

dcsaszar (approved):

dcsaszar (approved):

PR #393, opened 2024-05-06
Partly baseUrl

apepper:
Reverts `SCRIVITO_PORTAL_APP_ROOT_CONTENT_ID` and `baseUrl` attribute consideration in `scrivitoSites.ts` (as introduced in #388).

Also improves `siteForUrl`.

Closes #389, since this is a different approach to the issue (dropping the functionality).


krishan (approved):
LGTM 

Solves the performance problems.

It won't work under node AFAICS (if it finds no `location`, it just does nothing). But I guess that's fine for now.

PR #395, opened 2024-05-06
Show loading indicator for data widgets

dcsaszar:
PR best viewed with whitespace off.

image

The main purpose is to indicate temporary issues connecting to backends.

As far as I can see we don't have any means of detecting permanent issues, because both known error cases
* PisaSales backend not available (TypeError)
* Not authorized as a user (status 401)

are transparently handled (retried) by the SDK.

One way we could provide more info to the user could be a pessimistic guess:
Add a timeout to the loader component and switch to some form of "Sorry, try again later" message.

### DIFF
--- a/src/Components/Loading.tsx
+++ b/src/Components/Loading.tsx
@@ -2,15 +2,15 @@ import { connect } from 'scrivito'
 import { getCurrentLanguage } from '../utils/currentLanguage'
 
 export const Loading = connect(function Loading() {
+  const message = getMessage()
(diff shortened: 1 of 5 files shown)


apepper (approved):
LGTM ([d0487ee](https://github.com/Scrivito/scrivito-portal-app/pull/395/commits/d0487ee69cd3e94e14c6c9b35acf84fc01f49b03))

Nice 🚀!

PR #399, opened 2024-05-07
CurrentUserDataItem: Show errors to user

apepper:
Screenshot 2024-05-07 at 14 06 16

### DIFF
--- a/src/Data/CurrentUser/CurrentUserDataItem.ts
+++ b/src/Data/CurrentUser/CurrentUserDataItem.ts
@@ -15,7 +15,7 @@ export const CurrentUser = provideDataItem('CurrentUser', {
     try {
       neoletterProfile = await neoletterClient().get('my/profile')
       if (!isNeoletterData(neoletterProfile)) {
-        throw new Error('Neoletter data is not in the expected format')
+        throw new Error('Invalid user profile')
       }
     } catch (error) {
       errorToast('Unable to connect to Neoletter', error)
@@ -80,9 +80,7 @@ async function pisaIds() {
   let whoAmI
   try {
     whoAmI = await pisaClient('whoami').get('')
-    if (!isWhoAmI(whoAmI)) {
-      throw new Error('Whoami data is not in the expected format')
-    }
+    if (!isWhoAmI(whoAmI)) throw new Error('Invalid user ID')
   } catch (error) {
     errorToast('Unable to connect to PisaSales', error)
     throw error


src/Data/CurrentUser/CurrentUserDataItem.ts
```diff
@@ -4,15 +4,22 @@ import { ensureString } from '../../utils/ensureString'
...
-    if (!isNeoletterData(neoletterProfile)) {
-      throw new Error('Neoletter data is not in the expected format')
+    let neoletterProfile
+    try {
+      neoletterProfile = await neoletterClient().get('my/profile')
+      if (!isNeoletterData(neoletterProfile)) {
+        throw new Error('Neoletter data is not in the expected format')
+      }
+    } catch (error) {
+      errorToast('Unable to connect to Neoletter', error)
```
dcsaszar, line 21: * [ ] The important message is that "A technical error occurred". Not sure if it should be passed as an argument. Currently all error toasts are technical errors, so we could move this to the helper instead of repeating it.
* [ ] Given the main message above I'm not sure if we have to pass in both an additional message and the error. It depends on how likely these errors are and how likely a custom message will help the user. "Unable to connect" may even be misleading. Maybe just `error.message` (which is already customizable if needed, like inside the try block) is good enough for technical breakdowns?

src/Data/CurrentUser/CurrentUserDataItem.ts
```diff
@@ -4,15 +4,22 @@ import { ensureString } from '../../utils/ensureString'
...
     if (!user) return {}
 
-    const neoletterProfile = await neoletterClient().get('my/profile')
-    if (!isNeoletterData(neoletterProfile)) {
-      throw new Error('Neoletter data is not in the expected format')
+    let neoletterProfile
+    try {
+      neoletterProfile = await neoletterClient().get('my/profile')
+      if (!isNeoletterData(neoletterProfile)) {
+        throw new Error('Neoletter data is not in the expected format')
```
dcsaszar, line 18: * [ ] Mentioning Neoletter may be confusing to visitors. How about something more neutral like "Invalid user profile" here and "Invalid user ID" for `whoami`?

src/Data/CurrentUser/errorToast.tsx
```diff
@@ -0,0 +1,23 @@
+import { toast } from 'react-toastify'
+import { ClientError } from 'scrivito'
+
+export function errorToast(title: string, error: unknown): void {
+  toast.error(
+    <>
+      <div>{title}</div>
+      {error instanceof ClientError ? (
+        <>
+          <div className="text-small">{error.message}</div>
```
dcsaszar, line 10: * [ ] FYI `<small>{error.message}</small>` also looks good and would even be more sematic.

dcsaszar (changes requested):
Nice so far. See comments for potential simplifications.

src/Data/CurrentUser/errorToast.tsx
```diff
@@ -0,0 +1,23 @@
...
+          <pre className="text-small text-muted">
+            {JSON.stringify(error.details, null, 1)}
```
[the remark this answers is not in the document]
apepper: I would actually like to see this for the regular user. I received already some screenshots, where stuff was not working - but without a console. The error.details would have helped me.

PR #400, opened 2024-05-07
Mobile menu order improvements

dcsaszar:
The order is open to further refinement, but at least it's no longer completely off.
Maybe unifying the menu item style on mobile is more important as a next step?

image


apepper (approved):
LGTM ([4305199](https://github.com/Scrivito/scrivito-portal-app/pull/400/commits/43051995f0b1c980534e5b2b25afb954b286a53e))

Very nice 🚀!

PR #402, opened 2024-05-08
DataPersonCard: Calculate user differently

apepper:
This also has the advantage that no headline is shown, if the user could not be found.

Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://workaround-datapersoncardwid.scrivito-portal-app.pages.dev/en/my-tynacoon?_scrivito_workspace_id=je9ba13d7a074d63&_scrivito_display_mode=view

### DIFF
--- a/src/Widgets/DataPersonCardWidget/DataPersonCardWidgetComponent.tsx
+++ b/src/Widgets/DataPersonCardWidget/DataPersonCardWidgetComponent.tsx
@@ -1,4 +1,4 @@
-import { connect, DataItem, provideComponent, useData } from 'scrivito'
+import { connect, ContentTag, DataItem, provideComponent } from 'scrivito'
 
 import { DataBinaryImage } from '../../Components/DataBinaryImage'
 import { ensureString } from '../../utils/ensureString'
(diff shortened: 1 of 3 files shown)


apepper:
REVIEW ([1178cfe](https://github.com/Scrivito/scrivito-portal-app/pull/402/commits/1178cfe976e5f47c425ee20ce9fb82511089fc90))

src/Widgets/DataPersonCardWidget/DataPersonCardWidgetComponent.tsx
```diff
@@ -1,26 +1,39 @@
...
   DataPersonCardWidget,
-  () => {
-    const dataScope = useData()
+  ({ widget }) => {
+    // TODO: Remove workaround, once #10883 is resolved
+    const name = ensureString(widget.get('attributeName'))
+    if (!name) return <EditorNote>Data is empty.</EditorNote>
 
-    if (dataScope.isEmpty()) return <EditorNote>Data is empty.</EditorNote>
+    const value = CurrentUser.get(name)
```
dcsaszar, line 20: * [ ] FYI instead of hardcoding this to CurrentUser and an `attributeName: 'string'` we could use `data: 'datalocator'`, let the user pick the data attribute, and access it by `value = useDataLocator(widget.get('data')).dataItemAttribute()?.get()`.
image

dcsaszar (approved):
See FYI for alternatives.

src/Widgets/DataPersonCardWidget/DataPersonCardWidgetComponent.tsx
apepper: Thanks for the pointer. `const dataScope = useDataLocator(widget.get('data'))` also works.

src/Widgets/DataPersonCardWidget/DataPersonCardWidgetEditingConfig.ts
```diff
@@ -5,4 +5,8 @@ import Thumbnail from './thumbnail.svg'
 provideEditingConfig(DataPersonCardWidget, {
   title: 'Data Person Card',
   thumbnail: Thumbnail,
+  properties: ['attributeName', 'headline'],
```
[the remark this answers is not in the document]
apepper: I would like to be able to edit the headline as an editor even if my current user does not have a sales representative.

PR #411, opened 2024-05-15
Allow to be framed by Scrivito preview deployments

apepper:
...and localhost:8090


markus-codewrangler (approved):
Nice! Works like a charm! 😍

PR #430, opened 2024-05-29
DataImageWidget

apepper:
Extracted `imageFromDataItem` option from `ImageWidget` to a dedicated `DataImageWidget`. Also make use of the datalocator attribute feature.

Needs content: https://edit-beta.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dataimagewidget.scrivito-portal-app.pages.dev/en/equipment?_scrivito_workspace_id=gae569e8ef77a2fc&_scrivito_display_mode=view


src/Widgets/DataImageWidget/DataImageWidgetEditingConfig.ts
```diff
@@ -0,0 +1,34 @@
...
+      values: [
+        { value: 'left', title: 'Left' },
+        { value: 'center', title: 'Center' },
+        { value: 'right', title: 'Right' },
+      ],
+    },
+    alternativeText: {
+      title: 'Alternative text',
+      description: 'Brief description of what the image is about.',
+    },
```
dcsaszar, line 21: * [ ] I think for a data image this makes no sense b/c at editing time the content of the image is unknown, so there is no decision basis. How about always using `alt=""`?

src/Widgets/DataImageWidget/DataImageWidgetClass.ts
```diff
@@ -0,0 +1,13 @@
+import { provideWidgetClass } from 'scrivito'
+
+export const DataImageWidget = provideWidgetClass('DataImageWidget', {
+  attributes: {
+    alignment: ['enum', { values: ['left', 'center', 'right'] }],
+    alternativeText: 'string',
+    data: 'datalocator',
+    roundCorners: 'boolean',
+    link: 'link',
```
dcsaszar, line 9: * [ ] FYI none of the existing widgets uses the `link` feature. We could drop it to make the widget simpler.

dcsaszar (approved):
OK so far. See comments for potential simplifications

src/Widgets/DataImageWidget/DataImageWidgetClass.ts
apepper: I would like to keep it.

src/Widgets/DataImageWidget/DataImageWidgetComponent.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { alignmentClassName } from '../../utils/alignmentClassName'
+import {
+  DataImageWidget,
+  DataImageWidgetInstance,
+} from './DataImageWidgetClass'
+import { isDataBinary } from '../../utils/dataBinaryToUrl'
+import { DataBinaryImage } from '../../Components/DataBinaryImage'
+
+provideComponent(DataImageWidget, ({ widget }) => {
+  const dataScope = useData()
```
kostia, line 23: You don't need to use the data here and forward it to `ImageComponent` – the data can be used in `ImageComponent` directly.

src/Widgets/DataImageWidget/DataImageWidgetComponent.tsx
```diff
@@ -0,0 +1,84 @@
...
+  dataScope: DataScope
+  widget: DataImageWidgetInstance
+}) {
+  const className = widget.get('roundCorners') ? 'rounded' : undefined
+
+  const dataItemAttribute = dataScope.dataItemAttribute()
+  if (!dataItemAttribute) return null
+
+  const objValue = dataItemAttribute.dataItem().obj()
+  if (objValue instanceof Obj) {
```
kostia, line 64: You don't need to check the class – `#obj` returns either an instance of `Scrivito.Obj` or nothing.

src/Widgets/DataImageWidget/DataImageWidgetComponent.tsx
```diff
@@ -0,0 +1,84 @@
...
+}: {
+  dataScope: DataScope
+  widget: DataImageWidgetInstance
+}) {
+  const className = widget.get('roundCorners') ? 'rounded' : undefined
+
+  const dataItemAttribute = dataScope.dataItemAttribute()
+  if (!dataItemAttribute) return null
+
+  const objValue = dataItemAttribute.dataItem().obj()
```
kostia, line 63: You can ask the scope for its item as well:
```ts
const obj = dataScope.dataItem()?.obj();
```

src/Widgets/DataImageWidget/DataImageWidgetComponent.tsx
```diff
@@ -77,22 +75,10 @@ const ImageComponent = connect(function ImageComponent({
...
   if (typeof attributeValue !== 'string') return null
 
-  return (
-    <img
-      src={attributeValue}
-      alt={widgetAlternativeText}
-      className={className}
-    />
-  )
+  return <img src={attributeValue} className={className} />
```
dcsaszar, line 83: * [ ] Leaving out `alt` is bad for accessibility. We should set `alt=""`. See https://www.w3.org/WAI/tutorials/images/decision-tree/

dcsaszar (changes requested):
See comment.

src/Widgets/DataImageWidget/DataImageWidgetComponent.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const dataItemAttribute = dataScope.dataItemAttribute()
+  if (!dataItemAttribute) return null
+
+  const objValue = dataItemAttribute.dataItem().obj()
+  if (objValue instanceof Obj) {
+    return (
+      <InPlaceEditingOff>
+        <ImageTag
+          content={objValue}
+          attribute={dataItemAttribute.attributeName()}
```
kostia, line 69: You could have asked the scope about its name as well:
```tsx
attribute={dataScope.attributeName()}
```

but unfortunately `DataScope#attributeName` returns `null` as default and not `undefined`. Which makes me realize that returning `undefined` in that case would be generally a better idea.

kostia (commented):
Nice 👍 

I've only looked at data-integration related aspects in `DataImageWidgetComponent.tsx`. Please find my feedback below.

kostia (approved):
Nice 👍

src/Widgets/DataImageWidget/DataImageWidgetComponent.tsx
```diff
@@ -80,5 +80,5 @@ const ImageComponent = connect(function ImageComponent({
 
   if (typeof attributeValue !== 'string') return null
 
-  return <img src={attributeValue} className={className} />
+  return <img src={attributeValue} className={className} alt="" />
```
dcsaszar, line 77: * [ ] FYI the component has another code path with `ImageTag`. This one should have the `alt=""`, too.

apepper: => https://github.com/Scrivito/scrivito-portal-app/pull/436

PR #438, opened 2024-06-06
DataBatchContext: Improve "hasMore"

apepper:
Avoids potentially inaccurate "dataScope.count()" function. It also potentially uses less requests.

Since scrivito internally first requests one item and then uses the continuation with another 10 items and a continuation `.take({ limit: limit + 1})` can be responded directly from the already loaded data.


dcsaszar (approved):
LGTM.

FYI The explanation for why this may be more efficient relies on a lot of internal details. To a naive reader, the old implementation looks more efficient.

PR #439, opened 2024-06-07
Icon & attachment attribute aware

apepper:
Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://icon-attachment-attribute-aw.scrivito-portal-app.pages.dev/en/my-tynacoon?_scrivito_workspace_id=jf6431200f480927&_scrivito_display_mode=view

### DIFF
--- a/src/Widgets/DataAttachmentsWidget/DataAttachmentsWidgetClass.ts
+++ b/src/Widgets/DataAttachmentsWidget/DataAttachmentsWidgetClass.ts
@@ -5,6 +5,7 @@ export const DataAttachmentsWidget = provideWidgetClass(
   {
     attributes: {
       attributeName: 'string',
+      data: 'datalocator',
       label: 'string',
     },
   },
--- a/src/Widgets/DataAttachmentsWidget/DataAttachmentsWidgetClass.ts
+++ b/src/Widgets/DataAttachmentsWidget/DataAttachmentsWidgetClass.ts
@@ -4,7 +4,6 @@ export const DataAttachmentsWidget = provideWidgetClass(
   'DataAttachmentsWidget',
   {
     attributes: {
-      attributeName: 'string',
       data: 'datalocator',
       label: 'string',
     },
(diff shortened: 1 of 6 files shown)


src/Widgets/index.ts
```diff
@@ -2,6 +2,7 @@ import { initNeoletterFormWidgets } from 'scrivito-neoletter-form-widgets'
 import.meta.glob(['./**/*WidgetClass.ts', './**/*WidgetComponent.tsx'], {
   eager: true,
 })
+import './migrateAttributeNameToData'
```
dcsaszar: * [ ] Why make this part of the code history? It looks like a one-shot migration that can be run on the console.

src/Widgets/index.ts
```diff
@@ -2,7 +2,6 @@ import { initNeoletterFormWidgets } from 'scrivito-neoletter-form-widgets'
 import.meta.glob(['./**/*WidgetClass.ts', './**/*WidgetComponent.tsx'], {
   eager: true,
 })
-import './migrateAttributeNameToData'
```
dcsaszar: * [ ] 😿 Drop the last two commits from the history?

dcsaszar (changes requested):
LGTM. I only briefly checked the content changes.

See comments for a potentially cleaner history.

src/Widgets/index.ts
apepper: It is a one-shot migration. But not a trivial one, that is why I wanted to have it at least in a file, that I can later restore if needed.

dcsaszar: How about a branch in a fork?
Making the script part of the history has some drawbacks:
* It creates noise in the history
* There are no guidelines for a reviewer for how to review code and/or functionality

apepper: I forced pushed a version without those two commits.

apepper:
REVIEW ([b771287](https://github.com/Scrivito/scrivito-portal-app/pull/439/commits/b77128778e50be836f3857ffb18ad98d67d7ece2))

dcsaszar (approved):
🏄

PR #440, opened 2024-06-07
Enforce passing a dependency array to useEffect

dcsaszar:
https://www.npmjs.com/package/@arabasta/eslint-plugin-require-useeffect-dependency-array

This takes away one degree of freedom but
this is virtually never used and most likely an error. See, for example, #264.

Feel free to close if the added complexity is not worth it. The endless loop was a logic mistake in the first place and there's no reliable way to prevent it. We always executed the effect (no dependency array) and always set a state in the callback.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,7 @@
 {
   "extends": [
     "eslint:recommended",
+    "plugin:@arabasta/require-useeffect-dependency-array/recommended-legacy",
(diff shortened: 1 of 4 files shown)


apepper (approved):
LGTM ([1926511](https://github.com/Scrivito/scrivito-portal-app/pull/440/commits/192651195ce1ee3ad761386382d51962e116b42c))

Nice 🚀! Thanks a lot!

PR #441, opened 2024-06-10
DataLabelWidget: Format numbers language aware

apepper:
Formats a numbers attribute according to the language of a site. To test you could use the `size` attribute of a Document.

### DIFF
--- a/src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
+++ b/src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
@@ -69,6 +69,10 @@ const AttributeValue = connect(function AttributeValue({
   }
   if (showAs === 'link') return <Link value={attributeValue} />
 
+  if (typeof attributeValue === 'number') {
+    return <NumberText value={attributeValue} />
+  }
+
   const value = localizeAttributeValue({
     dataClass: dataItemAttribute.dataClass(),
     attributeName: dataItemAttribute.attributeName(),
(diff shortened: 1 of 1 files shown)


src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
```diff
@@ -82,6 +86,10 @@ function Text({ value }: { value: unknown }) {
   return value ? value.toString() : localizeNotAvailable()
 }
 
+function NumberText({ value }: { value: number }) {
+  return new Intl.NumberFormat(getCurrentLanguage()).format(value)
```
dcsaszar, line 90: * [ ] FYI From experience users often prefer `{useGrouping: "min2"}` to keep four-digit numbers as a whole for better readability.

dcsaszar (approved):
Good, but see comment.

src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
apepper: This seems to be a fairly recent addition to the standard. At least according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat Safari only supports it with 16.4 or newer (which was released last year).

dcsaszar: A workaround for older browsers is `useGrouping: Math.abs(value) >= 10000`.

apepper:
REVIEW ([2c53be2](https://github.com/Scrivito/scrivito-portal-app/pull/441/commits/2c53be235281cfe372505821efe679eaa1460e9c))

PR #442, opened 2024-06-10
Logout now redirects to root page after logout

apepper:


src/Widgets/NavigationWidget/SubComponents/CurrentUserDropdown.tsx
```diff
@@ -68,13 +70,16 @@ export const CurrentUserDropdown = connect(function CurrentUserDropdown({
...
       <NavDropdown.Item
-        as={LinkTag}
         eventKey="MetaNavigation-LogOut"
         key="MetaNavigation-LogOut"
-        onClick={() => {
-          logout()
+        onClick={async () => {
+          const target = await load(() => {
+            const root = Obj.root()
+            return root ? urlFor(root) : undefined
```
dcsaszar, line 78: * [ ] FYI If we move the `urlFor` loading out of the click handler the button will be more responsive on clicking and the `onClick` callback would be simpler.

apepper:
REVIEW ([86c064a](https://github.com/Scrivito/scrivito-portal-app/pull/442/commits/86c064af47fb7c32c7a3e3543c0aad0aefad1b12))

Force pushed a new version to avoid conflicts. It (hopefully) contains all the feedback as well.

PR #444, opened 2024-06-11
Scrivito based currentLanguage

apepper:
Makes use of the newly introduced `Scrivito.currentLanguage()`.

### DIFF
--- a/src/Widgets/NavigationWidget/SubComponents/LanguageSwitch.tsx
+++ b/src/Widgets/NavigationWidget/SubComponents/LanguageSwitch.tsx
@@ -6,7 +6,6 @@ import {
   currentSiteId,
   ImageTag,
   InPlaceEditingOff,
-  Link,
   LinkTag,
   Obj,
 } from 'scrivito'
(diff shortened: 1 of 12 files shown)


src/Widgets/NavigationWidget/SubComponents/LanguageSwitch.tsx
```diff
@@ -44,11 +45,8 @@ export const LanguageSwitch = connect(function LanguageSwitch() {
         {versions.map(({ version, root }) => (
           <NavDropdown.Item
             key={root.id()}
-            href={
-              version
-                ? urlFor(version, { query: currentQuery() })
-                : urlFor(root)
-            }
+            as={LinkTag}
+            to={new Link({ obj: version || root, query: currentQuery() })}
```
dcsaszar, line 49: * [ ] Instead of using the workaround of `new Link` we should use the `to` and `params` props. I think this will even simplify `currentQuery`.

dcsaszar (changes requested):
Looks good so far. One thing can be improved, see comment.

apepper:
REVIEW ([2d7c071](https://github.com/Scrivito/scrivito-portal-app/pull/444/commits/2d7c0717a8f9a19ce7089b8ee0ab7fc728940cd5))

PR #449, opened 2024-06-12
Pisa url from content

apepper:
Instead of hardcoding the pisa endpoint, the endpoint is now read out from the content of the default homepage. This allows to run multiple instance much easier, since no app adjustments are needed anymore.

The "pisaUrl" content has already been set on the "golden master" tenant.

Still open todos:
* [x] Set `ENABLE_PISA` accordingly on Cloudflare
* [x] Think of, how to avoid the `pisaUrl` in the content dump. First idea: Adjust `dumpContent.ts` to filter out specific `pisaUrl` attribute. => https://github.com/Scrivito/scrivito-portal-app/pull/453

### DIFF
--- a/src/Data/pisaClient.ts
+++ b/src/Data/pisaClient.ts
@@ -1,7 +1,7 @@
 import { createRestApiClient, currentLanguage, load } from 'scrivito'
 import { ensureString } from '../utils/ensureString'
 
-export function pisaUrl(): string {
+export async function pisaUrl(): Promise<string> {
   const url = ensureString(import.meta.env.PISA_URL)
   if (!url) throw new Error('PISA_URL is not set!')
 
@@ -9,7 +9,7 @@ export function pisaUrl(): string {
 }
 
 export async function pisaClient(subPath: string) {
-  const url = `${pisaUrl()}/${subPath}`
+  const url = `${await pisaUrl()}/${subPath}`
 
   const language = await load(() => currentLanguage() ?? 'en')
   const headers = { 'Accept-Language': language }
--- a/src/Data/pisaClient.ts
+++ b/src/Data/pisaClient.ts
@@ -1,9 +1,15 @@
-import { createRestApiClient, currentLanguage, load } from 'scrivito'
-import { ensureString } from '../utils/ensureString'
+import { Obj, createRestApiClient, currentLanguage, load } from 'scrivito'
+import { isHomepage } from '../Objs/Homepage/HomepageObjClass'
 
 export async function pisaUrl(): Promise<string> {
-  const url = ensureString(import.meta.env.PISA_URL)
-  if (!url) throw new Error('PISA_URL is not set!')
+  const defaultRoot = await load(() => Obj.onSite('default').root())
+  if (!isHomepage(defaultRoot)) return never()
+
+  const url = defaultRoot.get('pisaUrl')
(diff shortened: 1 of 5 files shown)


src/Objs/Homepage/HomepageEditingConfig.ts
```diff
@@ -16,6 +16,11 @@ provideEditingConfig(Homepage, {
       title: 'Page description',
       description: 'Limit to 175, ideally 150 characters.',
     },
+    pisaUrl: {
+      title: 'PisaSales Portal API URL',
+      description:
+        'Only editable on the default site. Reload the app after changing this value.',
```
dcsaszar, line 22: * [ ] FYI how is the editor supposed to know which site is the "default site"? Maybe we should put the site's `contentTitle` here dynamically?
* [ ] FYI not sure if the average editor knows how to "reload the app". Maybe just remove this hint and assume a reload is the default action of any user if something doesn't work as expected?

src/Data/pisaClient.ts
```diff
@@ -1,9 +1,15 @@
...
 
 export async function pisaUrl(): Promise<string> {
-  const url = ensureString(import.meta.env.PISA_URL)
-  if (!url) throw new Error('PISA_URL is not set!')
+  const defaultRoot = await load(() => Obj.onSite('default').root())
+  if (!isHomepage(defaultRoot)) return never()
+
+  const url = defaultRoot.get('pisaUrl')
+  if (!url) {
+    console.log('Please configure a pisaUrl on the default homepage.')
```
dcsaszar, line 10: * [ ] FYI Not sure if this is helpful. It spams the console which typically is invisible to editors while the fix needs content changes. Would a validation make more sense?

dcsaszar (approved):
Fine in general. Editing support could be improved, see comments.

src/Data/pisaClient.ts
apepper: I think this is helpful. Ideally you only have to setup the pisaUrl once - when you initialize the tenant. Validations are not ideal, because the would only be executed, if something changed on the homepage - or may even prevent the initial content script from working properly.

This console log is intended for a developer who tries to find out, why the app is not loading e.g. the current user.

apepper:
> Set `ENABLE_PISA` accordingly on Cloudflare

I added the following to the build command:

```bash
ENABLE_PISA=$([ "$CF_PAGES_BRANCH" = "pisa" ] && echo true)
```

apepper:
REVIEW ([0d0742c](https://github.com/Scrivito/scrivito-portal-app/pull/449/commits/0d0742c2416766ee49ef16ad6378b8bc10ccce64))

src/Objs/Homepage/HomepageEditingConfig.ts
dcsaszar: True. Other ideas:
* Change the text to "English homepage"
* Hide the option on other sites
* Allow editing and show a validation error

dcsaszar (approved):

apepper:
REVIEW ([b500372](https://github.com/Scrivito/scrivito-portal-app/pull/449/commits/b500372fb6fe3e6b116491a93d0efe1f149107a0))

dcsaszar (approved):

PR #453, opened 2024-06-13
dumpContent: filter out pisa_url of root on default site

apepper:
Each new instance should setup its own pisa_url

### DIFF
--- a/src/dumpContent.ts
+++ b/src/dumpContent.ts
@@ -3,7 +3,13 @@ import fs from 'fs'
 import { configure, createRestApiClient } from 'scrivito'
 import { loadEnv } from 'vite'
 
-type ObjData = { _id: string; _widget_pool?: Record<string, WidgetData> }
+type ObjData = {
+  _id: string
+  _path?: string
+  _site_id?: string
+  _widget_pool?: Record<string, WidgetData>
+  pisa_url?: unknown
+}
 type WidgetData = Record<string, unknown>
 type SearchData = { continuation?: string; objs: ObjData[] }
 type BlobsData = { private_access: { get: { url: string } } }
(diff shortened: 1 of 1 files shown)


src/dumpContent.ts
```diff
@@ -88,6 +95,15 @@ function dumpManifest(objIds: string[]) {
   )
 }
 
+function preprocessObjData(objData: ObjData): ObjData {
```
dcsaszar, line 98: * [ ] FYI Naming: For the use-case at hand instead of the generic-sounding "preprocess" we could also name it more specifically like `clearSensitiveData`, `removePrivateData`, `ignorePerInstanceData`.

src/dumpContent.ts
```diff
@@ -88,6 +95,15 @@ function dumpManifest(objIds: string[]) {
...
+  // Ignore pisa_url - this should be set up in the new instance
```
dcsaszar, line 99: * [ ] FYI I think readers of this comment are not necessarily people setting up instances. Remove?

src/dumpContent.ts
```diff
@@ -88,6 +95,15 @@ function dumpManifest(objIds: string[]) {
...
+  if (objData._site_id === 'default' && objData._path === '/') {
+    delete objData.pisa_url
+  }
```
dcsaszar, line 102: * [ ] FYI The code could be simplified to drop everything named `pisa_url`. In my mental model we basically remove all blacklisted attributes everywhere. (The length of the list of names is one at the moment.)

dcsaszar (approved):
Behavior looks good. See comments for coding style remarks.

apepper:
REVIEW ([175c281](https://github.com/Scrivito/scrivito-portal-app/pull/453/commits/175c2814a3506013c4f0db0a114c83561af1ff80))

dcsaszar (approved):
🏄

PR #454, opened 2024-06-14
Use "currentLanguage" instead of individual "addressFormat"

apepper:
Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://rip-addressformat.scrivito-portal-app.pages.dev/de?_scrivito_workspace_id=t1d41069b4741cc2&_scrivito_display_mode=diff

### DIFF
--- a/src/Widgets/AddressWidget/AddressWidgetComponent.tsx
+++ b/src/Widgets/AddressWidget/AddressWidgetComponent.tsx
@@ -40,7 +40,7 @@ const Logo = connect(() => {
 
 const Address = connect(
   ({ addressWidget }: { addressWidget: AddressWidgetInstance }) => {
-    const localityRegionPostalCode = [
+    const postalCodeLocalityRegion = [
       addressWidget.get('locationPostalCode'),
       addressWidget.get('locationLocality'),
       addressWidget.get('locationRegion'),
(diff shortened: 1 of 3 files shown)


dcsaszar (changes requested):
Technically this is OK. But content-wise it's questionable.
Currently we have a German address in both the EN and DE site.
A German address should be written [in German order](https://howtogermany.com/household/utilities/mailing-a-letter-to-germany/#:~:text=German%20addresses%20differ%20from%20those,in%20Germany%20are%20five%20digits.) regardless of the sender.

Options:
* Put a US address in the EN page (adjust the content to the logic)
* Always use German format (simplify the code)

apepper:
REVIEW ([a29556a](https://github.com/Scrivito/scrivito-portal-app/pull/454/commits/a29556a71b9ee7a9d6d73ba62297d7aff04fe8ad))

src/Widgets/AddressWidget/AddressWidgetComponent.tsx
```diff
@@ -40,21 +40,11 @@ const Logo = connect(() => {
...
-        addressWidget.get('locationLocality'),
-        addressWidget.get('locationRegion'),
-        addressWidget.get('locationPostalCode'),
-      ]
-    }
+    const localityRegionPostalCode = [
+      addressWidget.get('locationPostalCode'),
+      addressWidget.get('locationLocality'),
+      addressWidget.get('locationRegion'),
+    ]
```
dcsaszar, line 47: * [ ] FYI the variable name is now out of order (component-wise)
* [ ] FYI now we only have one address line we could think about turning this into a single string attribute with any space-separated content. Not sure how much we/editors gain from making this structured.

dcsaszar (approved):
Nice.

PR #455, opened 2024-06-17
fix for border-radius and box-shadow / disabled local variables and reintroduced globally

agessler:


apepper:
Great. I incorporated the changes in a different way. Thank you.

PR #460, opened 2024-06-19
Homepage: Added drop shadow and rounded corners option

apepper:
This is one step to make the design also work in "flat" as well.

Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://flat-design-options.scrivito-portal-app.pages.dev/de?_scrivito_workspace_id=q6afa2eb26a9a594&_scrivito_display_mode=view

### DIFF
--- a/src/Components/DesignAdjustments.tsx
+++ b/src/Components/DesignAdjustments.tsx
@@ -9,10 +9,10 @@ export const DesignAdjustments = connect(function DesignAdjustments() {
   const styles: string[] = []
 
   const dropShadow = root.get('siteDropShadow')
-  if (!dropShadow) styles.push("--jr-card-box-shadow: 'none';")
+  if (!dropShadow) styles.push('--jr-card-box-shadow: none;')
 
   const roundedCorners = root.get('siteRoundedCorners')
-  if (!roundedCorners) styles.push("--jr-card-border-radius: '0';")
+  if (!roundedCorners) styles.push('--jr-card-border-radius: 0;')
 
   return (
     <Helmet>
--- a/src/Components/DesignAdjustments.tsx
+++ b/src/Components/DesignAdjustments.tsx
@@ -9,14 +9,10 @@ export const DesignAdjustments = connect(function DesignAdjustments() {
   const styles: string[] = []
 
   const dropShadow = root.get('siteDropShadow')
-  styles.push(
-    `--jr-card-box-shadow: ${
-      dropShadow ? '0 0.375rem 1.5rem 0 rgba(140, 152, 164, 0.25)' : 'none'
-    };`,
-  )
+  if (!dropShadow) styles.push("--jr-card-box-shadow: 'none';")
 
   const roundedCorners = root.get('siteRoundedCorners')
-  styles.push(`--jr-card-border-radius: ${roundedCorners ? '0.5rem' : '0'};`)
+  if (!roundedCorners) styles.push("--jr-card-border-radius: '0';")
 
   return (
     <Helmet>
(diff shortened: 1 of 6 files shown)


src/Objs/Homepage/HomepageEditingConfig.ts
```diff
@@ -56,9 +64,15 @@ provideEditingConfig(Homepage, {
         'sitePortalOverviewPage',
         'siteSearchResultsPage',
         'siteUserProfilePage',
+        'siteDropShadow',
+        'siteRoundedCorners',
```
dcsaszar, line 68: * [ ] FYI These are a good candidates for a dedicated style/appearance tab.

src/Components/DesignAdjustments.tsx
```diff
@@ -0,0 +1,27 @@
...
+    };`,
+  )
+
+  const roundedCorners = root.get('siteRoundedCorners')
+  styles.push(`--jr-card-border-radius: ${roundedCorners ? '0.5rem' : '0'};`)
+
+  return (
+    <Helmet>
+      {/* @ts-expect-error helmet bug: https://github.com/nfl/react-helmet/issues/344*/}
+      <body style={styles.join('\n')}></body>
```
dcsaszar, line 24: * [ ] FYI Newline feels weird for an attribute value. Space or blank?

src/Components/DesignAdjustments.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  const dropShadow = root.get('siteDropShadow')
+  styles.push(
+    `--jr-card-box-shadow: ${
+      dropShadow ? '0 0.375rem 1.5rem 0 rgba(140, 152, 164, 0.25)' : 'none'
+    };`,
+  )
+
+  const roundedCorners = root.get('siteRoundedCorners')
+  styles.push(`--jr-card-border-radius: ${roundedCorners ? '0.5rem' : '0'};`)
```
dcsaszar, line 19: Having too specific CSS rules in the JS is a smell. As long as we only deal with on/off switches, how about this:
With the [fallbacks in the CSS](https://github.com/Scrivito/scrivito-portal-app/pull/460/files#diff-ac1966f8adc171be5f6109146d471ed6872e1065794f6b96e6569dbd3d6a4b20R75-R76) we could simplify this to do one of two things:
1. Don't set the variable. This will effectively activate the default from the CSS.
2. Set it to a simple null value that works independently of the actual CSS. E.g, `unset`, `none`, `0px`. This will override and effectively disable the default from the CSS.

* [ ] FYI maybe rethink

dcsaszar (approved):
Cool!

src/Objs/Homepage/HomepageEditingConfig.ts
apepper: I might do this with the next PR, that will introduce custom colors.

src/assets/stylesheets/theme/box-comment.scss
```diff
@@ -6,7 +6,7 @@
     min-width: 76px;
     img {
       width: 60px;
-      border-radius: 5px;
+      border-radius: var(--jr-card-border-radius);
```
dcsaszar, line 9: * [ ] FYI For variables we use in other places than cards the `--jr-card-` prefix is weird.

dcsaszar (changes requested):
See comments.

src/assets/stylesheets/theme/box-comment.scss
apepper: I'll address this in a follow up with a review of Alex G.

dcsaszar (approved):

PR #468, opened 2024-06-24
style color picker for light and dark mode

agessler:
dark
light


apepper:
Awesome! Incorporated in #469.

PR #470, opened 2024-06-24
Drop two workarounds

dcsaszar:


src/config/history.ts
```diff
@@ -13,7 +13,7 @@ export function configureHistory(): void {
 
 let currentHistory: BrowserHistory | undefined
 
-export function getHistory(): BrowserHistory | undefined {
+function getHistory(): BrowserHistory | undefined {
```
apepper, line 16: I would prefer the following:

```suggestion
/** @public */
export function getHistory(): BrowserHistory | undefined {
```
This way knip does not complain and the developer can use `getHistory` if needed.

apepper (commented):
FEEDBACK ([fd56f8e](https://github.com/Scrivito/scrivito-portal-app/pull/470/commits/fd56f8e587c7bc988344b14fd4dd80d8112f8e3a))

Nice 🚀. Please see my comment.

src/config/history.ts
dcsaszar: 1. That wasn't obvious from the Knip report and I did the most straightforwad thing. How will future developers handle similar cases?

2. As long as the Portal app doesn't demonstrate how and why to use the history the developer can't really use it except by scanning the codebase and some thinking.
If a developer does that anyway they may easily export `getHistory()`if needed.
  ("developer" includes "Portal App developer". As soon as we add logic that needs the history we will export it again.)

WDYT?

apepper: I can't speak for all "future developers" 😸.

Currently the whole purpose of `getHistory` is to provide the reference to exactly the same history instance as `configureHistory`. If this is no longer a exported method we could simplify `configureHistory` to just use one history and not remember it. But I would like to keep `getHistory`, because we occasionally needed it in the past.

Also regarding "scanning". At least my IDE suggests this method automatically, if I type "history" (see image):
Screenshot 2024-06-25 at 13 21 53

dcsaszar: I'm on the YAGNI side here but feel free to insist on having it available.
* [ ] Yes please `@public`
* [ ] Something else

apepper: > Yes please @public

apepper (approved):
LGTM ([00cf254](https://github.com/Scrivito/scrivito-portal-app/pull/470/commits/00cf254d81a87434108f3a67f64b197800df45ca))

PR #471, opened 2024-06-25
User: Give a hint, why one can't index all users

apepper:


src/Data/User/Pisa/pisaUserDataClass.ts
```diff
@@ -1,24 +1,16 @@
...
-
-        return {
-          results: await Promise.all(results.map(postProcessUserData)),
-          continuation,
-          count,
-        }
+      index: async () => {
+        throw new DataConnectionError(
+          'Pisa does not support "index" for users due to security considerations.',
+        )
```
dcsaszar, line 13: * [ ] I tried this out with a data list page and a data group with a list. I only see "Editor note: Data is empty.". Is this message supposed to be visible? Where? If it is:
* [ ] Pisa -> PisaSales, but:
* [ ] I think an editor will not understand what `"index" for users` means. Can we rephrase this in more explanatory way? Something along the lines of "Listing users is not supported due to data protection reasons."

apepper: > I only see "Editor note: Data is empty.". Is this message supposed to be visible? Where?

With the current SDK version this message is not visible. But it will become visible with the next version of the SDK again (see https://github.com/Scrivito/scrivito-portal-app/pull/431).

PR #484, opened 2024-07-02
Latest typescript packages

apepper:
Installed via:
```
npm i -D typescript @typescript-eslint/parser @typescript-eslint/eslint-plugin @cloudflare/workers-types @types/node
```

### DIFF
--- a/src/Widgets/ColumnContainerWidget/ColumnsEditorTab.tsx
+++ b/src/Widgets/ColumnContainerWidget/ColumnsEditorTab.tsx
@@ -398,8 +398,20 @@ class GridLayoutEditor extends React.Component<
     if (deltaColSize === 0) return
 
     const newGrid = [...this.props.currentGrid]
-    newGrid[colIndex] = newGrid[colIndex]! + deltaColSize
-    newGrid[colIndex + 1] = newGrid[colIndex + 1]! - deltaColSize
+
+    const nextColIndex = colIndex + 1
+    const previousColIndexValue = newGrid[colIndex]
+    const previousColNextIndexValue = newGrid[nextColIndex]
+
+    if (
+      typeof previousColIndexValue !== 'number' ||
+      typeof previousColNextIndexValue !== 'number'
+    ) {
+      return
+    }
+
+    newGrid[colIndex] = previousColIndexValue + deltaColSize
+    newGrid[nextColIndex] = previousColNextIndexValue - deltaColSize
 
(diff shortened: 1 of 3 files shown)


src/Widgets/ColumnContainerWidget/ColumnsEditorTab.tsx
```diff
@@ -398,8 +398,8 @@ class GridLayoutEditor extends React.Component<
     if (deltaColSize === 0) return
 
     const newGrid = [...this.props.currentGrid]
-    newGrid[colIndex] += deltaColSize
-    newGrid[colIndex + 1] -= deltaColSize
+    newGrid[colIndex] = newGrid[colIndex]! + deltaColSize
+    newGrid[colIndex + 1] = newGrid[colIndex + 1]! - deltaColSize
```
apepper, line 402: Without this change the latest typescript version prints:

```
src/Widgets/ColumnContainerWidget/ColumnsEditorTab.tsx:401:5 - error TS2532: Object is possibly 'undefined'.

401     newGrid[colIndex] += deltaColSize
        ~~~~~~~~~~~~~~~~~

src/Widgets/ColumnContainerWidget/ColumnsEditorTab.tsx:402:5 - error TS2532: Object is possibly 'undefined'.

402     newGrid[colIndex + 1] -= deltaColSize
        ~~~~~~~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/Widgets/ColumnContainerWidget/ColumnsEditorTab.tsx:401
```

dcsaszar (approved):
I see that we need to do something b/c TS doesn't narrow the length of arrays.
However, because of the complexity of the code it's not obvious for the reader when looking at the drag handler why the array indices are guaranteed to exist. The correctness can only be checked by reading and understanding the caller.
This may cause future maintenance effort or even subtle bugs, which is what `noUncheckedIndexedAccess` is all about.
We could apply the "please add a comment for casts" rule but this wouldn't fix the issue mentioned above.
Actually I'm nor sure we can ensure valid handler arguments at all times b/c of concurrent updates and asynchronous handler calls.
* [x] FYI So how about we make the handler failsafe if used with unexpected arguments? E.g., we could exit early (do nothing) if there is a non-existing index.

```tsx
    const newGrid = [...this.props.currentGrid]
    const nextColIndex = colIndex + 1
    if (!newGrid[colIndex] || !newGrid[nextColIndex]) return
```

PR #485, opened 2024-07-02
<main> refactoring

apepper:
Motivation: There might be multiple layout components rendered before the actual "main" content.

Diff best viewed without whitespace changes.

### DIFF
--- a/src/Objs/Homepage/HomepageComponent.tsx
+++ b/src/Objs/Homepage/HomepageComponent.tsx
@@ -2,7 +2,5 @@ import { ContentTag, provideComponent } from 'scrivito'
 import { Homepage } from './HomepageObjClass'
 
 provideComponent(Homepage, ({ page }) => (
-  <main id="main">
-    <ContentTag tag="div" content={page} attribute="body" />
-  </main>
+  <ContentTag tag="main" id="main" content={page} attribute="body" />
 ))
--- a/src/Objs/Homepage/HomepageComponent.tsx
+++ b/src/Objs/Homepage/HomepageComponent.tsx
@@ -2,5 +2,7 @@ import { ContentTag, provideComponent } from 'scrivito'
 import { Homepage } from './HomepageObjClass'
 
 provideComponent(Homepage, ({ page }) => (
-  <ContentTag tag="div" content={page} attribute="body" />
+  <main id="main">
+    <ContentTag tag="div" content={page} attribute="body" />
+  </main>
 ))
--- a/src/Objs/Product/ProductComponent.tsx
+++ b/src/Objs/Product/ProductComponent.tsx
@@ -23,7 +23,7 @@ provideComponent(Product, ({ page }) => {
     .map(toPlainParameter)
 
   return (
-    <>
+    <main id="main">
       <section className="bg-light-grey py-4">
         <div className="container">
           <div className="row align-items-stretch">
@@ -170,7 +170,7 @@ provideComponent(Product, ({ page }) => {
           </div>
         </div>
       </section>
-    </>
+    </main>
   )
 })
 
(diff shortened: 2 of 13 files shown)


src/Objs/Homepage/HomepageComponent.tsx
```diff
@@ -2,5 +2,7 @@ import { ContentTag, provideComponent } from 'scrivito'
 import { Homepage } from './HomepageObjClass'
 
 provideComponent(Homepage, ({ page }) => (
-  <ContentTag tag="div" content={page} attribute="body" />
+  <main id="main">
+    <ContentTag tag="div" content={page} attribute="body" />
+  </main>
```
dcsaszar, line 7: * [ ] FYI We could simplify this to `<ContentTag tag="main" id="main"...`

src/Objs/SubnavigationOverview/SubnavigationOverviewComponent.tsx
```diff
@@ -2,5 +2,7 @@ import { ContentTag, provideComponent } from 'scrivito'
 import { SubnavigationOverview } from './SubnavigationOverviewObjClass'
 
 provideComponent(SubnavigationOverview, ({ page }) => (
-  <ContentTag content={page} attribute="body" />
+  <main id="main">
+    <ContentTag content={page} attribute="body" />
+  </main>
```
dcsaszar, line 7: * [ ] FYI Same as other comment

dcsaszar (approved):
See comment for potential improvements.

src/Objs/Product/ProductComponent.tsx
```diff
@@ -23,7 +23,7 @@ provideComponent(Product, ({ page }) => {
     .map(toPlainParameter)
 
   return (
-    <>
+    <main id="main">
```
apepper, line 26: This seems to break the design, at least on https://main-tag-refactoring.scrivito-portal-app.pages.dev/de/produkte-ed043d1606eaf539 . Any ideas?

src/Objs/SubnavigationOverview/SubnavigationOverviewComponent.tsx
```diff
@@ -2,5 +2,5 @@ import { ContentTag, provideComponent } from 'scrivito'
 import { SubnavigationOverview } from './SubnavigationOverviewObjClass'
 
 provideComponent(SubnavigationOverview, ({ page }) => (
-  <ContentTag content={page} attribute="body" />
+  <ContentTag tag="main" id="main" content={page} attribute="body" />
```
apepper, line 5: Same here: https://main-tag-refactoring.scrivito-portal-app.pages.dev/de/mein-tynacoon-bd53fa4ef38c1c40

src/Objs/Product/ProductComponent.tsx
dcsaszar: The problem appears to be flex layout. Before, every `section` inside `#root` was a flex child, now the whole `#main` wraps several sections, which leads to different flex heights. We must keep `flex` to keep the footer at the bottom.
How about setting `section { flex-shrink: 0; }`?

dcsaszar: After a discussion with Alex G. there are two options:
1. We move the top sections which are no longer part of `#main` to the `header`. Note: This would be the best solution from a semantic point of view.
2. If this is not easily possible, we'd remove the "#root 100% height" rules. This moves the footer from the bottom of the window to the bottom of the content, but we don't consider the footer position critical.

apepper: "We move the top sections which are no longer part of #main to the header. " is my goal - once I'm done will all refactorings.

apepper: I'll disable "#root 100% height" for now.

apepper:
REVIEW ([34e5cdd](https://github.com/Scrivito/scrivito-portal-app/pull/485/commits/34e5cdd486c3d2606c7633a67828c03eba13ce8d))

PR #492, opened 2024-07-05
BreadcrumbWidget

apepper:
To be later used in layoutHeader to replace hardcoded `<Breadcrumb />` in SubnavigationOverviewLayout and ProductOverviewLayout.

Thumbnail will be added later.

### DIFF
--- a/src/Widgets/BreadcrumbWidget/BreadcrumbWidgetEditingConfig.ts
+++ b/src/Widgets/BreadcrumbWidget/BreadcrumbWidgetEditingConfig.ts
@@ -1,0 +1,6 @@
+import { provideEditingConfig } from 'scrivito'
+import { BreadcrumbWidget } from './BreadcrumbWidgetClass'
+
+provideEditingConfig(BreadcrumbWidget, {
+  title: 'Breadcrumb',
+})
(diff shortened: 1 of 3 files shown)


src/Widgets/BreadcrumbWidget/BreadcrumbWidgetEditingConfig.ts
```diff
@@ -0,0 +1,6 @@
+import { provideEditingConfig } from 'scrivito'
+import { BreadcrumbWidget } from './BreadcrumbWidgetClass'
+
+provideEditingConfig(BreadcrumbWidget, {
+  title: 'Breadcrumb',
```
dcsaszar, line 5: * [ ] FYI terminology: it's either "Breadcrumbs" or "Breadcrumb Navigation". In case we'd like to improve, please keep class names/filenames in sync.

dcsaszar (approved):
LGTM, but see potential wording improvement.

src/Widgets/BreadcrumbWidget/BreadcrumbWidgetEditingConfig.ts
apepper: The singular "Breadcrumb" is also in use, e.g. https://getbootstrap.com/docs/5.3/components/breadcrumb/ or https://developer.mozilla.org/en-US/docs/Glossary/Breadcrumb. I'll leave it as is.

PR #493, opened 2024-07-05
PageTitleWidget

apepper:
To be later used in layoutHeader to replace hardcoded current page title banner in SubnavigationOverviewLayout and ProductOverviewLayout.

Thumbnail will be added later.


src/Widgets/CurrentPageTitleWidget/CurrentPageTitleWidgetEditingConfig.ts
```diff
@@ -0,0 +1,28 @@
+import { provideEditingConfig } from 'scrivito'
+import { CurrentPageTitleWidget } from './CurrentPageTitleWidgetClass'
+
+provideEditingConfig(CurrentPageTitleWidget, {
+  title: 'Current Page Title',
```
dcsaszar, line 5: * [ ] FYI wording: I think the term "Current" is redundant. What would be a different in a "Page Title" widget?
* [ ] FYI not sure if "Page Title" is clear enough, b/c the purpose is limited to header captions. Ideas:
  * Rename to, e.g., "Title Caption"
  * Add a `description`

dcsaszar (approved):
Nice, see comments.

src/Widgets/CurrentPageTitleWidget/CurrentPageTitleWidgetEditingConfig.ts
apepper: Once this "CurrentPageTitleWidget" is added to e.g. a layoutHeader or layoutFooter widgetlist it will always show the title of the currently displayed page _not_ the title of the page, that defines the layoutHeader/layoutFooter. That's why I added the word "current" to the name. But I'm open to change that.

dcsaszar: If we stick with WYSIWYG editing, from the editor's point of view they will not really know/easily see where the widget is stored technically. So the concept of
> the page, that defines the layoutHeader/layoutFooter
doesn't really matter in my opinion.

apepper:
REVIEW ([7c09478](https://github.com/Scrivito/scrivito-portal-app/pull/493/commits/7c0947818fae5ab551dcd79b69a307d2b624720d))

PR #494, opened 2024-07-05
Make sure shopping cart is submitted only once

dcsaszar:

PR #495, opened 2024-07-08
Full-screen submit indicator

dcsaszar:
See https://github.com/Scrivito/scrivito-portal-app/pull/494#discussion_r1668561059

For now, the visuals are a bit more intrusive (dark overlay) compared to what we had before.
Depending on the overlay appearance we choose, we could consider getting rid of styling forms (`form-loading`) and buttons (`disabled`) differently while submitting.

We use a hidden dialog b/c a dialog background apparently can't be made transparent.

### DIFF
--- a/src/Components/Submitting.tsx
+++ b/src/Components/Submitting.tsx
@@ -1,14 +1,22 @@
+import { useEffect, useRef } from 'react'
 import { Modal } from 'react-bootstrap'
 import { connect } from 'scrivito'
 
 export const Submitting = connect(function Submitting() {
+  const dialogRef = useRef<HTMLDialogElement>(null)
+  useEffect(() => dialogRef.current?.showModal(), [])
+
   return (
-    <Modal
-      backdrop="static"
-      centered
-      dialogAs={() => <div className="loader" />}
-      fullscreen
-      show
-    />
+    <>
+      <dialog className="d-none" ref={dialogRef} />
+      <Modal
+        backdrop="static"
+        centered
+        dialogAs={() => <div className="loader" />}
+        fullscreen
+        keyboard={false}
+        show
+      />
+    </>
   )
 })
--- a/src/Components/Submitting.tsx
+++ b/src/Components/Submitting.tsx
@@ -1,0 +1,14 @@
+import { Modal } from 'react-bootstrap'
+import { connect } from 'scrivito'
+
+export const Submitting = connect(function Submitting() {
+  return (
+    <Modal
+      backdrop="static"
+      centered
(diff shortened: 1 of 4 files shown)


src/Components/Submitting.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Modal } from 'react-bootstrap'
+import { connect } from 'scrivito'
+
+export const Submitting = connect(function Submitting() {
+  const dialogRef = useRef<HTMLDialogElement>(null)
+  useEffect(() => dialogRef.current?.showModal(), [])
+
+  return (
+    <>
+      <dialog className="d-none" ref={dialogRef} />
```
apepper, line 11: FMI: Why is `showModal` etc needed?

src/Components/Submitting.tsx
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useRef } from 'react'
+import { Modal } from 'react-bootstrap'
+import { connect } from 'scrivito'
+
+export const Submitting = connect(function Submitting() {
```
apepper, line 5: How about renaming this component to something like `LoaderModal` or `FullscreenLoader`? My gut feeling tells me, that this could be used in other then submitting situations as well.

apepper (commented):
FEEDBACK ([34be7db](https://github.com/Scrivito/scrivito-portal-app/pull/495/commits/34be7dbc85a9660520122d78516389cf4e06c31c))

Nice 🚀.

Please see my comments.

src/Components/Submitting.tsx
dcsaszar: This is how dialog works. From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#open
> It is recommended to use the .show() or .showModal() method to render dialogs, rather than the open attribute. If a <dialog> is opened using the open attribute, it is non-modal.

apepper: But isn't this the responsibility of `Modal` from `react-bootstrap`?

dcsaszar: 1. By only using Modal I was not able to prevent clicking a link if the spinner was inside a link.
2. By only using dialog I was not able to easily get the visuals of Modal.

I think we could make it work with 2 (use only dialog and a div, get rid of Modal) if we invest in dedicated CSS.

apepper (approved):
LGTM ([c937f1e](https://github.com/Scrivito/scrivito-portal-app/pull/495/commits/c937f1e51267a6250d64759dd8c1673d5be434d8))

Nice 🚀!

PR #500, opened 2024-07-10
Improved accessibility

apepper:
This PR fixes some errors, that where discovered by https://wave.webaim.org/extension/ .

Needs content: ~https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://hide-background-images.scrivito-portal-app.pages.dev/en?_scrivito_workspace_id=v8c6b62913247647&_scrivito_display_mode=view~ https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://hide-background-images.scrivito-portal-app.pages.dev/en?_scrivito_workspace_id=s7ce7a5a0255e874&_scrivito_display_mode=view


src/Widgets/DataPersonCardWidget/DataPersonCardWidgetComponent.tsx
```diff
@@ -66,8 +67,8 @@ const PersonCard = connect(function PersonCard({
               <tbody>
                 {email ? (
                   <tr>
-                    <th className="align-top">
-                      <i className="bi bi-envelope"></i>
+                    <th className="align-top" aria-label={getLocalizedEmail()}>
```
dcsaszar, line 70: * [ ] FYI This looks more like a workaround than an actual solution. 

The complaint is:

> Empty table header
> A <th> (table header) contains no text.

The question is: should we present the email address as a table from a semantic point of view?

From https://www.w3.org/WAI/tutorials/tables/tips/:

> Tables for Layout: Tables should not be used for layout purposes. Use Cascading Style Sheets (CSS) for layout. If there are already layout tables present, don’t use structural elements (like <th> or <caption>) and attributes discussed in this tutorial, and do add role="presentation" to the <table> element.

dcsaszar (changes requested):
Nice to see accessibility improved. See comments for some pointers.

dcsaszar:
* [ ] FYI maybe we should also rethink the description of the "alternative text" attribute. It currently reads
> Brief description of what the image is about.

Suggestion by ChatGPT:

> For images containing text, provide alternative text based on the text's function and visibility. Use descriptive alternative text for meaningful images and leave it empty for decorative ones.

src/Widgets/NavigationWidget/SubComponents/Brand.tsx
```diff
@@ -15,7 +17,7 @@ export const Brand = connect(function Brand({
           content={root}
           attribute="siteLogoDark"
           className="navbar-brand-logo"
-          alt="Logo"
+          alt={alt}
```
dcsaszar, line 20: * [ ] Why don't we use the `siteLogoDark`'s alternative text? This way, editors only have to set the alt text once.
* [ ] FYI checking this I realized we have an unused `siteLogoLight` setting.

src/Widgets/AddressWidget/AddressWidgetClass.ts
```diff
@@ -3,6 +3,7 @@ import { provideWidgetClass } from 'scrivito'
 export const AddressWidget = provideWidgetClass('AddressWidget', {
   attributes: {
     showLogo: 'boolean',
+    brandName: 'string',
```
dcsaszar, line 6: * [ ] Now editors have to set the same alt text three times. I don't think the added text attribute is worth it.

dcsaszar (changes requested):
Nice. Now I'm only worried about the two new places for setting the same alt text.

apepper:
REVIEW ([51d3ee1](https://github.com/Scrivito/scrivito-portal-app/pull/500/commits/51d3ee1d4fcc1d7bf01319f8589f4b3e66a515b5))

src/Widgets/NavigationWidget/SubComponents/Brand.tsx
apepper: siteLogoLight will be addressed in a different PR.

dcsaszar (approved):

PR #517, opened 2024-07-17
Fix ServiceObjectDocument editing config labels

dcsaszar:
The term "equipment" is used in the visitor presentation layer only. Here, we're at the backend data layer.


apepper (approved):
LGTM ([e43064c](https://github.com/Scrivito/scrivito-portal-app/pull/517/commits/e43064c26606578b8959155d0657fdd3e30dfa6b))

PR #518, opened 2024-07-19
Extend multi-site config to allow for isolated sites

dcsaszar:
In order to inform the app about the default site and its language versions, a new optional env `SCRIVITO_ROOT_CONTENT_ID` is added.
In order to work properly in a multi-site instance with isolated sites, root pages must have a `baseUrl` set.

### Todo
Isolated sites (or language versions) which share a URL prefix with another site/version don't work yet, e.g.,

```
https://example.com/en
https://example.com/en-US
```

### DIFF
--- a/src/config/scrivitoSites.ts
+++ b/src/config/scrivitoSites.ts
@@ -34,23 +34,20 @@ export function siteForUrl(
   }
 
   const baseAppUrl = getBaseAppUrl()
-  if (!baseAppUrl) return
-
   const language = languageForUrl(url, baseAppUrl)
-  if (!language) return
+  const languageSite = language
+    ? appWebsites().and('_language', 'equals', language).first()
+    : undefined
+  const languageSiteId = languageSite?.siteId()
 
-  const languageSite = appWebsites()
-    .and('_language', 'equals', language)
-    .first()
-  if (!languageSite) return
-
-  const languageSiteId = languageSite.siteId()
   if (!languageSiteId) return
 
   return { baseUrl: `${baseAppUrl}/${language}`, siteId: languageSiteId }
 }
 
-function languageForUrl(url: string, baseAppUrl: string) {
+function languageForUrl(url: string, baseAppUrl?: string) {
+  if (!baseAppUrl) return
+
   const regex = new RegExp(`^${baseAppUrl}\\/(?<lang>[a-z]{2})([?/]|$)`)
   return regex.exec(url)?.groups?.lang
 }
--- a/src/config/scrivitoSites.ts
+++ b/src/config/scrivitoSites.ts
@@ -1,4 +1,12 @@
-import { Obj, currentSiteId, getInstanceId, load, navigateTo } from 'scrivito'
+import {
+  Link,
+  Obj,
+  currentSiteId,
+  getInstanceId,
+  load,
+  navigateTo,
+  urlFor,
+} from 'scrivito'
 import { isMultitenancyEnabled } from './scrivitoTenants'
 
 const location = typeof window !== 'undefined' ? window.location : undefined
@@ -14,7 +22,9 @@ export function baseUrlForSite(siteId: string): string | undefined {
   const siteRoot = Obj.onSite(siteId).root()
   if (!siteRoot) return
 
-  if (rootContentId && siteRoot.contentId() !== rootContentId) return
+  if (rootContentId && siteRoot.contentId() !== rootContentId) {
+    return baseUrlsFor(siteRoot)[0]
+  }
 
   const baseAppUrl = getBaseAppUrl()
   if (!baseAppUrl) return
@@ -40,7 +50,7 @@ export function siteForUrl(
     : undefined
   const languageSiteId = languageSite?.siteId()
 
-  if (!languageSiteId) return
+  if (!languageSiteId) return findSiteForUrl(url)
 
   return { baseUrl: `${baseAppUrl}/${language}`, siteId: languageSiteId }
 }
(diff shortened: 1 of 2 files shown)


src/config/scrivitoSites.ts
```diff
@@ -1,7 +1,16 @@
...
+  currentSiteId,
+  getInstanceId,
+  load,
+  navigateTo,
+  urlFor,
+} from 'scrivito'
 import { isMultitenancyEnabled } from './scrivitoTenants'
 
 const location = typeof window !== 'undefined' ? window.location : undefined
+const rootContentId = import.meta.env.SCRIVITO_ROOT_CONTENT_ID
```
apepper, line 13: High level feedback: I would prefer, that there is _always_ a `rootContentId`. Since we want to run this in a multi-tenancy fashion custom env variables is not an option (or only an option if it applies to _all_ tenants).

apepper (changes requested):
FEEDBACK ([38469f3](https://github.com/Scrivito/scrivito-portal-app/pull/518/commits/38469f30a9d7826fe050a613b374268129fef343))

Nice start 🚀!

I left some high-level feedback. Please address it first, before I do a more thorough review.

src/config/scrivitoSites.ts
dcsaszar: Is this requirement still conditional depending on `isMultitenancyEnabled`?
What if the Portal App is used as a baseline for a new project without a multi-tenant requirement?

apepper: > Is this requirement still conditional depending on `isMultitenancyEnabled`?

No. I think these are two independent things. The rootContentId is used to ensure that in _one_ tenant you can have multiple sites where some sites are _not_ rendered by the scrivito portal app. `isMultitenancyEnabled` is to tell the scrivito portal app, that it can render _multiple_ tenants.

> What if the Portal App is used as a baseline for a new project without a multi-tenant requirement?

In that case they would start out with the initial content, that contains the rootContentId. If they have other requirements in the project they are free to adjust the code.

src/config/scrivitoSites.ts
```diff
@@ -1,8 +1,19 @@
...
+  load,
+  navigateTo,
+  urlFor,
+} from 'scrivito'
 import { isMultitenancyEnabled } from './scrivitoTenants'
 
 const location = typeof window !== 'undefined' ? window.location : undefined
 
+const rootContentId: string =
+  import.meta.env.SCRIVITO_ROOT_CONTENT_ID || 'c2a0aab78be05a4e'
```
apepper, line 15: How about
```suggestion
const rootContentId =
  ensureString(import.meta.env.SCRIVITO_ROOT_CONTENT_ID) || 'c2a0aab78be05a4e'
```

This way we ensure, that this is an actual string and not just claim it is.

src/config/scrivitoSites.ts
```diff
@@ -31,27 +46,56 @@ export function siteForUrl(
...
+  const site = sites[0]
+  const siteId = site?.siteId()
+  if (!site || !siteId) return
 
-  return { baseUrl: `${baseAppUrl}/${language}`, siteId: languageSiteId }
+  const baseUrl = baseUrlsFor(site)[0]
+  if (!baseUrl) return
+
+  return { baseUrl, siteId }
+}
```
apepper, line 83: Do I understand this correctly, that `baseUrl` can be potentially have many URLs and that this code will "normalize" the `baseUrl` to the first in the list?

apepper: If we don't need the normalization we could potentially simplify this code.

apepper (commented):
FEEDBACK ([26a47aa](https://github.com/Scrivito/scrivito-portal-app/pull/518/commits/26a47aa463144294d1da4e5fc447db0e937f8f4e))9

Good stuff 🚀! Please see my comments.

src/config/scrivitoSites.ts
dcsaszar: `findSiteForUrl` is the fallback for sites that don't match any of the app sites/versions.
* Yes, such a site may have more than one `baseUrl` (for example, prod and staging).
* Yes, we use the first URL as the primary URL, in order to avoid a mismatch between `baseUrlForSite` and `siteForUrl`.

According to this
>  If the site has the domain all to itself (like in the above example), it is sufficient for siteForUrl to return this ID. Otherwise, if two or more sites share a domain (like in the example below), siteForUrl should return the site’s ID as well as its base URL.

we should always return the base URL, because we don't have control over the content.

Under some circumstances Scrivito will complain if the base URLs of the two callbacks don't match. Unfortunately the docs don't touch these cases, so I'm not sure about potential risks. I could try it out, but even if it works now, will it continue to do so in the future?

src/config/scrivitoSites.ts
dcsaszar: I went with what we get from `import.meta.env`. This should be good enough for an equals comparison.

src/config/scrivitoSites.ts
apepper: If I'm not mistaken this function can be simplified to:

```ts
function findSiteForUrl(url: string) {
  for (const site of appWebsites().toArray()) {
    const baseUrls = baseUrlsFor(site)
    const siteId = site.siteId()
    if (!siteId) continue

    if (baseUrls.some((baseUrl) => url.startsWith(baseUrl))) {
      return { baseUrl: baseUrls[0], siteId }
    }
  }
}
```

The only difference I see to the previous implementation is, that if two or more websites do have a matching `baseUrlsFor` the previous algorithm would return `undefined`, where as my proposal would return the first match it can find.

apepper (commented):
FEEDBACK ([26a47aa](https://github.com/Scrivito/scrivito-portal-app/pull/518/commits/26a47aa463144294d1da4e5fc447db0e937f8f4e))

Please see comments.

src/config/scrivitoSites.ts
dcsaszar: > the previous algorithm would return undefined

This was intentional to clearly indicate that after copy/pasting a new website the UI cannot handle the source and destination, and a base URL must be changed.
However, I'll remove the check for now and simplify.

apepper (commented):
FEEDBACK ([d10c130](https://github.com/Scrivito/scrivito-portal-app/pull/518/commits/d10c1303fbdc7c09330c76e392223eefa8d8ce04))

Really nice 🚀!

Now only merge conflicts left.

apepper (approved):
LGTM ([6fac6e9](https://github.com/Scrivito/scrivito-portal-app/pull/518/commits/6fac6e9b4edbff2d23cdec4d1c17e613571e1f3d))

Nice 🚀!

PR #525, opened 2024-07-26
Modern filter specs

apepper:


src/utils/provideLocalStorageDataClass.ts
```diff
@@ -195,3 +192,47 @@ function orderItems(
...
+      throw new Error(
+        `Invalid comparison: Item value ${itemValue} (type: ${typeof itemValue}) and` +
+          ` filter value ${filterValue} (type: ${typeof filterValue}) must be of the same type.`,
+      )
+    }
+    if (
+      typeof itemValue !== 'number' ||
+      typeof filterValue !== 'number' ||
+      typeof itemValue !== 'string' ||
+      typeof filterValue !== 'string'
```
dcsaszar, line 216: * [ ] Always `true`.

src/utils/provideLocalStorageDataClass.ts
```diff
@@ -195,3 +192,47 @@ function orderItems(
...
+}: {
+  itemValue: unknown
+  filterValue: string | number | boolean | null
+  opCode: 'eq' | 'neq' | 'gt' | 'lt' | 'gte' | 'lte'
+}): boolean {
+  if (['gt', 'lt', 'gte', 'lte'].includes(opCode)) {
+    if (typeof itemValue !== typeof filterValue) {
+      throw new Error(
+        `Invalid comparison: Item value ${itemValue} (type: ${typeof itemValue}) and` +
+          ` filter value ${filterValue} (type: ${typeof filterValue}) must be of the same type.`,
```
dcsaszar, line 209: * [ ] FYI Not necessarily. String and number could be compared.

src/utils/provideLocalStorageDataClass.ts
```diff
@@ -195,3 +192,47 @@ function orderItems(
...
+        `Invalid comparison: ${itemValue} and ${filterValue} must be numbers or strings.`,
+      )
+    }
+
+    if (opCode === 'gt') return itemValue > filterValue
+    if (opCode === 'lt') return itemValue < filterValue
+    if (opCode === 'gte') return itemValue >= filterValue
+    if (opCode === 'lte') return itemValue <= filterValue
+
+    throw new Error(`Invalid comparison operator: ${opCode}`)
```
dcsaszar, line 228: * [ ] FYI Unreachable code according to the outer if.

src/utils/provideLocalStorageDataClass.ts
```diff
@@ -195,3 +192,47 @@ function orderItems(
...
+    ) {
+      throw new Error(
+        `Invalid comparison: ${itemValue} and ${filterValue} must be numbers or strings.`,
+      )
+    }
+
+    if (opCode === 'gt') return itemValue > filterValue
+    if (opCode === 'lt') return itemValue < filterValue
+    if (opCode === 'gte') return itemValue >= filterValue
+    if (opCode === 'lte') return itemValue <= filterValue
```
dcsaszar, line 226: * [ ] FYI alternatively we could extract to something like
```ts
return comparators[opCode](itemValue, filterValue)

const comparators = {
  gt: (a: string | number, b: string | number) => a > b,
  lt: (a: string | number, b: string | number) => a < b,
  gte: (a: string | number, b: string | number) => a >= b,
  lte: (a: string | number, b: string | number) => a <= b,
}
```

dcsaszar (commented):
Some comments from my side.

src/Data/toClientParams.ts
```diff
@@ -30,13 +47,55 @@ export function toClientParams(params: IndexParams): FetchParams {
...
   const params: Record<string, string> = {}
 
-  Object.entries(filters).forEach(([name, { opCode, value }]) => {
-    const key = opCode === 'eq' ? name : [name, opCode].join('.')
-    params[key] = value
+  Object.entries(filters).forEach(([name, filter]) => {
+    let filterCollection: FilterSpec[]
+
+    if (filter.operator === 'and') {
+      assertNoConflicts(filter.value)
```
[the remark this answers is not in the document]
apepper: This is inspired how the SDK does this.

src/Data/toClientParams.ts
```diff
@@ -30,13 +47,55 @@ export function toClientParams(params: IndexParams): FetchParams {
...
+        .slice(index + 1)
+        .some(
+          (innerSpec) =>
+            innerSpec.operator === outerSpec.operator &&
+            innerSpec.value !== outerSpec.value,
+        ),
+    )
+  ) {
+    throw new ArgumentError(
+      `Multiple filters on the same attribute with the same operator but different values are currently not supported: ${JSON.stringify(
```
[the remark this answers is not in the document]
apepper: Again, inspired by the SDK.

src/Data/toClientParams.ts
```diff
@@ -30,13 +47,55 @@ export function toClientParams(params: IndexParams): FetchParams {
...
+    let filterCollection: FilterSpec[]
+
+    if (filter.operator === 'and') {
+      assertNoConflicts(filter.value)
+      filterCollection = filter.value
+    } else {
+      filterCollection = [filter]
+    }
+
+    filterCollection.forEach((currentFilter) => {
```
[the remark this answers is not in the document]
apepper: Again, inspired by the SDK.

src/utils/provideLocalStorageDataClass.ts
apepper: This is more a safeguard, that this also stays that way.

apepper:
REVIEW ([3adbbd4](https://github.com/Scrivito/scrivito-portal-app/pull/525/commits/3adbbd40d4e5f15ae4953ae954eaabd070bd6069))

src/utils/provideLocalStorageDataClass.ts
dcsaszar: This is unclear. If we expect the code to not handle all cases, see https://github.com/Scrivito/scrivito-portal-app/pull/525#discussion_r1692980563 for a solution.
If we expect unknown input from the SDK, we should type the input as `string`.

apepper: FYI: I adjusted the error message.

apepper:
REVIEW ([df53166](https://github.com/Scrivito/scrivito-portal-app/pull/525/commits/df5316654c3a1d130d2b329cbb0555b98b53ddb1))

src/utils/provideLocalStorageDataClass.ts
dcsaszar: I see. So yes, I'd drop the check and use TypeScript to make sure we handle all cases.

dcsaszar (commented):
My actionable feedback is implemented.
The remaining comments need clarification from somebody more familiar with the new API.
Handing over to @markus-codewrangler

src/Data/toClientParams.ts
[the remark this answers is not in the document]
markus-codewrangler: Filters are already [combined](https://github.com/infopark/scrivito_js/blob/73315175d39e788a0dc9e4426586b185fc86ab63/js/src/scrivito_sdk/data_integration/data_class.ts#L305-L342) into `AndOperatorSpecs` at this point, so conflicts can only arise within them (see also [specs](https://github.com/infopark/scrivito_js/blob/73315175d39e788a0dc9e4426586b185fc86ab63/js/spec/scrivito_sdk/data_integration/external_data_class/external_data_scope_transform_spec.ts#L212-L291)) AFAICS ✅

src/Data/toClientParams.ts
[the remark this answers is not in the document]
markus-codewrangler: > why not let the backend fail?

Fine with me 😜 

@apepper AFAIU you are planning to switch to `provideDataClass({ restApi })` soon, then the SDK would take care of this case anyway, right? 😉

src/Data/toClientParams.ts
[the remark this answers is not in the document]
markus-codewrangler: > We're just flattening and operations as if they were top-level.

Is this covered by my [comment](https://github.com/Scrivito/scrivito-portal-app/pull/525#discussion_r1694973663) above? Otherwise please elaborate what you mean by this.

> why don't we receive a flat input already?

Quote from the [concept](https://docs.google.com/document/d/1-BdfNqypQctWiJmaNYHzyHSgboWU7b37nTqrFq__4Qs/edit#heading=h.lchp0odx7lb3):

> But other places model filters as a map from attribute name to operator. In these places, it's not yet possible to have multiple operators on the same attribute.

This applies to the `IndexParams#filters` API. So flattening is not possible here, unless we also change the format to a list (which I cannot derive from the concept)

markus-codewrangler:
> Handing over to @markus-codewrangler

I added my 2 cents. So back to @dcsaszar 😁

dcsaszar (approved):
LGTM. I'll clarify the need for conflict checks in the SDK.

PR #526, opened 2024-07-26
MetaNavigation: Only show cart count if available.

apepper:
One small preparation to smooth the migration to `1.43.0-rc1` once available.

### DIFF
--- a/src/Data/CartItem/Cart.ts
+++ b/src/Data/CartItem/Cart.ts
@@ -43,7 +43,7 @@ export function containsItems(): boolean {
   return CartItem.all().containsData()
 }
 
-export function numberOfCartItems(): number {
+export function numberOfCartItems(): number | null {
   if (!isUserLoggedIn()) return 0 // TODO: remove, once CartItem itself requires a login
 
   // @ts-expect-error until out of private beta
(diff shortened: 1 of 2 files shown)


dcsaszar (approved):
LGTM apart from
* [ ] FYI If `containsItems()` changes from unknown to true we have a layout shift, which is probably unavoidable. If `numberOfCartItems()` changes from null to a number after that, we have another layout shift. This one could be avoided.

apepper:
> FYI If `containsItems()` changes from unknown to true we have a layout shift, which is probably unavoidable. If `numberOfCartItems()` changes from null to a number after that, we have another layout shift. This one could be avoided.

I assume you have something like this in mind:

```ts
const itemsCount = !!siteCartPage && numberOfCartItems()
```

With that the `numberOfCartItems()` items would be executed regardless of `containsItems()` having any items. So if I have to choose one more layout shift vs. one more network request I would pick the layout shift.

dcsaszar:
Actually what I had in mind is something like:
```tsx
<span className="badge rounded-pill text-bg-secondary">{numberOfCartItems() || ""}</span>
```
```css
.badge.rounded-pill:empty {
  display: inline-block !important;
  min-width: 16px;
  opacity: 0;
}
```
But since the number uses a variable font and can be > 9 layout shifts would still occur.

PR #527, opened 2024-07-26
Handle ambiguous base URLs

dcsaszar:
Todo from #518

Also added support for xx-XX language codes for language versions.

I ignored the sort performance assuming the number of sites is smallish and the callback is not called frequently.

### DIFF
--- a/src/config/scrivitoSites.ts
+++ b/src/config/scrivitoSites.ts
@@ -72,6 +72,7 @@ function findSiteForUrl(url: string) {
       const siteId = site.siteId()!
       return baseUrlsFor(site).map((baseUrl) => ({ baseUrl, siteId }))
     })
+    .sort((a, b) => b.baseUrl.length - a.baseUrl.length)
     .find(({ baseUrl }) => url.startsWith(baseUrl))
 }
 
--- a/src/config/scrivitoSites.ts
+++ b/src/config/scrivitoSites.ts
@@ -58,7 +58,9 @@ export function siteForUrl(
 }
 
 function languageForUrl(url: string) {
-  const regex = new RegExp(`^${getBaseAppUrl()}\\/(?<lang>[a-z]{2})([?/]|$)`)
+  const regex = new RegExp(
+    `^${getBaseAppUrl()}\\/(?<lang>[a-z]{2}(-[A-Z]{2})?)([?/]|$)`,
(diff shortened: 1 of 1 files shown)


src/config/scrivitoSites.ts
```diff
@@ -58,7 +58,9 @@ export function siteForUrl(
 }
 
 function languageForUrl(url: string) {
-  const regex = new RegExp(`^${getBaseAppUrl()}\\/(?<lang>[a-z]{2})([?/]|$)`)
+  const regex = new RegExp(
+    `^${getBaseAppUrl()}\\/(?<lang>[a-z]{2}(-[A-Z]{2})?)([?/]|$)`,
```
apepper, line 62: According to [RFC 5646](https://www.ietf.org/rfc/bcp/bcp47.txt) in "2.1.1.  Formatting of Language Tags":

> At all times, language tags and their subtags, including private use and extensions, are to be treated as case insensitive: there exist conventions for the capitalization of some of the subtags, but these MUST NOT be taken to carry meaning.
>
> Thus, the tag "mn-Cyrl-MN" is not distinct from "MN-cYRL-mn" or "mN-cYrL-Mn" (or any other combination), and each of these variations conveys the same meaning: Mongolian written in the Cyrillic script as used in Mongolia.

So `lang` should be treaded `cAsE iNsensitIVE`.

apepper (commented):
FEEDBACK ([e194e1f](https://github.com/Scrivito/scrivito-portal-app/pull/527/commits/e194e1f6cfcd897b6672fbac366fa2a19273958b))

Nice 🚀!

There is one smaller issue (see comments).

src/config/scrivitoSites.ts
dcsaszar: Since we only handle two-letter languages so far, and the UI automatically uppercases the locale part, I went with the simple implementation.
Still update?

apepper: Nice. I didn't know, that the SDK automatically corrects the casing.

apepper (approved):
LGTM ([e194e1f](https://github.com/Scrivito/scrivito-portal-app/pull/527/commits/e194e1f6cfcd897b6672fbac366fa2a19273958b))

Nice 🚀!

PR #529, opened 2024-07-29
Scrivito 1.43.0-rc1

apepper:
There are more improvements/usage of newer APIs and feature coming. But as a start this should be already mergable.

### DIFF
--- a/src/Data/toClientParams.ts
+++ b/src/Data/toClientParams.ts
@@ -1,6 +1,10 @@
-import { provideDataClass, createRestApiClient } from 'scrivito'
+import { provideDataClass, createRestApiClient, ArgumentError } from 'scrivito'
 
-type DataConnection = Parameters<typeof provideDataClass>[1]['connection']
+type DataConnection = Parameters<
+  typeof provideDataClass
+>[1]['connection'] extends Partial<infer U> | Promise<Partial<infer U>>
+  ? Partial<U>
+  : never
 type IndexCallback = NonNullable<DataConnection['index']>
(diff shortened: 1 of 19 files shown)


dcsaszar (approved):

src/Data/toClientParams.ts
```diff
@@ -30,13 +47,55 @@ export function toClientParams(params: IndexParams): FetchParams {
   }
 }
 
-function toClientFilterParam(filters: Record<string, OperatorSpec>) {
+function toClientFilterParam(
```
krishan, line 50: Here, you update the `toClientParams` functions to replicate the new logic that has been built into the SDK.

Question: Since the new SDK version supports dynamic headers and also supports object values, why do we even need this module (`toClientParams`) anymore? 

I would have expected the portal app to switch to using `restApi` everywhere and to remove this code.

Or is there still some feature missing in the SDK?

krishan: It's only the binary processing that still causes the need for this code, right? 
Maybe that can be refactored somehow? 

Because I would really love to get rid of this module, since it creates a liability for every developer forking this repo.

apepper: With https://github.com/Scrivito/scrivito-portal-app/pull/536 there are now only two remaining callers of `toClientFilterParam`/`toClientParams`: `Ticket` and `Message`, since both do need custom binary handling for during update/create.

To my knowledge the user of the SDK can either only use `restApi` or have to implement all `connection` functions. So if we want to support all of the filters etc. for `Ticket` and `Message` we would still need something like `toClientParams`.

PR #530, opened 2024-07-30
Bugfix CallbackRequest: If no callback can be found return empty string

apepper:
With Scrivito 1.43.0-rc1 errors in coonection.index will be thrown again.

### DIFF
--- a/src/Data/CallbackRequest/Pisa/pisaCallbackRequestDataClass.ts
+++ b/src/Data/CallbackRequest/Pisa/pisaCallbackRequestDataClass.ts
@@ -1,4 +1,4 @@
-import { provideDataClass } from 'scrivito'
+import { ClientError, provideDataClass } from 'scrivito'
 import { pisaClient } from '../../pisaClient'
 import { DataIndexResponse, RawItem } from '../../types'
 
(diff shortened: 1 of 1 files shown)


src/Data/CallbackRequest/Pisa/pisaCallbackRequestDataClass.ts
```diff
@@ -8,10 +8,18 @@ export async function pisaCallbackRequestDataClass() {
...
+      index: async () => {
+        try {
+          return {
+            results: [await callbackRequestClient.get('')],
+          } as DataIndexResponse
+        } catch (error) {
+          if (!(error instanceof ClientError)) throw error
+          if (error.code !== 'record.not_found') throw error
+
+          return { results: [] }
```
dcsaszar, line 20: * [ ] FYI We could invert the conditions to make the fact that we're only handling `'record.not_found'` clearer to the reader. E.g. (untested):
```ts
if ((error instanceof ClientError) && error.code === 'record.not_found') return { results: [] }
throw error
```

dcsaszar (approved):

apepper:
REVIEW ([95530b8](https://github.com/Scrivito/scrivito-portal-app/pull/530/commits/95530b8a47554431277cd17473e4fab2f09c1310))

PR #533, opened 2024-07-30
Flex columns basics

dcsaszar:
### Todos

These need design or discussion:

* [x] Decide for UX of flex/grid/responsive grid. I could imagine this to be an enum-like control. But we could also keep flex/grid a toggle and conditionally show (or grey out) the responsive toggle.
* [ ] Implement read-only styling for the flex grow toggle switches.
* [x] Fix flex visualization (right padding) if the add-column button is not shown.
* [x] Small flex visualization fix: the left blue border of the leftmost flex column is not visible.
* [x] Make vertical column stretch in flex mode work (or decide to conditionally hide the option).
* [ ] Try to improve behavior if columns overflow the container. (See [here](https://www.bigbinary.com/blog/understanding-the-automatic-minimum-size-of-flex-items))


agessler:
@dcsaszar Looks good so far. The option for mobile adaption is not there, yet. 

d-flex ->  no mobile adaption linearization
d-md-flex -> mobile adaption linearization

src/Widgets/ColumnContainerWidget/ColumnContainerWidgetComponent.tsx
```diff
@@ -6,18 +6,33 @@ provideComponent(ColumnContainerWidget, ({ widget }) => {
...
-  const classNames = ['row', `align-items-${alignment}`]
+  const classNames = [`align-items-${alignment}`, isFlex ? 'd-flex' : 'row']
 
   return (
     <div className={classNames.join(' ')}>
       {columns.map((columnWidget) => {
+        const key = columnWidget.id()
+
+        if (isFlex) {
+          const className = 'flex-grow-0'
```
apepper: Could you also add `h-100`?

src/Widgets/ColumnContainerWidget/ColumnsEditorTab.tsx
```diff
@@ -204,12 +213,24 @@ const ColumnsEditor = connect(
...
     function adjustGrid(newGrid: number[]) {
       if (readOnly) return
+      adjustCols(newGrid)
+      adjustFlexGrowFromGrid(widget.get('columns'), newGrid)
+    }
+
+    function adjustGrow(newGrow: boolean[]) {
+      if (readOnly) return
+      const newGrid =
+        length === 5 ? [2, 2, 2, 2, 4] : newGrow.map(() => 12 / newGrow.length)
```
apepper, line 214: Where does `length` come from?

src/Widgets/ColumnContainerWidget/ColumnContainerWidgetComponent.tsx
```diff
@@ -5,29 +5,43 @@ provideComponent(ColumnContainerWidget, ({ widget }) => {
...
+          const className = flexGrow ? 'flex-grow-1' : undefined
+          return (
+            <ContentTag
+              key={key}
+              className={className}
+              content={columnWidget}
+              attribute="content"
+            />
+          )
+        }
```
apepper, line 29: I think it would be useful, if there is some spacing between the columns. The search input and the button of the next screenshot are placed in a flex column widget:

Screenshot 2024-07-31 at 11 41 04

apepper (changes requested):
FIRST-LOOK-FEEDBACK ([37ec93a](https://github.com/Scrivito/scrivito-portal-app/pull/533/commits/37ec93a03386c397ac0a44a0f6bbb644f0602df4))

Very nice 🚀! See my comments for the first look. There is probably more to come, but a really good start!

agessler:
@dcsaszar Here my solution..

### **without responsive adaption**
**wrapper class**
`d-flex`
**each column has this class**
`mx-2`

### **with responsive adaption**
**wrapper class**
`d-md-flex`
**each column has these classes**
`my-md-0 m-2`

src/Widgets/ColumnContainerWidget/ColumnContainerWidgetComponent.tsx
```diff
@@ -1,33 +1,76 @@
...
+  isResponsive: boolean
+  isStretch: boolean
+}
+
+const FlexColumn = connect(function FlexColumn({
+  columnWidget,
+  isResponsive,
+  isStretch,
+}: ColumnProps) {
+  const classNames = isResponsive ? ['my-md-0', 'm-2'] : ['mx-2']
```
apepper, line 45: The `m-2` does a margin to every column - also to the edges. With that there is "empty space" on the left and right of the column:

Screenshot 2024-08-06 at 09 40 23

Compare this to the "grid" version, where the left and right align with the other elements:

Screenshot 2024-08-06 at 09 40 35

src/Widgets/ColumnContainerWidget/ColumnContainerWidgetComponent.tsx
```diff
@@ -1,33 +1,76 @@
...
+  if (columnWidget.get('flexGrow')) classNames.push('flex-grow-1')
```
apepper, line 46: I noticed, that one "grow" column of two "grow" columns can take more space, than 50%. E.g. if the right grow column contains an image:

Screenshot 2024-08-06 at 10 08 03

Is this expected? By looking at the editor I would have expected to take "max 50%" of the available space. Is this a wrong expectation?

src/Widgets/ColumnContainerWidget/ColumnContainerWidgetComponent.tsx
```diff
@@ -1,33 +1,76 @@
...
+  if (isStretch && columnWidget.get('content').length < 2) {
+    classNames.push(isResponsive ? 'd-md-flex' : 'd-flex')
```
apepper, line 48: * [ ] This seems to not cover the full height (see middle column).
* [ ] Also this seems to make the left and middle column have less space available - which is unexpected to me:

Screenshot 2024-08-06 at 09 31 34
Screenshot 2024-08-06 at 09 31 30

src/Widgets/ColumnContainerWidget/ColumnsEditorTab.tsx
```diff
@@ -56,14 +63,23 @@ const ColumnsEditor = connect(
           }}
           readOnly={readOnly}
         />
+        <Switch
+          className="two_valued"
+          labels={['Grid', 'Flex']}
+          onClick={() =>
+            widget.update({ layoutMode: isFlex ? 'grid' : 'flex' })
+          }
+          title="Display mode"
```
apepper, line 72: The dialog version uses some "blue" color where mostly only green should be used:

Screenshot 2024-08-06 at 10 26 41


Compare this to the current `main` version:

Screenshot 2024-08-06 at 10 26 24

src/Widgets/ColumnContainerWidget/ColumnsEditorTab.tsx
```diff
@@ -566,3 +652,50 @@ function AlignmentDescription({ alignment }: { alignment: string | null }) {
...
+  className?: string
+  labels: string[]
+  onClick: () => void
+  title: string
+  value: number
+}) {
+  return (
+    <>
+      <div className="scrivito_detail_label">
+        <span style={{ fontSize: '11px' }}>{title}</span>
```
apepper, line 672: Is this `fontSize: '11px'` still needed?

apepper (changes requested):
FEEDBACK ([904ee45](https://github.com/Scrivito/scrivito-portal-app/pull/533/commits/904ee456d96ba3b1d8420afcac07b5b570a46614))

Wow, this is massiv 🚀!

Please see my comments for areas, that still need work.

src/Widgets/ColumnContainerWidget/ColumnContainerWidgetComponent.tsx
dcsaszar: Yes, this is intentional. Alex and I discussed this. Since there are many ways to use grid and flex which the current widget settings don't support (yet?) the idea is to provide two sane defaults as follows:
* The **grid** display is used for traditional multi-column layout, like columns in a newspaper which are aligned left and right with gaps in between.
* The **flex** display is used for horizontal stacks, with gaps between elements and spacing around.

The two will automatically look good for their respective use cases.
If you find a use case that requires flex but no outer spacing, we need to discuss the options.

apepper: I do have this use-case in mind:
Screenshot 2024-08-06 at 11 18 22

I would have expected, that the "Create new ticket" button is right aligned with the card-list below it.

src/Widgets/ColumnContainerWidget/ColumnContainerWidgetComponent.tsx
dcsaszar: > Is this a wrong expectation?
Yes, this is a consequence of how flex-grow works by default. And yes, the visualization can/will be misleading in these cases. Flex display is a tool for advanced editors.

src/Widgets/ColumnContainerWidget/ColumnContainerWidgetComponent.tsx
dcsaszar: > * [ ]  This seems to not cover the full height (see middle column).

* Our card widget has a bottom margin
* For grid display we use a workaround of a `h-100` wrapper, which through some CSS quirk moves the margin of the card to an invisible location, but only on desktop. In the **responsive** view the issue is **also visible for grid**.
* For flex the workaround will not work, b/c flex has no support for `h-100`.

So, actually I'd say the (only) unexpected behavior are desktop grid columns.
Maybe we should rethink card (and other?) widget margins in general and make them work in all cases, grid, flex, mobile, desktop.

> * [ ]  Also this seems to make the left and middle column have less space available - which is unexpected to me:

This is how flex works and how it's supposed to be for now. But I see that an average editor may expect a different
outcome. We should loop this feedback to Alex. Maybe there is some flex magic that leads to a more deterministic layout.

src/Widgets/ColumnContainerWidget/ColumnContainerWidgetComponent.tsx
apepper: > Currently flex columns are optimized/intended for using them in the white cards under the input form.

Can you describe this use-case further? Maybe provide a working copy where this is visible with the new gained functionality.

apepper:
Please resolve the conflicts in the PR.

apepper (approved):
LGTM ([e5efd26](https://github.com/Scrivito/scrivito-portal-app/pull/533/commits/e5efd26f2a26e8efd286e7276a0d98cccbb6def1))

Wow, nice work 🚀!

Kudos to @dcsaszar and @agessler to bring this over the finish line!

PR #548, opened 2024-08-05
Replace useDataItem() with useData().dataItem()

apepper:
useDataItem() might not return the most "top" item on the stack, if there is e.g. a "DataScope" on top of the stack. "useData()" will always return the newest thing on top of the stack, e.g. a "dataItem" or a "dataScope".

Follow up of https://github.com/Scrivito/scrivito-portal-app/pull/547.


dcsaszar (approved):
LGTM

PR #551, opened 2024-08-05
[circle] Use node version 18.18.

apepper:
Otherwise some package will complain, that they are not compatible.

Avoids "Unsupported engine" warnings, such as https://app.circleci.com/pipelines/github/Scrivito/scrivito-portal-app/1709/workflows/67bed7bd-0d09-4639-bfde-8061eacc69be/jobs/5272?invite=true#step-102-0_40

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
       npm-run-command:
         type: string
     docker:
-      - image: cimg/node:18.17
+      - image: cimg/node:18.18
     steps:
       - checkout
       - run: npm ci


dcsaszar (approved):
FYI We could also go for `cimg/node:lts` which is a good compromise for:
* Having the latest stable node features
* Avoiding deprecation warnings from the `current` version
* Faster job startup b/c of high likelihood of cached images

apepper:
If I'm not mistaken `cimg/node:lts` is currently using `node` `18.17`, which would still produce the notice.

dcsaszar:
lts is 20.16.0, see https://circleci.com/developer/images/image/cimg/node#image-tags

apepper:
Confusing. So there is a `lts` tag with `node 18.17.1` and one `lts` alias with `node 20.16.0`:

Screenshot 2024-08-05 at 12 14 11


Would `cimg/node:lts` install the tag or the alias?

dcsaszar:
I just tried it out:
image

PR #554, opened 2024-08-05
Lighthouse CI

dcsaszar:


.circleci/config.yml
```diff
@@ -1,29 +1,49 @@
...
+      - steps: << parameters.setup >>
       - run: npm ci
       - run: npm run << parameters.npm-run-command >>
 
 workflows:
   main:
     jobs:
-      - execute-npm:
-          name: npm run build
-          npm-run-command: build
```
apepper, line 20: Please keep `build` as a dedicated job. Otherwise if "lighthouse" fails it is not clear, if this is because of a broken build or if some lighthouse score was not achieved.

package.json
```diff
@@ -8,6 +8,7 @@
     "check-types": "tsc",
     "preview": "vite preview",
     "lint": "eslint --max-warnings 0 'src/**/*.ts' 'src/**/*.tsx' && knip",
+    "lighthouse": "SCRIVITO_TENANT=d0a154d76edf2a7bd991fc658e700a1d SCRIVITO_ORIGIN=http://localhost:8080 PRERENDER_OBJ_ID=c2a0aab78be05a4e npm run prerender && precompress --include '(**).(css|html|js)' --types br dist-ssr && for run in $(seq 6); do lhci autorun && break; done",
```
apepper, line 11: Do we need to hard-code the `SCRIVITO_TENANT` ID? Why not take the tenant from the env? This way the lighthouse test would be automatically transferred.

apepper, line 11: Regarding `precompress --include '(**).(css|html|js)' --types br dist-ssr` - similar to the `startServerCommand` comment from above: Could this be consolidated with `npm run prerender-preview`?

apepper, line 11: Regarding `for run in $(seq 6); do lhci autorun && break; done`: Do I understand this correctly, that lighthouse will try up to six times to reach the target score? Isn't `numberOfRuns` for that?

apepper (commented):
FEEDBACK ([4d02505](https://github.com/Scrivito/scrivito-portal-app/pull/554/commits/4d02505df93ee58c61751953e287b92b9c2ace00))

Good stuff 🚀!

Please see my comments for questions/suggestions.

.lighthouserc.yml
```diff
@@ -0,0 +1,17 @@
...
+    includePassedAssertions: true
+  collect:
+    numberOfRuns: 1
+    startServerCommand: http-server dist-ssr --brotli -c31536000
+    startServerReadyPattern: Hit CTRL-C to stop the server
+    url:
+      - http://localhost:8080/en
+  upload:
+    outputDir: .lighthouseci/results
+    target: filesystem
```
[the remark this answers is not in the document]
dcsaszar: No, they aren't, but the upload must go somewhere, so I just put them in an ignored-anyway folder. If you are fine with different values feel free to suggest some:
https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md#target

.lighthouserc.yml
```diff
@@ -0,0 +1,17 @@
...
+  assert:
+    assertions:
+      categories:accessibility: [error, minScore: 0.94]
+      categories:best-practices: [error, minScore: 1]
+      categories:performance: [error, minScore: 0.68]
+      categories:seo: [error, minScore: 0.92]
+    includePassedAssertions: true
+  collect:
+    numberOfRuns: 1
+    startServerCommand: http-server dist-ssr --brotli -c31536000
```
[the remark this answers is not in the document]
dcsaszar: Yes there are multiple degrees, depending on these goals:
1. Do we want to stick with a single tool to preview prerendered pages?
  * This would be easy: `http-server dist-ssr`
2. Should the preview be suited for performance measurements?
  * This would require the `precompress` step. We could add it to `prerender-preview`, which would slow this script down a bit. Or we'd have two `prerender-preview` scripts?

package.json
dcsaszar: The current CI env has no `SCRIVITO_TENANT` set. The consequence of moving the setting to the CI is that we force forks with CI integration to set the env too. If this is fine, yes we could do this.

dcsaszar: Yes, but
* `numberOfRuns` will *always* collect N runs first and find the max score afterwards. This means **always** N times the CI duration.
* `numberOfRuns > 1` currently reports wrong numbers, see https://github.com/GoogleChrome/lighthouse-ci/issues/1067

.lighthouserc.yml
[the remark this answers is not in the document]
apepper: Thanks for the investigation.

`filesystem` is ok, but how about moving it to `tmp/lighthouseci/results` and gitignore `tmp`?

package.json
```diff
@@ -8,8 +8,10 @@
     "check-types": "tsc",
     "preview": "vite preview",
     "lint": "eslint --max-warnings 0 'src/**/*.ts' 'src/**/*.tsx' && knip",
+    "lighthouse": "SCRIVITO_ORIGIN=http://localhost:8080 PRERENDER_OBJ_ID=c2a0aab78be05a4e npm run prerender && npm run prerender-precompress && for run in $(seq 6); do lhci autorun && break; done",
     "prerender": "vite build && vite-node --options.deps.external='/scrivito/' src/prerenderContent.ts",
-    "prerender-preview": "vite preview --outDir dist-ssr",
+    "prerender-precompress": "precompress --include '(**).(css|html|js)' --types br dist-ssr",
+    "prerender-preview": "http-server dist-ssr --port 8080 --brotli -c31536000",
```
apepper, line 14: This does not seem to work - at least on my machine.

Steps to reproduce:

```bash
SCRIVITO_ORIGIN=http://localhost:8080 npm run prerender && npm run prerender-preview
```

Visit http://localhost:8080 in a private browser window.

Expected: A redirect to /en is happening and the actual portal app is rendered.

Actual: The "index" page of the folder "en/" is visible:
Screenshot 2024-08-07 at 14 43 15

.lighthouserc.yml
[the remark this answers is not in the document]
apepper: I like the current setup with `npm run prerender-precompress` and `npm run prerender-preview`.

apepper (commented):
FEEDBACK ([f7091b9](https://github.com/Scrivito/scrivito-portal-app/pull/554/commits/f7091b9ab227fadebf398ae308c646186a61199f))

Nice 🚀! Please see my comments.

.lighthouserc.yml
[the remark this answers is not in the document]
dcsaszar: `.lighthouseci` is written to anyway (see https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md#collect).

What we could do is something like:
```yaml
  upload:
    outputDir: .lighthouseci
    reportFilenamePattern: tmp
    target: filesystem
```
This would no longer accumulate reports for local runs and even hide the (over)written dummy file from most IDEs.
But in case someone is actually interested in archived local results, this would be a blocker. Leave as is?

package.json
```diff
@@ -8,8 +8,10 @@
     "check-types": "tsc",
     "preview": "vite preview",
     "lint": "eslint --max-warnings 0 'src/**/*.ts' 'src/**/*.tsx' && knip",
+    "lighthouse": "SCRIVITO_ORIGIN=http://localhost:8080 PRERENDER_OBJ_ID=c2a0aab78be05a4e npm run prerender && npm run prerender-precompress && for run in $(seq 6); do lhci autorun && break; done",
     "prerender": "vite build && vite-node --options.deps.external='/scrivito/' src/prerenderContent.ts",
-    "prerender-preview": "vite preview --outDir dist-ssr",
+    "prerender-precompress": "precompress --include '(**).(css|html|js)' --types br dist-ssr",
+    "prerender-preview": "ws --directory dist-ssr --port 8080 --rewrite '(/..)(/.*|$) -> $&.html' --static.maxage 31536000",
```
apepper, line 14: We are getting closer. Now only a fallback file is missing:

*How to reproduce*
```
SCRIVITO_ORIGIN=http://localhost:8080 npm run prerender && npm run prerender-preview
```
Visit http://localhost:8080/en/my-tynacoon

*Expected*
A nicely rendered "Page not found"
Screenshot 2024-08-08 at 09 30 01

*Actual*
A white page stating "Not found"
Screenshot 2024-08-08 at 09 30 31

apepper (commented):
FEEDBACK ([a68fb50](https://github.com/Scrivito/scrivito-portal-app/pull/554/commits/a68fb5017f65f0ef3de2dffa9bcfbb8a9ca336f0))

Almost there 🚀!

apepper (approved):
LGTM ([5b2d689](https://github.com/Scrivito/scrivito-portal-app/pull/554/commits/5b2d689cc02ff71324cd85d0f9a03908fefa8c9a))

Awesome 🚀! Great work!

PR #555, opened 2024-08-05
Rip toClientParams

apepper:
Moved the "convert blob to upload binary" logic from the `connection` to `DataFormContainerWidget`.

With this it is now possible to remove `toClientParams` and `convertBlobAttributes`. Only `blobToBinary` remains.


src/Widgets/DataFormContainerWidget/DataFormContainerWidgetComponent.tsx
```diff
@@ -131,9 +141,12 @@ function valueFromElement(
...
     const files = element.files
-    if (element.multiple) return files ? [...files] : []
-    if (files?.length === 1) return files[0]!
-    return null
+    if (!files) return element.multiple ? [] : null
+
+    const blobToBinaryPromises = [...files].map(blobToBinary)
+    return element.multiple
+      ? Promise.all(blobToBinaryPromises)
+      : blobToBinaryPromises[0]!
```
dcsaszar, line 146: * [ ] The old code did a `length === 1` check to allow for the bang cast. The `files` can be zero-length (https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/files#javascript). We could do `: blobToBinaryPromises[0] || null`.

dcsaszar (changes requested):
Nice. One comment.

apepper:
REVIEW ([a6b3d2c](https://github.com/Scrivito/scrivito-portal-app/pull/555/commits/a6b3d2c507fa9023ab7a8d7d359cf80edc536821))

dcsaszar (approved):

PR #558, opened 2024-08-07
Drop hardcoded site ID "default"

dcsaszar:
Instead of `SCRIVITO_ROOT_CONTENT_ID` we now configure `SCRIVITO_ROOT_OBJ_ID`. This allows to infer both
* the default site ID and
* the root content ID

from the obj, which was not possible the other way around.

This makes it easier to integrate the app into existing instances that already use the site ID `default`.

### DIFF
--- a/src/Data/pisaClient.ts
+++ b/src/Data/pisaClient.ts
@@ -3,7 +3,7 @@ import { isHomepage } from '../Objs/Homepage/HomepageObjClass'
 
 export async function pisaUrl(): Promise<string> {
   const defaultRoot = await load(() =>
-    Obj.onAllSites().get(import.meta.env.SCRIVITO_ROOT_CONTENT_ID),
+    Obj.onAllSites().get(import.meta.env.SCRIVITO_ROOT_OBJ_ID),
   )
   if (!isHomepage(defaultRoot)) return never()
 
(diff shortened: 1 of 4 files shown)


apepper (approved):
LGTM ([8213a05](https://github.com/Scrivito/scrivito-portal-app/pull/558/commits/8213a05698fc7734622cdb4b68cce411b9ba2fee))

Nice 🚀!

Is it maybe time to even change the siteId `default` in the example content to a random hex value?

dcsaszar:
> Is it maybe time to even change the siteId `default` in the example content to a random hex value?

https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://scrivito-portal-app.pages.dev/en?_scrivito_workspace_id=veaf24726e018d30&_scrivito_display_mode=diff

ID genarated by https://www.browserling.com/tools/random-hex

markus-codewrangler:
> Is it maybe time to even change the siteId default in the example content to a random hex value?

FYI I just stumbled on this and realized that `default` is still used in several places in the SDK/UI code.

Is it maybe time to also replace it there? 🤔

dcsaszar:
> FYI I just stumbled on this and realized that `default` is still used in several places in the SDK/UI code.
> **Edit**: or rephrased: is it a good idea to change the default `siteId` in the portal-app, if the SDK/UI still uses it in some places?

The Portal App is multi-site. The site ID `default` has some legacy/backwards-compatibility features which only make sense in a single-site environment. That's why the SDK knows about `default` and that's also why it's cleaner to not have a `default` site at all here. From https://docs.scrivito.com/js-sdk/configure:

> homepage (Function) – a callback that returns a CMS Obj representing the page at the root URI / (default: Scrivito.Obj.root). This callback is only executed for the site whose siteId is “default” which is the only site in a single-site setup. It is not recommended to use this callback in a multi-site setup.

PR #565, opened 2024-08-09
Simplify base URL logic

dcsaszar:
* we configure `baseUrl: 'stringlist'`, thus
* the SDK will never return a `link` or `linklist` (even if this is stored as such on another site)
* the new simplified logic still supports `string` thanks to `autoConvertAttributes: true`

Line 72 could alternatively read `if (!Array.isArray(baseUrl)) return []`. I kept the current version to make switching to `string` easier. Maybe YAGNI but we need a check for type safety anyway.

### DIFF
--- a/src/config/scrivitoSites.ts
+++ b/src/config/scrivitoSites.ts
@@ -1,12 +1,4 @@
-import {
-  Link,
-  Obj,
-  currentSiteId,
-  getInstanceId,
-  load,
-  navigateTo,
-  urlFor,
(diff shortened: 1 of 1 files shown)


apepper (commented):
LGTM ([97b65a0](https://github.com/Scrivito/scrivito-portal-app/pull/565/commits/97b65a083606d035ec66c199f65a9b4005386fa3))

Nice 🚀!

PR #570, opened 2024-08-13
Top nav selector fixes

dcsaszar:
The previous position-based selectors caused two issues:

* The logo did not overlap if the nav header had a section below the top nav
* All sections in the top nav had a grey bar

Apart from not matching sections that are not actual top-nav sections, the selector matches exactly as before, including the landing page layout and slim nav. Not fully sure if this is intended, but it worked before and still works.


agessler:
@dcsaszar  Fine.. please add my commit here for all fixes..
https://github.com/Scrivito/scrivito-portal-app/pull/571

agessler (approved):

PR #574, opened 2024-08-23
Add `treatLocalhostLike`

markus-codewrangler:
Make sure the Portal-App only gets an app token (`stas_*`) from IAM – not a product token (`st_*`) (caused by the `X-Treat-Localhost-Like` header of the UI's webpack proxy).

This fixes a deviation from the production env, for a local portal-app running in a local UI.

Outcome of https://infopark.slack.com/archives/G011EJEGKD5/p1721287390884569


apepper:
Should be fixed on IAM side.

PR #581, opened 2024-09-03
Readme

apepper:


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
dcsaszar, line 1: * [ ] FYI The Example App also mentions the license in the package.json (https://github.com/Scrivito/scrivito_example_app_js/blob/master/package.json#L4)

readme.mdown
```diff
@@ -0,0 +1,62 @@
...
+
+[Scrivito](https://www.scrivito.com/?utm_source=github&utm_medium=natural&utm_campaign=github_portal_app) is the world's most secure, serverless JavaScript CMS. Based on ReactJS, it is completely maintenance-free, cost-effective, and has unprecedented security. Learn more about Scrivito at [scrivito.com](https://www.scrivito.com/?utm_source=github&utm_medium=natural&utm_campaign=github_portal_app).
+
+This portal app shows Scrivito’s features and is a great basis for your own Scrivito website projects.
```
dcsaszar, line 7: * [ ] FYI In our internal communication we always treat "Portal App" as a name, with uppercase initials. How about: "This portal app" -> "The Portal App" or -> "The Scrivito Portal App"?

readme.mdown
```diff
@@ -0,0 +1,62 @@
...
+Make sure that you have [![node ≥ 20.0](https://img.shields.io/badge/node-≥%2020.0-blue.svg)](https://nodejs.org) installed.
+
+```
+node -v
+# should output something like v20.17.0
+```
+
+### Install dependencies
+
+Now call `npm install` in the command line to install all JS dependencies (which are placed in the folder `node_modules/`).
```
dcsaszar, line 38: * [ ] FYI since this app is geared towards experienced (TypeScript) developers the parenthesized part could be left out. 
* [ ] FYI "all JS dependencies". Actually it's also TS dependencies. Leave out?

readme.mdown
```diff
@@ -0,0 +1,62 @@
...
+
+### Specify the Scrivito CMS to use
+
+The Scrivito portal app requires the tenant ID of your CMS. The ID is taken from the environment variable `SCRIVITO_TENANT`. The environment variable can also be set via [`.env`](https://github.com/motdotla/dotenv).
+
+See `.env.example` for an example.
+
+## How to develop
+
+Calling `npm start` will start a webserver, which listens to [localhost:8080](http://localhost:8080/). Open this URL in your browser. It should also automatically reload the page after changes to the code (in `src/`) have been made.
```
dcsaszar, line 48: * [ ] FYI "to the code (in `src/`) have been made". Vite also reloads after changes to the config outside of src.

readme.mdown
```diff
@@ -0,0 +1,62 @@
...
+# should output something like v20.17.0
+```
+
+### Install dependencies
+
+Now call `npm install` in the command line to install all JS dependencies (which are placed in the folder `node_modules/`).
+
+### Specify the Scrivito CMS to use
+
+The Scrivito portal app requires the tenant ID of your CMS. The ID is taken from the environment variable `SCRIVITO_TENANT`. The environment variable can also be set via [`.env`](https://github.com/motdotla/dotenv).
```
dcsaszar, line 42: * [ ] FYI maybe use the more modern/neutral "instance ID" instead of "tenant ID"

dcsaszar (approved):
Nice. Left some FYI.

readme.mdown
```diff
@@ -0,0 +1,62 @@
+# Scrivito Portal App
+
+[![CMS: Scrivito](https://img.shields.io/badge/CMS-Scrivito-brightgreen.svg)](https://scrivito.com) [![License: MIT](https://img.shields.io/badge/License-MIT-blue.svg)](https://opensource.org/licenses/MIT) [![CircleCI](https://dl.circleci.com/status-badge/img/gh/Scrivito/scrivito-portal-app/tree/main.svg?style=svg)](https://dl.circleci.com/status-badge/redirect/gh/Scrivito/scrivito-portal-app/tree/main) [![GitHub last commit](https://img.shields.io/github/last-commit/Scrivito/scrivito-portal-app.svg)](https://github.com/Scrivito/scrivito-portal-app) [![GitHub top language](https://img.shields.io/github/languages/top/Scrivito/scrivito-portal-app.svg)](https://github.com/Scrivito/scrivito-portal-app)
```
dcsaszar, line 3: * [ ] FYI there's one badge that shows "HTML 60%". Not very useful.

apepper: That is the case, because of the `design/` folder. Maybe its time to move this somewhere else. Ideas?

readme.mdown
apepper: I would keep it as is.

apepper:
REVIEW ([08a9a70](https://github.com/Scrivito/scrivito-portal-app/pull/581/commits/08a9a70baf3ddc55eb339a9e28c0d609bf339130))

readme.mdown
dcsaszar: If we no longer need the prototypes and design scripts on a daily basis, maybe a `design` branch than can be rebased to `main` as needed will do?

dcsaszar (approved):
🏄

readme.mdown
apepper: Nice! See https://github.com/Scrivito/scrivito-portal-app/pull/585.

apepper: With #585 being merged now it looks like this:
Screenshot 2024-09-04 at 12 52 20

Also the languages look like this:
Screenshot 2024-09-04 at 12 51 04

PR #582, opened 2024-09-03
DataFormUploadWidget: Drag 'n' drop, preview and file size limit

apepper:
Allow to drag 'n' drop files on the screen.

Also show a preview of the selected files:

Screenshot 2024-09-03 at 13 58 38

Last, but not least, limit the files to max. 50 MB:

Screenshot 2024-09-03 at 13 59 06

### DIFF
--- a/src/Components/Attachment.tsx
+++ b/src/Components/Attachment.tsx
@@ -5,8 +5,10 @@ import { FullDataBinary, dataBinaryToUrl } from '../utils/dataBinaryToUrl'
 
 export const Attachment = connect(function Attachment({
   attachment,
+  readonly,
 }: {
   attachment: FullDataBinary
+  readonly?: boolean
 }) {
   const [binaryUrl, setBinaryUrl] = useState<string | undefined>(undefined)
   const [trigger, setTrigger] = useState<number>(0)
@@ -23,12 +25,8 @@ export const Attachment = connect(function Attachment({
     }
   }, [attachment, trigger])
 
-  return (
-    <a
-      href={binaryUrl}
-      className="box-attachment"
-      title={`Download ${attachment.filename}`}
-    >
+  const content = (
+    <>
       <div className="box-preview">
         <BoxPreviewContent binaryUrl={binaryUrl} attachment={attachment} />
       </div>
@@ -40,6 +38,18 @@ export const Attachment = connect(function Attachment({
           })}
         </span>
       </div>
+    </>
+  )
+
+  return readonly ? (
+    <div className="box-attachment">{content}</div>
+  ) : (
+    <a
+      href={binaryUrl}
+      className="box-attachment"
+      title={`Download ${attachment.filename}`}
+    >
+      {content}
     </a>
   )
 })
--- a/src/Components/Attachment.tsx
+++ b/src/Components/Attachment.tsx
@@ -1,0 +1,70 @@
+import prettyBytes from 'pretty-bytes'
+import { useState, useEffect } from 'react'
+import { connect, currentLanguage } from 'scrivito'
+import { FullDataBinary, dataBinaryToUrl } from '../utils/dataBinaryToUrl'
+
+export const Attachment = connect(function Attachment({
+  attachment,
+}: {
+  attachment: FullDataBinary
+}) {
+  const [binaryUrl, setBinaryUrl] = useState<string | undefined>(undefined)
+  const [trigger, setTrigger] = useState<number>(0)
+
+  useEffect(() => {
+    let timeoutId: NodeJS.Timeout
+    dataBinaryToUrl(attachment).then(({ url, maxAge }) => {
+      setBinaryUrl(url)
+      timeoutId = setTimeout(() => setTrigger(Date.now()), maxAge * 1000)
+    })
+
+    return () => {
+      if (timeoutId) clearTimeout(timeoutId)
+    }
+  }, [attachment, trigger])
+
+  return (
+    <a
+      href={binaryUrl}
+      className="box-attachment"
+      title={`Download ${attachment.filename}`}
+    >
+      <div className="box-preview">
+        <BoxPreviewContent binaryUrl={binaryUrl} attachment={attachment} />
+      </div>
+      <div className="box-meta">
+        <span className="box-name">{attachment.filename}</span>
+        <span className="box-size">
+          {prettyBytes(attachment.contentLength, {
+            locale: currentLanguage() ?? 'en',
+          })}
+        </span>
+      </div>
+    </a>
+  )
+})
+
+function BoxPreviewContent({
+  binaryUrl,
+  attachment,
+}: {
+  binaryUrl?: string
+  attachment: FullDataBinary
+}) {
+  if (binaryUrl && attachment.contentType.startsWith('image/')) {
+    return <img src={binaryUrl} alt="" />
+  }
+
+  let iconName = 'bi-file-earmark'
+  const filename = attachment.filename
+  if (filename.endsWith('.pdf')) iconName = 'bi-filetype-pdf'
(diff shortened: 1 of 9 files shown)


src/Widgets/DataFormUploadWidget/DataFormUploadWidgetComponent.tsx
```diff
@@ -1,12 +1,38 @@
...
+  const { acceptedFiles, getRootProps, getInputProps, inputRef } = useDropzone()
+
+  useEffect(() => {
+    if (!inputRef.current) return
+
+    const dataTransfer = new DataTransfer()
+    acceptedFiles.forEach((file) => dataTransfer.items.add(file))
+
+    inputRef.current.files = dataTransfer.files
+  }, [acceptedFiles, inputRef])
```
dcsaszar, line 38: * [ ] The `inputRef` dependency is misleading. This is the 2nd case described [here](https://www.epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array#custom-hooks-and-the-ref-object-itself). I didn't read the full article and can't comment on a working solution but the current code could lead to confusion if devs start copy/pasting it. If there is no clean solution maybe a comment would help?

src/Widgets/DataFormUploadWidget/DataFormUploadWidgetComponent.tsx
```diff
@@ -50,14 +76,33 @@ provideComponent(DataFormUploadWidget, ({ widget }) => {
...
+        </div>
+      </div>
     </div>
   )
 })
+
+function getDropMessage(multiple: boolean) {
+  switch (currentLanguage()) {
+    case 'de':
+      return `Wählen oder legen Sie ${multiple ? 'Dateien' : 'eine Datei'} hier ab.`
```
dcsaszar, line 126: * [ ] Haha: "Wählen [...] Sie eine Datei [...] ab.". How about:

Multiple Files:

	•	DE: Dateien auswählen oder hierher ziehen.
	•	EN: Choose files or drag them here.

Single File:

	•	DE: Datei auswählen oder hierher ziehen.
	•	EN: Choose a file or drag it here.

src/Widgets/DataFormUploadWidget/DataFormUploadWidgetComponent.tsx
```diff
@@ -106,3 +123,18 @@ function getDropMessage(multiple: boolean) {
       return `Choose or drop ${multiple ? 'files' : 'a file'} here.`
   }
 }
+
+function getTooLargeMessage() {
+  switch (currentLanguage()) {
+    case 'de':
+      return `Eine oder mehrere Dateien sind zu groß. Bitte Dateien bis maximal ${prettyBytes(
```
dcsaszar, line 135: * [ ] FYI "Bitte laden Sie..." sounds more friendly.
* [ ] UX: It would be nice to mention the limit *before* users start adding/dropping files.

src/Widgets/DataFormUploadWidget/DataFormUploadWidgetComponent.tsx
```diff
@@ -87,7 +88,11 @@ provideComponent(DataFormUploadWidget, ({ widget }) => {
         </>
       ) : null}
 
-      <div {...getRootProps({ className: 'dropzone' })}>
+      <div
+        {...getRootProps({
+          className: `dropzone ${isDragActive ? 'border border-3 rounded-3' : ''}`,
```
dcsaszar, line 93: * [ ] The larger border causes layout shifts on hovering. In the worst case this may make dropping impossible. How about reusing our style for `.form-control:focus`?

dcsaszar (changes requested):
Nice stuff. See comments for mostly UX stuff.

src/Widgets/DataFormUploadWidget/DataFormUploadWidgetComponent.tsx
apepper: I would argue, that this is more the 3rd case described in the same section, since the `inputRef` comes from `useDropzone`. At least eslint is complaining:

Screenshot 2024-09-04 at 11 08 01

To quote from the same article:

> The reason we get the warning there when we didn't get it within the component is because ESLint is pretty limited in its ability to trace what you're doing with your JavaScript. So the React plugin for ESLint can't know that cbRef is actually a ref. So just to be safe, it warns you.
> 
> Luckily, as we just learned, including it in the dependency array doesn't make any difference anyway, so just include it!

I would just leave it as is.

src/Widgets/DataFormUploadWidget/DataFormUploadWidgetComponent.tsx
apepper: > UX: It would be nice to mention the limit before users start adding/dropping files.

I want to keep the text brief for uploading and only show the information when needed. It will also be shown right away without an upload delay.

apepper:
REVIEW ([7b5978a](https://github.com/Scrivito/scrivito-portal-app/pull/582/commits/7b5978a5d71086b15ff8731df16c87686a17c9eb))

src/Widgets/DataFormUploadWidget/DataFormUploadWidgetComponent.tsx
```diff
@@ -90,7 +89,8 @@ provideComponent(DataFormUploadWidget, ({ widget }) => {
 
       <div
         {...getRootProps({
-          className: `dropzone ${isDragActive ? 'border border-3 rounded-3' : ''}`,
+          className: 'dropzone form-control',
+          style: { cursor: 'pointer' },
```
dcsaszar, line 93: * [ ] Q: Code style: Why not use CSS?

dcsaszar (approved):
* [x] FYI The attachments have a pointer cursor and hover style, which makes them look as if they're clickable.
* [ ] FYI Currently we have no drop-hover style on the file input.

apepper:
REVIEW ([6bdb4be](https://github.com/Scrivito/scrivito-portal-app/pull/582/commits/6bdb4beb313e0a838f37f09960eca9c34b76e1e5))

src/Components/Attachment.tsx
```diff
@@ -40,6 +38,18 @@ export const Attachment = connect(function Attachment({
...
+    </>
+  )
+
+  return readonly ? (
+    <div className="box-attachment">{content}</div>
+  ) : (
+    <a
+      href={binaryUrl}
+      className="box-attachment"
+      title={`Download ${attachment.filename}`}
```
dcsaszar, line 50: * [ ] FYI not localized

src/assets/stylesheets/index.scss
```diff
@@ -1188,6 +1188,12 @@ section:focus >,
   transition:
     border-color 0.15s ease-in-out,
     box-shadow 0.15s ease-in-out;
+  &.dropzone {
+    cursor: pointer;
+    &.drag-active {
+      border: 0.15rem dashed;
+    }
```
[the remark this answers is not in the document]
dcsaszar: * [ ] Instead of new custom styling we could reuse `.form-control:focus`. I.e.:

```css
.form-control.drag-active,
.form-control:focus {
  color: #1e2022;
  background-color: #fff;
  border-color: rgba(140, 152, 164, 0.25);
  outline: 0;
  box-shadow: 0 0 1rem 0 rgba(140, 152, 164, 0.25);
}
```

dcsaszar (changes requested):
See comments.

apepper:
REVIEW ([b615ebd](https://github.com/Scrivito/scrivito-portal-app/pull/582/commits/b615ebd6254d734563910520cba10a23b1ab76ea))

dcsaszar (approved):

PR #585, opened 2024-09-04
Remove "design" folder

apepper:
This folder contained html wireframes and prototypes. It will be moved to an [independent "design" branch](https://github.com/Scrivito/scrivito-portal-app/tree/design).

Inspired by https://github.com/Scrivito/scrivito-portal-app/pull/581#discussion_r1741611550.

FYI @agessler .


dcsaszar (changes requested):
Looks good but
* [x] We should also remove `npm run start-design`

apepper:
Forced pushed a new version ([e6c46c4](https://github.com/Scrivito/scrivito-portal-app/pull/585/commits/e6c46c47134d058e9c7b400dcf52ab0c8c648282)) which removed `npm run start-design` as well as some "ignore design folder" configs.

dcsaszar (approved):
Top!

PR #586, opened 2024-09-04
Protected root

apepper:
This PR tries to solve two issue:

* If the tenant in question has no scrivito portal content site show an error
* If the root for the `NotFoundErrorPage` can't be found due to not being logged in - log the user in. This allows for root pages to be protected content.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,4 @@
-import {
-  CurrentPage,
-  NotFoundErrorPage as ScrivitoNotFoundErrorPage,
-} from 'scrivito'
+import { CurrentPage } from 'scrivito'
 import { HelmetProvider, HelmetServerState } from 'react-helmet-async'
 
 import { CurrentPageMetadata } from './Components/CurrentPageMetadata'
@@ -26,9 +23,7 @@ export function App({
           </a>
           <DesignAdjustments>
             <CurrentPage />
-            <ScrivitoNotFoundErrorPage>
-              <NotFoundErrorPage />
-            </ScrivitoNotFoundErrorPage>
+            <NotFoundErrorPage />
           </DesignAdjustments>
           <CurrentPageMetadata />
           <Toasts />
--- a/src/Components/NotFoundErrorPage.tsx
+++ b/src/Components/NotFoundErrorPage.tsx
@@ -2,22 +2,30 @@ import { useEffect } from 'react'
 import {
   connect,
   ContentTag,
+  isUserLoggedIn,
   Obj,
   NotFoundErrorPage as ScrivitoNotFoundErrorPage,
 } from 'scrivito'
+import { Loading } from './Loading'
+import { isNoSitePresent } from '../config/scrivitoSites'
 
 // Make sure, that you have a proxy running for these URLs, otherwise you'll see an endless loop.
 const RELOAD_SUBPATHS = ['/auth']
 
-export const NotFoundErrorPage = connect(function NotFoundErrorPage() {
-  return (
-    <>
-      <ScrivitoNotFoundErrorPage>
-        <NotFound />
-      </ScrivitoNotFoundErrorPage>
-    </>
-  )
-})
+export const NotFoundErrorPage = connect(
+  function NotFoundErrorPage() {
+    if (isUserLoggedIn() && isNoSitePresent()) return <NotFound />
+
+    return (
+      <>
+        <ScrivitoNotFoundErrorPage>
+          <NotFound />
+        </ScrivitoNotFoundErrorPage>
+      </>
+    )
+  },
+  { loading: Loading },
+)
 
 const NotFound = connect(function NotFound() {
   const root = Obj.root()
--- a/src/Components/NotFoundErrorPage.tsx
+++ b/src/Components/NotFoundErrorPage.tsx
@@ -17,11 +17,9 @@ export const NotFoundErrorPage = connect(
     if (isUserLoggedIn() && isNoSitePresent()) return <NotFound />
 
     return (
-      <>
-        <ScrivitoNotFoundErrorPage>
-          <NotFound />
-        </ScrivitoNotFoundErrorPage>
-      </>
+      <ScrivitoNotFoundErrorPage>
+        <NotFound />
+      </ScrivitoNotFoundErrorPage>
     )
   },
   { loading: Loading },
--- a/src/Components/NotFoundErrorPage.tsx
+++ b/src/Components/NotFoundErrorPage.tsx
@@ -14,7 +14,14 @@ const RELOAD_SUBPATHS = ['/auth']
 
 export const NotFoundErrorPage = connect(
   function NotFoundErrorPage() {
-    if (isUserLoggedIn() && isNoSitePresent()) return <NotFound />
+    // TODO: remove window check after #11101 is implemented
+    if (
+      typeof window !== 'undefined' &&
+      isUserLoggedIn() &&
+      isNoSitePresent()
+    ) {
+      return <NotFound />
+    }
 
     return (
       <ScrivitoNotFoundErrorPage>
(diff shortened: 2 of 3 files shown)


src/Components/NotFoundErrorPage.tsx
```diff
@@ -1,12 +1,25 @@
...
 const RELOAD_SUBPATHS = ['/auth']
 
 export const NotFoundErrorPage = connect(function NotFoundErrorPage() {
   const root = Obj.root()
 
+  useEffect(() => {
+    if (root) return
+    if (isUserLoggedIn()) return
+
+    ensureUserIsLoggedIn()
```
dcsaszar, line 20: * [ ] This logic may try to log users in on a 404 even if the homepage exists. `Obj.root()` is `null` while it's loaded from the backend.

src/App.tsx
```diff
@@ -28,6 +29,7 @@ export function App({
           <ScrivitoNotFoundErrorPage>
             <NotFoundErrorPage />
           </ScrivitoNotFoundErrorPage>
+          <SiteNotFoundPage />
```
dcsaszar, line 32: * [ ] FYI It feels weird to have two components responsible for `*NotFound`. From a readers point of view the consequences of this can only be understood by reading both components. How about having a single component responsible for any 404 handling?

src/Components/SiteNotFoundPage.tsx
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react'
+import { connect, load } from 'scrivito'
+import { rootContentId } from '../config/scrivitoSites'
+
+// This page is rendered, if there are no portal sites available.
+export const SiteNotFoundPage = connect(function SiteNotFoundPage() {
+  const [showSiteNotFound, setShowSiteNotFound] = useState(false)
```
dcsaszar, line 7: * [ ] Using state to freeze Scrivito content takes away the ability to rerender in a reactive way. Not sure if and when "Site not found" is shown in editing, but there is a chance that, e.g., fixing an accidental edit will not be reflected in the preview.

dcsaszar (changes requested):
From the PR description I don't see which problem we're solving with this.
Whatever it is, the approach may have some flaws. See comments.

apepper:
REVIEW ([6c77b97](https://github.com/Scrivito/scrivito-portal-app/pull/586/commits/6c77b97598641c7e76799e19c3e40296fe260f06))

I reimplemented the features. I hope they are now better.

src/Components/NotFoundErrorPage.tsx
```diff
@@ -43,3 +60,27 @@ export const NotFoundErrorPage = connect(function NotFoundErrorPage() {
...
+    return <div>Page not found.</div>
+  },
+  { loading: () => null },
+)
+
+const PortalSiteNotFound = connect(
+  function PortalSiteNotFound() {
+    if (portalSiteAvailable()) return null
+
+    return <div>Site not found.</div>
```
dcsaszar, line 83: * [ ] FYI I don't see how a user would profit from "Site not found" vs. "Page not found". The distinction may be more confusing than helpful. How about showing `<NotFound />` here, too? 
* [ ] This implementation is also missing the auto-login feature for login-only websites. `<NotFound />` would solve it.

src/config/scrivitoSites.ts
```diff
@@ -110,6 +110,10 @@ function rootContentId() {
     ?.contentId()
 }
 
+export function portalSiteAvailable(): boolean {
+  return !!rootContentId()
+}
+
```
dcsaszar, line 116: * [ ] Instead of introducing a custom helper we could use `Scrivito.currentSiteId() !== null` for "site available".

src/Components/NotFoundErrorPage.tsx
```diff
@@ -1,12 +1,30 @@
...
 const RELOAD_SUBPATHS = ['/auth']
 
 export const NotFoundErrorPage = connect(function NotFoundErrorPage() {
-  const root = Obj.root()
+  return (
+    <>
+      <ScrivitoNotFoundErrorPage>
+        <NotFound />
+      </ScrivitoNotFoundErrorPage>
+      <PortalSiteNotFound />
```
dcsaszar, line 22: * [ ] FYI We could make the fact we're only rendering one of these two clearer by conditional rendering, e.g.:
```tsx
  return currentSiteId() === null ? (
    <PortalSiteNotFound />
  ) : (
    <ScrivitoNotFoundErrorPage>
      <NotFound />
    </ScrivitoNotFoundErrorPage>
  )
```

src/Components/NotFoundErrorPage.tsx
```diff
@@ -43,3 +60,27 @@ export const NotFoundErrorPage = connect(function NotFoundErrorPage() {
     </>
   )
 })
+
+const NoRootError = connect(
```
dcsaszar, line 64: * [ ] FYI to be more inline with the naming of the other components in this module: `NoRoot`.

src/Components/NotFoundErrorPage.tsx
```diff
@@ -43,3 +60,27 @@ export const NotFoundErrorPage = connect(function NotFoundErrorPage() {
...
+        if (root === null && !isUserLoggedIn()) ensureUserIsLoggedIn()
+      })
+    }, [])
+
+    return <div>Page not found.</div>
+  },
+  { loading: () => null },
+)
+
+const PortalSiteNotFound = connect(
```
dcsaszar, line 79: * [ ] FYI to be more inline with the naming of the other components in this module: `NoSite`.

dcsaszar (changes requested):
See comments.

apepper:
REVIEW ([8da65e0](https://github.com/Scrivito/scrivito-portal-app/pull/586/commits/8da65e0676db91a4862e838d0b9fcda572d1cb62))

src/Components/NotFoundErrorPage.tsx
```diff
@@ -1,10 +1,33 @@
...
 // Make sure, that you have a proxy running for these URLs, otherwise you'll see an endless loop.
 const RELOAD_SUBPATHS = ['/auth']
 
-export const NotFoundErrorPage = connect(function NotFoundErrorPage() {
+export const NotFoundErrorPage = connect(
+  function NotFoundErrorPage() {
+    if (isUserLoggedIn() && isNoSitePresent()) return <NotFound />
+
+    return (
+      <>
```
dcsaszar, line 20: * [ ] FYI the fragment is not needed

dcsaszar (approved):
LGTM. One potential cleanup.

dcsaszar (approved):

apepper:
REVIEW ([92ab86e](https://github.com/Scrivito/scrivito-portal-app/pull/586/commits/92ab86e6b541de634cfc88f5ecf324d48c044883))

src/config/scrivitoSites.ts
```diff
@@ -86,7 +86,10 @@ export function isNoSitePresent(): boolean {
 export async function ensureSiteIsPresent() {
   if ((await load(currentSiteId)) === null) {
     if (await load(isNoSitePresent)) {
-      ensureUserIsLoggedIn()
+      // TODO: remove window check after #11101 is implemented
+      if (typeof window !== 'undefined') {
+        ensureUserIsLoggedIn()
+      }
```
dcsaszar, line 92: * [ ] We don't need this change.
  1. `ensureUserIsLoggedIn()` will still throw after #11101. However,
  2. This block is not in the prerendering code path

dcsaszar (changes requested):
See comment

apepper:
REVIEW ([2a2fe32](https://github.com/Scrivito/scrivito-portal-app/pull/586/commits/2a2fe322d37bff17362587964e798859bf8dca68))

PR #587, opened 2024-09-04
Avoid flash of color while loading color data

dcsaszar:

PR #588, opened 2024-09-04
Try to avoid unwanted color editing

dcsaszar:
Todo

* [ ] Set the colors in the Golden Master content

### DIFF
--- a/src/Objs/Homepage/SiteColorsPicker.tsx
+++ b/src/Objs/Homepage/SiteColorsPicker.tsx
@@ -1,6 +1,6 @@
 import Color from 'color'
 import { ColorPicker, useColor } from 'react-color-palette'
-import { connect, uiContext } from 'scrivito'
+import { canEdit, connect, uiContext } from 'scrivito'
 import { HomepageInstance } from './HomepageObjClass'
 import './SiteColorsPicker.scss'
 import { useState } from 'react'
@@ -24,6 +24,7 @@ export const SiteColorsPicker = connect(function SiteColorsPicker({
           </div>
           <AdvancedColorPicker
             color={page.get('siteColorPrimary') || '#274486'}
+            disabled={!canEdit(page)}
             setColor={(color) => {
               page.update({
                 siteColorPrimary: color,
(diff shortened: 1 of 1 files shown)


src/Objs/Homepage/HomepageEditingConfig.ts
```diff
@@ -89,6 +89,12 @@ provideEditingConfig(Homepage, {
   properties: [...defaultPageProperties],
   initialContent: {
     ...defaultPageInitialContent,
+    siteColorPrimary: '#274486',
+    siteColorPrimaryDarken: '#1E3466',
+    siteColorPrimaryLighten: '#3054A6',
+    siteColorSecondary: '#39a9eb',
+    siteColorSecondaryDarken: '#1798E4',
+    siteColorSecondaryLighten: '#5EB9EF',
```
apepper, line 97: Why do we need this? The value `274486` is already spread between `index.scss` and `SiteColorsPicker.tsx`. I would prefer not to add one more place.

apepper (commented):
FEEDBACK ([331ee0a](https://github.com/Scrivito/scrivito-portal-app/pull/588/commits/331ee0a42ef2909cb7a232d04e96af3c66f92419))

Good stuff. See comments for questions.

src/Objs/Homepage/HomepageEditingConfig.ts
dcsaszar: Working on #587 I realized that if the designer changes the default primary and secondary color *and* the color content is unset, the SiteColorsPicker shows the wrong color.
1. This change prevents the unset state for new homepages.
2. The todo in the description prevents the unset state for initial content.

We can probably stick with (2) and revert (1) since homepages are typically created as copies.

apepper: One understanding question:

> if the designer changes the default primary and secondary color

By that you mean, the designer adjusted `index.scss` and forgot to adjust `SiteColorsPicker.ts` as well?

apepper: I don't think that adding even more places, that a designer would have to adjust is a good idea. So given we have the color in the css, a default color in javascript and a default color in the content - if the designer only changes the css variable it would still only show the default color because of the content.

Because of that I would simply revert [35684d5](https://github.com/Scrivito/scrivito-portal-app/pull/588/commits/35684d5a296f8ee37fc70b9093c1abbefcd4ab92).

apepper (approved):
LGTM ([9752357](https://github.com/Scrivito/scrivito-portal-app/pull/588/commits/9752357968e44390ba230bb973d68660ebcd92ae))

Nice 🚀!

PR #593, opened 2024-09-09
Bugfix: Errors during data form submission no longer freeze the app

apepper:
https://github.com/Scrivito/scrivito-portal-app/pull/589 already provided a bugfix for a specific issue.

The issue could be reproduced, by trying to upload https://drive.google.com/file/d/1t5KUq5yxc9wYqNUI4FDm4yNsr3ukyR0z/view?usp=sharing (this can still be done in the PR). Since prior to https://github.com/Scrivito/scrivito-portal-app/pull/589 the mime type did not match and an error was passed to the `rejected` callback of the promise. This error message was shown in a `toast.error`.

But since the error message was quite long (~3 mb) it crashed the react rendering resulting in an endless "spinner".

This PR prevents the endless "spinner" by limiting the length of string passed to the `toast.error`. You still can't upload the file in this PR, but you will see an error message and be able to continue to use the application.

The bug could potentially also be not regarding the length of the string, but because of the content. It could also be, that this is a bug within React itself. I did not investigate further.

### DIFF
--- a/src/Data/CurrentUser/errorToast.tsx
+++ b/src/Data/CurrentUser/errorToast.tsx
@@ -3,7 +3,7 @@ import { toast } from 'react-toastify'
 import { ClientError } from 'scrivito'
 
 export function errorToast(title: string, error: unknown): void {
-  console.log('errorToast', title, error)
+  console.error('errorToast', title, error)
   const message =
     error instanceof Error ? error.message : JSON.stringify(error, null, 1)
 
--- a/src/Data/CurrentUser/errorToast.tsx
+++ b/src/Data/CurrentUser/errorToast.tsx
@@ -1,15 +1,15 @@
+import { truncate } from 'lodash-es'
 import { toast } from 'react-toastify'
 import { ClientError } from 'scrivito'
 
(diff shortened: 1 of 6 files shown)


src/Data/CurrentUser/errorToast.tsx
```diff
@@ -3,6 +3,7 @@ import { toast } from 'react-toastify'
 import { ClientError } from 'scrivito'
 
 export function errorToast(title: string, error: unknown): void {
+  console.log('errorToast', title, error)
```
dcsaszar, line 6: * [ ] `console.error` is the better choice b/c it provides the developer with the appropriate UI
* [ ] FYI an abstraction for logging errors could be nice at some point to allow for easy integration of Honeybadger etc.

dcsaszar (changes requested):
mostly LGTM

apepper:
REVIEW ([c19f825](https://github.com/Scrivito/scrivito-portal-app/pull/593/commits/c19f825d021dd0b0a316741e7f12302e62330971))

I also adjusted the title of the PR.

PR #596, opened 2024-09-10
second/third level nav

agessler:
First try..
Bildschirmfoto 2024-09-10 um 14 33 50


apepper:
Included in https://github.com/Scrivito/scrivito-portal-app/pull/608. Thanks!

PR #612, opened 2024-09-17
[lighthouse] Fail early, if prerendering failed

apepper:
Otherwise lighthouse will try to measure the performance for a non prerendered site.

Discovered in https://github.com/Scrivito/scrivito-portal-app/pull/586#pullrequestreview-2302362288.


dcsaszar (approved):
LGTM.

FYI In the past I've seen some cases where a failed prerender caused some hassle b/c there is no notification about prerendering failures.
However, I don't know how often prerendering fails *without* causing issues.

How about changing the default from warning to failing in general?

PR #614, opened 2024-09-18
Fail prerender fast

apepper:
Inspired by https://github.com/Scrivito/scrivito-portal-app/pull/612#pullrequestreview-2308873548  - closes #612.

Often times prerender errors are not noticed, because the prerender process will report about the error(s), but will not fail. So a "green" batch is often taken as "everything is ok".

This PR changes the behavior: Now all pages are prerendered, but if there is at least one failed prerender page, the whole process is marked as "failed".


dcsaszar (approved):
😻

PR #616, opened 2024-09-18
Static schema: Added enum attributes

apepper:
Also removed our custom "enum" code.


src/utils/useEnumOptions.ts
```diff
@@ -0,0 +1,33 @@
+import { useAttributeDefinition } from 'scrivito'
+
+// TODO: Remove/trim, once #11238 is resolved
+export function useEnumOptions(): Array<{ title: string; value: string }> {
```
dcsaszar, line 4: * [ ] FYI maybe put React hooks in a hooks directory

dcsaszar (approved):
LGTM

src/utils/useEnumOptions.ts
apepper: Hopefully this is only temporary, until #11238 lands.

PR #621, opened 2024-09-20
Load schema dynamically from PisaSales

apepper:


src/Data/CallbackRequest/LocalStorage/localStorageCallbackRequestDataClass.ts
```diff
@@ -1,8 +1,16 @@
+import { currentLanguage, load } from 'scrivito'
 import { provideLocalStorageDataClass } from '../../../utils/provideLocalStorageDataClass'
-import { DataClassAttributes } from '../../types'
+import { DataClassSchema } from '../../types'
 
-export function localStorageCallbackRequestDataClass(
-  attributes: DataClassAttributes,
-) {
+async function attributes(): Promise<DataClassSchema> {
```
dcsaszar, line 5: * [ ] FYI the function is not exported, so we don't necessarily need a return type definition. If we `return {...} as const`, we can get rid of the `import { DataClassSchema }`.

src/Data/filterSchema.ts
```diff
@@ -0,0 +1,52 @@
...
+      .filter(
+        ([_, attribute]) => !['boolean', 'reference'].includes(attribute[0]),
+      )
+      .map(([key, value]) => [
+        key,
+        value[0] === 'enum'
+          ? ['enum', correctWrongEnumValues(value[1])]
+          : value,
+      ]),
+  )
```
dcsaszar, line 26: * [ ] FYI could be simplified with more destructuring:

```ts
  return Object.fromEntries(
    Object.entries(schema.attributes)
      .filter(([, [type]]) => !['boolean', 'reference'].includes(type))
      .map(([key, [type, options]]) => [
        key,
        [type, type === 'enum' ? correctWrongEnumValues(options) : options],
      ]),
  )
```

src/Data/filterSchema.ts
```diff
@@ -0,0 +1,52 @@
+import { pisaClient } from './pisaClient'
+import { DataClassAttributes } from './types'
+
+/**
+ * Fetches the schema for `subPath`.
+ * Filters out boolean and reference attributes, since they are currently not fully supported.
+ * It also converts enum values, if in a wrong format.
+ **/
+export async function filterSchema(subPath: string) {
```
dcsaszar, line 9: * [ ] FYI Missing return type. Not sure which policy we use. There are other exported functions in this PR without return type.
* [ ] FYI since we're only returning the `attributes` part the function name is misleading.

src/Data/filterSchema.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type LocalizedEnumValueConfig = {
+  value: string
+  title: string
+}
+
+type EnumValueConfig = string | LocalizedEnumValueConfig
+
+function correctWrongEnumValues({
```
dcsaszar, line 36: * [ ] Why do we need this? According to the Scrivito SDK type there's nothing wrong with string-only values. Am I missing something? https://github.com/Scrivito/scrivito_sdk_js/blob/1446f9ab9a89c0ec105f3cdd03b130d259bae3a2/src/scrivito_sdk/data_integration/data_class_schema.ts#L75

src/Data/CallbackRequest/Pisa/pisaCallbackRequestDataClass.ts
```diff
@@ -1,16 +1,12 @@
...
-} from '../../types'
+import { DataConnection, DataIndexResponse, RawItem } from '../../types'
+import { filterSchema } from '../../filterSchema'
 
-export function pisaCallbackRequestDataClass(attributes: DataClassAttributes) {
+export function pisaCallbackRequestDataClass() {
   // callback-request is more or less a "singleton". It only offers PUT, GET and DELETE.
   return provideDataClass('CallbackRequest', {
-    attributes,
+    attributes: () => filterSchema('callback-request'),
```
dcsaszar, line 9: * [ ] FYI I don't see why we need a function wrapper. How about `attributes: filterSchema('callback-request')`?

dcsaszar (approved):
LGTM so far. See comments.

src/Data/CallbackRequest/Pisa/pisaCallbackRequestDataClass.ts
apepper: Without the function wrapper `filterSchema` would be requested for all data classes during boot - regardless of them being used or not. With the function wrapper the attributes are only requested, if the data class is in use.

src/Data/CallbackRequest/LocalStorage/localStorageCallbackRequestDataClass.ts
dcsaszar: I see. Yes sometimes TS errors are a mess. Actually the last line of the convoluted report contains the important part here. 😆

PR #625, opened 2024-09-25
Show a spinner for not-yet-loaded options

dcsaszar:
I made sure a form with a required enum can't be sent.

There are two other widgets using `useEnumOptions` (DataIconWidget and DataLabelWidget), but every instance of these is used inside data or detail pages, so their loading behavior already matches how the rest of the page behaves.

image

### DIFF
--- a/src/Components/Loading.tsx
+++ b/src/Components/Loading.tsx
@@ -1,15 +1,29 @@
+import { PropsWithChildren } from 'react'
 import { connect, currentLanguage } from 'scrivito'
 
-export const Loading = connect(function Loading() {
+export const Loading = (
+  props: object | PropsWithChildren<{ className?: string }> = {},
+) => <LoadingPlaceholder {...props} />
+
+const LoadingPlaceholder = connect(function LoadingPlaceholder({
+  children,
+  className,
+}: PropsWithChildren<{ className?: string }>) {
+  const classNames = ['loading-placeholder']
+  if (className) classNames.push(className)
+
   const message = getMessage()
+
   return (
     <div
       aria-busy="true"
       aria-valuetext={message}
-      className="loading-placeholder"
+      className={classNames.join(' ')}
       role="progressbar"
       title={message}
-    />
+    >
+      {children}
+    </div>
   )
 })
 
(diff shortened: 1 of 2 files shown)


src/Widgets/DataFormOptionsWidget/DataFormOptionsWidgetComponent.tsx
```diff
@@ -61,28 +60,64 @@ provideComponent(DataFormOptionsWidget, ({ widget }) => {
...
+  },
+  {
+    loading: ({ isRequired }: { isRequired: boolean }) => (
+      <div
+        aria-busy="true"
+        className="w-100 loading-placeholder"
+        role="progressbar"
+      >
+        <select className="form-select" required={isRequired} />
+      </div>
```
apepper, line 120: How about extending the `Loading` component with two new options:
```ts
{
  children?: React.ReactNode
  className?: string
}
```

With that the above could reuse `Loading`:

```suggestion
      <Loading className="w-100">
        <select className="form-select" required={isRequired} />
      </Loading>
```

apepper (commented):
FEEDBACK ([10a9969](https://github.com/Scrivito/scrivito-portal-app/pull/625/commits/10a99694a2e89142474d250974037e7e15f0a982))

Very nice 🚀!

I left one improvement.

src/Components/Loading.tsx
```diff
@@ -1,15 +1,29 @@
+import { PropsWithChildren } from 'react'
 import { connect, currentLanguage } from 'scrivito'
 
-export const Loading = connect(function Loading() {
+export const Loading = (
+  props: object | PropsWithChildren<{ className?: string }> = {},
+) => <LoadingPlaceholder {...props} />
```
apepper, line 6: This looks a bit strange to me. But I have not found a different version, that passes tsc. Good enough for me.

apepper (approved):
LGTM ([9fb9c50](https://github.com/Scrivito/scrivito-portal-app/pull/625/commits/9fb9c5044e4b2d6e51acc3a8477d6896fbf3d221))

See comments. Feel free to merge.

PR #626, opened 2024-09-26
Improved DataWidget loading spinner + one Bugfix

apepper:
Adds a `<Loading />` spinner to most DataX widgets.

### DIFF
--- a/src/Widgets/DataIconWidget/DataIconWidgetComponent.tsx
+++ b/src/Widgets/DataIconWidget/DataIconWidgetComponent.tsx
@@ -15,7 +15,6 @@ import {
 import { IconComponent } from '../../Components/Icon'
 import { ensureString } from '../../utils/ensureString'
 import { useEnumOptions } from '../../utils/useEnumOptions'
-import { Loading } from '../../Components/Loading'
 
 provideComponent(DataIconWidget, ({ widget }) => {
   return (
@@ -69,7 +68,7 @@ const Icon = connect(
       />
     )
   },
-  { loading: Loading },
+  { loading: () => null },
 )
 
 function localizeNotAvailable(): string {
--- a/src/Widgets/DataIconWidget/DataIconWidgetComponent.tsx
+++ b/src/Widgets/DataIconWidget/DataIconWidgetComponent.tsx
@@ -15,6 +15,7 @@ import {
 import { IconComponent } from '../../Components/Icon'
 import { ensureString } from '../../utils/ensureString'
 import { useEnumOptions } from '../../utils/useEnumOptions'
+import { Loading } from '../../Components/Loading'
 
 provideComponent(DataIconWidget, ({ widget }) => {
   return (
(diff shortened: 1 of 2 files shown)


src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
```diff
@@ -50,34 +51,37 @@ provideComponent(DataLabelWidget, ({ widget }) => {
...
+    }
+    if (showAs === 'link') return <Link value={attributeValue} />
+
+    if (typeof attributeValue === 'number') {
+      return <NumberText value={attributeValue} />
+    }
+
+    return <Text value={attributeValue} />
+  },
+  { loading: Loading },
```
dcsaszar, line 83: * [ ] FYI Typically many labels are used on the same details page. Using the regular Loading spinner for each data label introduces a lot of animation in many places. A visitor may be confused by seeing more than one spinner and the page may look broken. I suggest to use a more decent placeholder instead of a spinner in these places. Quick prototype: `{ loading: () => <span className="opacity-25">▆▆▆▆</span> }`


image
image

src/Widgets/DataIconWidget/DataIconWidgetComponent.tsx
```diff
@@ -1,23 +1,48 @@
...
+        conditions={widget.get('conditions').filter(isDataIconConditionWidget)}
+        fallbackIcon={widget.get('fallbackIcon') || 'bi-question-octagon'}
+        size={widget.get('size') || 'bi-2x'}
+      />
+    </WidgetTag>
+  )
+})
 
-  const conditions = widget.get('conditions').filter(isDataIconConditionWidget)
+const AttributeValue = connect(function AttributeValue({
```
dcsaszar: * [ ] FYI the name `AttributeValue` reads a bit strange, as we actually render an icon for the attribute value. How about `Icon`?

src/Widgets/DataIconWidget/DataIconWidgetComponent.tsx
```diff
@@ -33,40 +34,43 @@ provideComponent(DataIconWidget, ({ widget }) => {
...
+            ? matchingCondition.get('icon') || 'bi-box'
+            : fallbackIcon
+        }
+        size={size}
+        link={null}
+        title={title || localizeNotAvailable()}
+      />
+    )
+  },
+  { loading: Loading },
```
dcsaszar, line 72: * [ ] FYI since the icon is basically an image we could go with a more decent placeholder like the generic browser default for images that are still loading. E.g. `{ loading: () => null }`

dcsaszar (approved):
Makes sense in general. But I think the visuals should be improved.

src/Widgets/DataLabelWidget/DataLabelWidgetComponent.tsx
apepper: To me this is good enough for now. We can improve later on, if there is demand.

apepper:
REVIEW ([7368992](https://github.com/Scrivito/scrivito-portal-app/pull/626/commits/7368992654fa525fe334f3bcb2c26624873e23bd))

PR #630, opened 2024-09-30
Update usability of vertical nav

dcsaszar:
* Non-active paths close on nav
* Active paths open on nav
* Remove support for following the hierarchy without closing or scrolling

### DIFF
--- a/src/Widgets/VerticalNavigationWidget/VerticalNavigationWidgetComponent.tsx
+++ b/src/Widgets/VerticalNavigationWidget/VerticalNavigationWidgetComponent.tsx
@@ -11,7 +11,6 @@ import Navbar from 'react-bootstrap/Navbar'
 import Nav from 'react-bootstrap/Nav'
 import { ObjIconAndTitle } from '../../Components/ObjIconAndTitle'
 import { VerticalNavigationWidget } from './VerticalNavigationWidgetClass'
-import { useEffect, useState } from 'react'
 
 provideComponent(VerticalNavigationWidget, ({ widget }) => {
   const page = widget.obj()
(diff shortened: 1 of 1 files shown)


src/Widgets/VerticalNavigationWidget/VerticalNavigationWidgetComponent.tsx
```diff
@@ -84,40 +84,32 @@ const SubNavItems = connect(
...
         <Nav.Link as={LinkTag} eventKey={key} key={key} to={obj}>
           <button
-            className={`dropdown-toggle nav-link${isExpanded ? ' show' : ''}`}
+            className={`dropdown-toggle nav-link${isActive ? ' show' : ''}`}
             onClick={(e) => {
               e.preventDefault()
               e.stopPropagation()
-              setIsExpanded((isExpanded) => !isExpanded)
+              navigateTo(obj, { hash: '›' })
             }}
```
apepper, line 107: I don't think that the `onClick` is needed anymore. I would simply remove it.

src/Widgets/VerticalNavigationWidget/VerticalNavigationWidgetComponent.tsx
```diff
@@ -84,40 +84,32 @@ const SubNavItems = connect(
 
 const ExpandableNavItem = connect(
```
apepper, line 85: This is no longer "expandable". I suggest to rename it to `NestedNavItem`. I would also change the `key` to `VerticalNavigationWidget-nested-${obj.id()}`.

apepper (commented):
FEEDBACK ([8d259b8](https://github.com/Scrivito/scrivito-portal-app/pull/630/commits/8d259b8e882cb5996f3d4dd3c82b239ca8fc14c1))

Two smaller issues (see comments).

Other then that really nice!

src/Widgets/VerticalNavigationWidget/VerticalNavigationWidgetComponent.tsx
dcsaszar: If we remove this, a click on the toggle button would behave exactly like a click on the outer element:
* close the nav b/c of `collapseOnSelect`
* scroll to the top of the page b/c of the default `LinkTag`/`navigateTo` implementation
 
A mobile visitor must reopen the menu to reach a child item.
All visitors (but it's more likely on mobile) may have to scroll down to reach a child item.

I'm fine with one or both of these consequences. WDYT?

apepper: Yes, please remove this. Maybe also remove the button and make it a simple span.

dcsaszar:
REVIEW

We should discuss the mobile behavior with @agessler.
All the simple approaches have some issues with nested navs.
Maybe removing the `collapseOnSelect` is the preferred option for the nested case?

apepper:
> Maybe removing the `collapseOnSelect` is the preferred option for the nested case?

Sounds like a reasonable suggestion. Can you try it out?

apepper (approved):
LGTM ([a4e0ae5](https://github.com/Scrivito/scrivito-portal-app/pull/630/commits/a4e0ae5c9cb35681ddeda16da7ca3cb494a760de))

Nice 🚀!

Let's keep the `collapseOnSelect` (option (B)) for now and wait for actual user feedback if this needs any changing.

PR #636, opened 2024-10-02
DataSearchWidget: Add aria-label for better accessibility

apepper:


src/Widgets/DataSearchWidget/DataSearchWidgetComponent.tsx
```diff
@@ -43,7 +44,11 @@ provideComponent(DataSearchWidget, ({ widget }) => {
         defaultValue={search}
       />
       <span className="input-group-btn">
-        <button className={`btn ${buttonColor}`} type="submit">
+        <button
+          className={`btn ${buttonColor}`}
+          type="submit"
+          aria-label={localizeSearchInputLabel()}
```
dcsaszar, line 50: * [ ] FYI looking at the current values for `placeholder` (typically "Search for X") we could repurpose these and use them for both the placeholder and the button label. If we do so, it should be made clear in the editing config though.

dcsaszar (approved):
See idea/comment.

apepper:
REVIEW ([bf298f7](https://github.com/Scrivito/scrivito-portal-app/pull/636/commits/bf298f72b382fcc0e8a2ac47592132e4d998e7fa))

src/Widgets/DataSearchWidget/DataSearchWidgetEditingConfig.ts
```diff
@@ -17,7 +17,11 @@ provideEditingConfig(DataSearchWidget, {
         { value: 'btn-secondary', title: 'Secondary color' },
       ],
     },
-    placeholder: { title: 'Placeholder text' },
+    placeholder: {
+      title: 'Placeholder text',
+      description:
+        'This value is also utilized for the aria label of the search button.',
```
dcsaszar, line 23: * [ ] FYI ARIA is uppercase (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA)

dcsaszar (approved):

PR #637, opened 2024-10-02
DIV.headline fix

agessler:


apepper:
Thanks! Incorporated in https://github.com/Scrivito/scrivito-portal-app/pull/642.

PR #642, opened 2024-10-02
Headline with level div.

apepper:
Allows to keep the styling of a heading, but remove it from the semantic hierarchy.

### DIFF
--- a/src/Widgets/HeadlineWidget/HeadlineWidgetClass.ts
+++ b/src/Widgets/HeadlineWidget/HeadlineWidgetClass.ts
@@ -4,7 +4,7 @@ export const HeadlineWidget = provideWidgetClass('HeadlineWidget', {
   attributes: {
     alignment: ['enum', { values: ['left', 'center', 'right'] }],
     headline: 'string',
-    level: ['enum', { values: ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'] }],
+    level: ['enum', { values: ['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'div'] }],
     style: [
       'enum',
       {
--- a/src/assets/stylesheets/index.scss
+++ b/src/assets/stylesheets/index.scss
@@ -404,19 +404,6 @@ h6 {
   margin-bottom: 0.7rem;
 }
 
-div {
-  &.h1,
-  &.h2,
-  &.h3,
-  &.h4,
-  &.h5,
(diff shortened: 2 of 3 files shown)


src/Widgets/HeadlineWidget/HeadlineWidgetClass.ts
```diff
@@ -4,7 +4,7 @@ export const HeadlineWidget = provideWidgetClass('HeadlineWidget', {
   attributes: {
     alignment: ['enum', { values: ['left', 'center', 'right'] }],
     headline: 'string',
-    level: ['enum', { values: ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'] }],
+    level: ['enum', { values: ['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'div'] }],
```
dcsaszar, line 7: * [ ] FYI `level: "div"` doesn't make a lot of sense. Maybe a neutral value like `none`/`text` is a better fit?
* [ ] FYI Similarly in the editing config "div" is not really something a SEO editor cares about, maybe something like "Text only" or "No semantic tag" makes more sense.

src/assets/stylesheets/index.scss
```diff
@@ -404,19 +404,6 @@ h6 {
   margin-bottom: 0.7rem;
 }
 
-div {
```
dcsaszar, line 407: * [ ] The styling of "Label headline" and "Label subtitle" changes when switching to the no-heading level. Changing the SEO level should never change the chosen style.
image
image

dcsaszar (changes requested):

src/assets/stylesheets/index.scss
apepper: I think this is an independent issues with `Label heading` and `Label subtitle` that already existed prior to this PR. I think it should be addressed separately, once our designer is available again.

dcsaszar: I'm confused. We introduce `div` headlines in this PR. How is it even possible for a `div` headline related issue to have existed before that?

apepper: `Label heading` and `Label subtitle` will already (prior to this PR) automatically select `<div>` if no heading tag is set.

dcsaszar: I see, it's the "unset" case. Thanks for clarifying!

PR #645, opened 2024-10-09
provideLocalStorageDataClass: Separate localstorage by tenantId

apepper:
This way - in a multi tenancy setup - the data from one tenant does _not_ show up in another tenant.

Please note, that all existing localstorage data will no longer be readable.

### DIFF
--- a/src/Data/CallbackRequest/LocalStorage/localStorageCallbackRequestDataClass.ts
+++ b/src/Data/CallbackRequest/LocalStorage/localStorageCallbackRequestDataClass.ts
@@ -1,5 +1,5 @@
 import { currentLanguage, load } from 'scrivito'
-import { provideLocalStorageDataClass } from '../../../utils/provideLocalStorageDataClass'
+import { provideLocalStorageDataClass } from '../../provideLocalStorageDataClass'
 import { DataClassSchema } from '../../types'
 
 async function attributes(): Promise<DataClassSchema> {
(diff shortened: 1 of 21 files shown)


src/utils/provideLocalStorageDataClass.ts
```diff
@@ -3,6 +3,7 @@ import { pseudoRandom32CharHex } from './pseudoRandom32CharHex'
 import { orderBy } from 'lodash-es'
 import { ensureString } from './ensureString'
 import { DataClassAttributes } from '../Data/types'
+import { scrivitoTenantId } from '../config/scrivitoTenants'
```
dcsaszar, line 6: * [ ] The `config` package already has a dependency on `utils`. Let's not introduce the other direction and move `scrivitoTenants` to utils.

dcsaszar (changes requested):
LGTM but see comment

src/utils/provideLocalStorageDataClass.ts
apepper: I see you point. But I don't really think, that `scrivitoTenants` is a good candidate for `utils`. I moved `provideLocalStorageDataClass` instead to `src/Data` to avoid the circular package dependency.

apepper:
REVIEW ([c2128b3](https://github.com/Scrivito/scrivito-portal-app/pull/645/commits/c2128b3110c0e450b347dcae458aa8bae5d5c638))

dcsaszar (approved):

PR #632, opened 2024-10-01
New primary/secondary colors for improved color contrast

agessler:
@apepper unglier but better contrast - I could try other combinations to improve visual design. Wave test is now happy on every page.

**OLD**
```
--bs-primary: #274486
--bs-secondary: #39a9eb
--bs-body-color:  #677788;
```

**NEW**
```
--bs-primary: #072b46; 
--bs-secondary: #0e5588; 
--bs-body-color: #4f557a;
```

PR #648, opened 2024-10-11
Small color contrast improvements

apepper:
Cherry-picks most stuff from https://github.com/Scrivito/scrivito-portal-app/pull/632 except for the primary/secondary color change.

With that the count of "axe" is reduced from 18 to 3 issues.


agessler (approved):
🎉

PR #649, opened 2024-10-11
Cart: Now officially reference a product

apepper:
Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://cart-reference.scrivito-portal-app.pages.dev/en/cart-a8ede3121df9e2f6?_scrivito_workspace_id=n35dbe1f47caf0ad&_scrivito_display_mode=view

Renamed `productId` to `product`. If someone still has some items in their cart left, the cart checkout page will not be able to display the products properly and will not be able to post an opportunity.

### DIFF
--- a/src/Data/CartItem/Cart.ts
+++ b/src/Data/CartItem/Cart.ts
@@ -5,21 +5,23 @@ import {
   isUserLoggedIn,
   load,
 } from 'scrivito'
-import { Product, ProductInstance } from '../../Objs/Product/ProductObjClass'
+import { ProductInstance } from '../../Objs/Product/ProductObjClass'
 import { CartItem } from './CartItemDataClass'
 import { Opportunity } from '../Opportunity/OpportunityDataClass'
 
 export async function addToCart(product: ProductInstance): Promise<void> {
   const productId = product.id()
 
-  await CartItem.create({ productId })
+  await CartItem.create({ product: productId })
 }
 
 export async function removeFromCart(product: ProductInstance): Promise<void> {
   const productId = product.id()
 
   const items: DataItem[] = await load(() =>
-    CartItem.all().transform({ filters: { productId } }).take(),
+    CartItem.all()
+      .transform({ filters: { product: productId } })
+      .take(),
   )
 
   items.forEach((item) => item.delete())
@@ -30,7 +32,9 @@ export function isInCart(product: ProductInstance): boolean {
 
   const productId = product.id()
 
-  return CartItem.all().transform({ filters: { productId } }).containsData()
+  return CartItem.all()
+    .transform({ filters: { product: productId } })
+    .containsData()
 }
 
 export function containsItems(): boolean {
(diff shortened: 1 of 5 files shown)


src/Data/CartItem/Cart.ts
```diff
@@ -48,14 +52,9 @@ export function numberOfCartItems(): number | null {
...
-  const products: ProductInstance[] = []
-  for (const item of cartItems) {
-    const productId = item.get('productId')
-    if (typeof productId !== 'string') continue
-
-    const product = await load(() => Product.get(productId))
-    if (product) products.push(product)
-  }
+  const products: DataItem[] = cartItems.map(
+    (item) => item.get('product') as DataItem,
```
dcsaszar, line 56: * [ ] FYI the validity of the cast depends on multiple conditions spread over the codebase.
  * How can I be sure product is a reference attribute?
  * How can I be sure the reference is not null?
  
   Maybe we should use a type guard.

dcsaszar (approved):
See comment.

apepper:
REVIEW ([b68befb](https://github.com/Scrivito/scrivito-portal-app/pull/649/commits/b68befb29f187afde98f57e5aafd5e311d77d084))

src/Data/CartItem/Cart.ts
```diff
@@ -52,9 +53,9 @@ export function numberOfCartItems(): number | null {
 export async function checkoutCart(): Promise<DataItem> {
   const cartItems: DataItem[] = await load(() => CartItem.all().take())
 
-  const products: DataItem[] = cartItems.map(
-    (item) => item.get('product') as DataItem,
-  )
+  const products: DataItem[] = cartItems
+    .map((item) => item.get('product'))
+    .filter(isDataItem)
```
dcsaszar, line 58: * [ ] How about `.filter((item) => item instanceof DataItem)`?

dcsaszar (changes requested):

src/Data/CartItem/Cart.ts
apepper: `DataItem` is only defined as an abstract typescript class and is not available during the javascript runtime.

dcsaszar: What do you mean by "abstract typescript class"? Is there an issue or some other pointer that explains why typeof doesn't work for abstract classes?
The class itself is public API.

image

apepper: The exact error is `TypeError: Right-hand side of 'instanceof' is not an object`.

*How to reproduce*
* Change `export function isDataItem(item: unknown): item is DataItem` to `return item instanceof DataItem` (and adjust the import if needed)
* Add something to the cart
* Try to "Request quote"

*Expected*
The checkout works regular

*Actual*
An error toast is shown:
Screenshot 2024-10-15 at 08 47 08

Stack-trace:

```
TypeError: Right-hand side of 'instanceof' is not an object
    at isDataItem (types.ts:29:10)
    at Array.filter (<anonymous>)
    at checkoutCart (Cart.ts:58:6)
    at async onClick (CheckoutButtonWidgetComponent.tsx:58:22)
```

dcsaszar: Hmm, I see. 

```ts
export type {
  DataClass,
  DataItem,
  DataScope,
} from 'scrivito_sdk/data_integration';
```

We're basically adding workaround for the SDK exporting only the type. Maybe we could change that?

apepper: Striking the `type` from `export type` sounds like a good idea (originally introduced by me in #10342). I'll prepare a PR.

apepper:
REVIEW ([b3a9633](https://github.com/Scrivito/scrivito-portal-app/pull/649/commits/b3a96337298230212a32ce073bd4f6d2c3762dca))

dcsaszar (approved):

PR #655, opened 2024-10-17
siteForUrl: Bugfix - ignore query and search from input URL

apepper:
Otherwise hash links like https://scrivito-portal-app.pages.dev/en#ecoline will not work. With this PR it works (see https://siteforurl-ignores-hash-and.scrivito-portal-app.pages.dev/en#ecoline).

### DIFF
--- a/src/config/scrivitoSites.ts
+++ b/src/config/scrivitoSites.ts
@@ -52,7 +52,7 @@ export function siteForUrl(
 
 function languageForUrl(url: string) {
   const regex = new RegExp(
-    `^${getBaseAppUrl()}\\/(?<lang>[a-z]{2}(-[A-Z]{2})?)([?/]|$)`,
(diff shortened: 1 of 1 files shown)


src/config/scrivitoSites.ts
```diff
@@ -33,8 +33,11 @@ export function baseUrlForSite(siteId: string): string | undefined {
 }
 
 export function siteForUrl(
-  url: string,
+  fullUrl: string,
 ): { baseUrl: string; siteId: string } | undefined {
+  const parsedUrl = new URL(fullUrl)
+  const url = `${parsedUrl.origin}${parsedUrl.pathname}`
+
```
dcsaszar, line 40: * [ ] The easier fix is to include `#` in the language detection regex. WDYT?

```ts
`^${getBaseAppUrl()}\\/(?<lang>[a-z]{2}(-[A-Z]{2})?)([?/#]|$)`,
```

PR #660, opened 2024-10-25
Remove type workaround for local storage data

dcsaszar:
This type was used module-internal only anyway.
The typed return values could be reintroduced with a publicly exported type later on.

The PR is a suggestion only. Feel free to stick with the workaround and close it.


markus-codewrangler:
IMHO the portal-app should not recreate _any_ types from the SDK (see `types.ts`), instead any missing type exports should be reported to the SDK-team so this can be fixed – and _all_ recreated SDK-types can be removed from the portal-app code with the next SDK release 🧹

Otherwise we are loosing some type checking of SDK releases through the portal-app as an early-adopter 😢 

WDYT?

Probably out of scope for this PR.

/cc @kostia

markus-codewrangler:
This should also be replaced by a type from the SDK IMHO:

https://github.com/Scrivito/scrivito-portal-app/blob/3a2d6b1f6f31dd87ee70276e1c932087bd70f943/src/Objs/defaultPageLayoutComponent.tsx#L9

PR #669, opened 2024-10-31
French and polish localizers

apepper:


src/Objs/ProductCategory/ProductCategoryComponent.tsx
```diff
@@ -67,4 +67,16 @@ const LOCALIZERS = {
...
+    headline: 'Catégorie de produit',
+    items0: 'Aucun article',
+    items1: '1 article',
+    itemsMany: '__count__ articles',
+  },
+  pl: {
+    headline: 'Kategoria produktu',
+    items0: 'Brak produktów',
+    items1: '1 produkt',
+    itemsMany: '__count__ produktów',
```
patrykboch, line 80: "0 produktów" is correct, but "1 produkt," "2-3-4 produkty," and only 5 and beyond will match the form "produktów" to 21 🤣 because "22 produkty" - requires the modified form once again. Depending on the number before the verb, "produkt" can appear in at least 12 different ways in Polish, if I am right IDK, maybe more. 

I would use "Liczba produktów: count" (Means: Number of products: X) if we wanted to make it more generic.

src/Widgets/DataFormUploadWidget/DataFormUploadWidgetComponent.tsx
```diff
@@ -146,6 +154,16 @@ function getTooLargeMessage() {
...
       )} hoch.`
+    case 'fr':
+      return `Un ou plusieurs fichiers sont trop volumineux. Veuillez télécharger des fichiers d'une taille maximale de ${prettyBytes(
+        MAX_FILE_SIZE,
+        { locale: 'fr' },
+      )}.`
+    case 'pl':
+      return `Jeden lub kilka plików jest za duży. Proszę przesłać pliki o maksymalnym rozmiarze ${prettyBytes(
+        MAX_FILE_SIZE,
+        { locale: 'pl' },
```
patrykboch, line 165: `Za duży rozmiar pliku lub plików. Maksymalny rozmiar to X`

patrykboch (commented):
I reviewed polish translations. See comments.

src/Objs/ProductCategory/ProductCategoryComponent.tsx
apepper: Wow, I didn't know that counting can be so complicated. Thanks for pointing this out!

patrykboch (commented):
PL translations LGTM 🥳

PR #679, opened 2024-11-07
Vite: adjust chunkSizeWarningLimit

apepper:


vite.config.ts
```diff
@@ -18,6 +18,7 @@ export default defineConfig(({ mode }) => {
 
   return {
     build: {
+      chunkSizeWarningLimit: 900,
```
dcsaszar, line 21: * [ ] FMI We're just muting the warning without any comment why. The warning is there for a reason and it even gives instructions on how to handle it. Do we have a backlog item to fix the underlying problems?

.circleci/config.yml
```diff
@@ -37,6 +37,9 @@ workflows:
           npm-run-command: build
           setup:
             - run: node --version
+            - run:
+                name: Set ENABLE_PISA
+                command: echo 'export ENABLE_PISA=true' >> $BASH_ENV
```
dcsaszar, line 42: * [ ] FYI setting the job env could be cleaner: https://circleci.com/docs/set-environment-variable/#set-an-environment-variable-in-a-job

apepper: See https://github.com/Scrivito/scrivito-portal-app/pull/717

PR #684, opened 2024-11-15
test site for safari bug

agessler:


apepper:
Included in https://github.com/Scrivito/scrivito-portal-app/pull/687.

PR #691, opened 2024-11-22
Variable manrope font

apepper:
Replaced the existing multiple manrope variants with one single variable font file.


agessler:
The new variable font file does not contain all weights..

Your variable font file size **52 kB**
Original variable font file size **165 kB**

We can go this way but is is deceptive offer.

[Manrope-VariableFont_wght.ttf.zip](https://github.com/user-attachments/files/17903932/Manrope-VariableFont_wght.ttf.zip)

agessler:
The Google File is not a WOFF 2. but I think in your file are parts missing.

apepper:
> [Manrope-VariableFont_wght.ttf.zip](https://github.com/user-attachments/files/17903932/Manrope-VariableFont_wght.ttf.zip)

If I unzip this file and run `woff2_compress` on it, I receive the exact `src/assets/fonts/Manrope-VariableFont_wght.woff2` file of this PR.

agessler (approved):

PR #693, opened 2024-11-25
NotFoundErrorPage: Added get started button

apepper:
Adds a "Get started with your CMS" button if the given tenant is empty. A visit to that link with automatically initialize the content for the tenant.

Screenshot 2024-11-25 at 14 50 10

### DIFF
--- a/src/Components/NotFoundErrorPage.tsx
+++ b/src/Components/NotFoundErrorPage.tsx
@@ -2,6 +2,7 @@ import { useEffect } from 'react'
 import {
   connect,
   ContentTag,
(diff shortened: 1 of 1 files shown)


dcsaszar (commented):
* [ ] FYI Instead of a "you have no choice but to click one button" UX I'd prefer an automatic redirect, but this can be improved in a follow-up.

apepper:
> FYI Instead of a "you have no choice but to click one button" UX I'd prefer an automatic redirect, but this can be improved in a follow-up.

Feel free to create a PR for this.

PR #699, opened 2024-12-02
VideoWidget: Option to hide download button

apepper:
This option is currently only supported by chromium-based browsers (Chrome, Edge, Brave etc.). Firefox and Safari do _not_ support this option, since they both don't offer a download button at all (see https://github.com/whatwg/html/pull/6715).

### DIFF
--- a/src/Widgets/VideoWidget/VideoWidgetEditingConfig.ts
+++ b/src/Widgets/VideoWidget/VideoWidgetEditingConfig.ts
@@ -29,8 +29,12 @@ provideEditingConfig(VideoWidget, {
         'This poster image is shown, until the video is loaded.' +
         ' Without an poster image, the browser may show the first frame of the video.',
     },
+    noDownload: {
+      title: 'Hide download button?',
+      description: 'Only supported by some browsers.',
+    },
   },
-  properties: ['source', 'poster', 'aspectRatio'],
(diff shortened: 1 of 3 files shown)


src/Widgets/VideoWidget/VideoWidgetEditingConfig.ts
```diff
@@ -29,8 +29,12 @@ provideEditingConfig(VideoWidget, {
         'This poster image is shown, until the video is loaded.' +
         ' Without an poster image, the browser may show the first frame of the video.',
     },
+    noDownload: {
+      title: 'Hide download button?',
+      description: 'Only supported by some browsers.',
```
dcsaszar, line 34: * This simple wording has the potential to confuse editors. We should make clear that this is just a presentational hint and not all visitors will see the difference.
* The new option lets editors configure a very minor cosmetic detail, which in fact may reduce the usability of the video for some visitors.
* There appears to be back-pressure from browser vendors to support this option.

I'd recommend to not add this feature. If a project knows the limitations and is willing to accept them, a developer can easily add the respective option. The base implementation should stick to widely supported options.

dcsaszar (changes requested):
See comment.

In the PR description I don't see a strong motivation for the feature, thus my suggestion is to close the PR.

PR #700, opened 2024-12-03
Data: reverseTitle attribute

apepper:
Added `reverseTitle` to applicable `'reference'` data attributes. Please note that not all `'reference'` have a `' reverseTitle` - only those that should also display the `1:n` relationship.

PisaSales also needs to provide these `reverseTitle` attributes (still open).


apepper:
FYI: PisaSales now offers the `reverseTitle` as well.

apepper:
Will be available with PisaSales `8.2 Patch 16` or `8.3 Patch 2`.

PR #701, opened 2024-12-04
Font design improvements

apepper:


src/assets/stylesheets/index.scss
```diff
@@ -1148,7 +1149,7 @@ section:focus >,
   border-radius: 50%;
   line-height: 60px;
   text-align: center;
-  font-weight: 500;
+  font-weight: var(--bs-body-font-weight);
```
apepper, line 1152: This attribute can be removed, since it will automatically take the body font weight.

src/assets/stylesheets/index.scss
```diff
@@ -1187,7 +1188,8 @@ section:focus >,
 // -- general form-control -- //
 
 .form-label {
-  font-family: var(--jr-headline-font-family), sans-serif;
+  font-family: var(--bs-body-font-family), sans-serif;
```
apepper, line 1191: Same here (just for `font-family`).

src/assets/stylesheets/index.scss
```diff
@@ -1187,7 +1188,8 @@ section:focus >,
...
+  font-weight: var(--bs-body-font-weight);
```
apepper, line 1192: How about `font-weight: bolder` here?

src/Components/customToastify.css
```diff
@@ -4,5 +4,5 @@
   --toastify-color-warning: var(--bs-warning);
   --toastify-color-error: var(--bs-danger);
 
-  --toastify-font-family: var(--jr-headline-font-family);
+  --toastify-font-family: var(--bs-body-font-family);
```
apepper, line 7: Do we actually have to provide a `toastify-font-family` if we already use the body font? Is there some default from the library that we need to overwrite?

src/assets/stylesheets/index.scss
```diff
@@ -298,6 +299,7 @@ html {
 
 body {
   font-family: var(--bs-body-font-family);
```
apepper, line 301: With other font-families we ususaly add a `, sans-serif;`. Is this useful to add it for all font families?

src/assets/stylesheets/theme/fonts.scss
```diff
@@ -1,6 +1,28 @@
...
 ================================================== */
 
+/* FiraSans-Light 300 */
+@font-face {
+  font-family: 'Fira Sans';
+  font-display: swap;
+  font-style: normal;
+  font-weight: 300;
+  src:
+    local(''),
```
apepper, line 11: FMI: What is this `local('')` for? Which effect does it have?

apepper (commented):
FEEDBACK ([3d6018d](https://github.com/Scrivito/scrivito-portal-app/pull/701/commits/3d6018d5de294b9907b5dccb0b1a1d5dded6c385))

Good stuff!

Please see my comments for mainly nitpicks.

src/assets/stylesheets/index.scss
agessler: Yes. But it depends on the used font. Maybe you choose an serif font, then the fallback would be 'serif'.

src/assets/stylesheets/theme/fonts.scss
agessler: It was there before. I think it is about not to use a local font. 
Please see here..

https://stackoverflow.com/questions/3837249/font-face-src-local-how-to-use-the-local-font-if-the-user-already-has-it
https://www.paulirish.com/2009/bulletproof-font-face-implementation-syntax/

src/Components/customToastify.css
agessler: Sure we can replace. I just handed over the right variable.

src/assets/stylesheets/index.scss
agessler: Should work... we have to test it.

src/assets/stylesheets/index.scss
agessler: Yes.

src/assets/stylesheets/index.scss
agessler: We can delete family and weight here. 'Bolder' is a deign decision. From my point of view no.

src/assets/stylesheets/theme/fonts.scss
agessler: I would leave it as it is for now.

src/Components/customToastify.css
agessler: I am not sure if I should replace it within the tastify component. I found several lines. I would leave it as it is for now.

agessler:
@apepper https://github.com/Scrivito/scrivito-portal-app/pull/703

src/Components/customToastify.css
apepper: I just checked it. It's not working, since the default of `--toastify-font-family` is `sans-serif`. So we need to explicitly set the variable.

apepper (commented):
LGTM ([3a6e363](https://github.com/Scrivito/scrivito-portal-app/pull/701/commits/3a6e3638d0b19df582ad9e0304709dc45ec02eff))

Nice 🚀!

PR #708, opened 2024-12-09
Keep URL path if falling back to the default site

dcsaszar:
This leads to better fallbacks for outdated/legacy permalinks, e.g.,

* https://example.com/contact -> https://example.com/en/contact (instead of https://example.com/en)
* https://example.com/en-US/contact -> https://example.com/en/contact (instead of https://example.com/en)

### DIFF
--- a/src/config/scrivitoSites.ts
+++ b/src/config/scrivitoSites.ts
@@ -85,28 +85,28 @@ export function isNoSitePresent(): boolean {
 }
 
 export async function ensureSiteIsPresent() {
-  if ((await load(currentSiteId)) === null) {
-    if (await load(isNoSitePresent)) {
-      ensureUserIsLoggedIn()
-      return
-    }
+  if (await load(currentSiteId)) return
 
-    const site = await load(getPreferredSite)
-    if (!site) return
-
-    const language = languageForUrl(window.location.href)
-    const slashesCount = language ? 2 : 1
-    const prefixLength = (language?.length || 0) + slashesCount
-    const path = window.location.pathname.substring(prefixLength)
-
-    if (!path) {
-      navigateTo(site)
-      return
-    }
-
-    const url = await load(() => urlFor(site))
-    window.location.assign([url, path].join(url.endsWith('/') ? '' : '/'))
+  if (await load(isNoSitePresent)) {
+    ensureUserIsLoggedIn()
+    return
   }
+
+  const site = await load(getPreferredSite)
+  if (!site) return
+
+  const language = languageForUrl(window.location.href)
+  const slashesCount = language ? 2 : 1
+  const prefixLength = (language?.length || 0) + slashesCount
+  const path = window.location.pathname.substring(prefixLength)
+
+  if (!path) {
+    navigateTo(site)
+    return
+  }
+
+  const url = await load(() => urlFor(site))
+  window.location.assign([url, path].join(url.endsWith('/') ? '' : '/'))
 }
 
 function getPreferredSite() {
--- a/src/config/scrivitoSites.ts
+++ b/src/config/scrivitoSites.ts
@@ -5,6 +5,7 @@ import {
   getInstanceId,
   load,
   navigateTo,
+  urlFor,
 } from 'scrivito'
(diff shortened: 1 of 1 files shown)


src/config/scrivitoSites.ts
```diff
@@ -84,24 +85,40 @@ export function isNoSitePresent(): boolean {
...
+    return
+  }
+
+  const site = await load(getPreferredSite)
+  if (!site) return
+
+  const language = languageForUrl(window.location.href)
+  const slashesCount = language ? 2 : 1
+  const prefixLength = (language?.length || 0) + slashesCount
+  const path = window.location.pathname.substring(prefixLength)
```
apepper, line 101: Why do we need this logic? It looks quite complex to me.

src/config/scrivitoSites.ts
```diff
@@ -84,24 +85,40 @@ export function isNoSitePresent(): boolean {
...
+
+  if (!path) {
+    navigateTo(site)
+    return
   }
+
+  const url = await load(() => urlFor(site))
+  window.location.assign([url, path].join(url.endsWith('/') ? '' : '/'))
```
apepper, line 109: If I'm not mistaken query parameter and hashes would not be transferred to the new URL. Is this intentional?

src/config/scrivitoSites.ts
apepper: If we don't need it the logic could swap `location.origin` from `location.href` with `urlFor(site)`.

apepper (commented):
FEEDBACK ([c201e4c](https://github.com/Scrivito/scrivito-portal-app/pull/708/commits/c201e4c506f4b3bd0c627a3d027038887de33346))

Very nice 🚀.

Please see my comments to maybe simplify this PR.

src/config/scrivitoSites.ts
dcsaszar: Yes and no. I only cared about permalink-like URLs b/c the original feature request only mentioned these. Everything else works or doesn't work by chance.

src/config/scrivitoSites.ts
dcsaszar: Depends on which kinds of old permalinks we want to handle:
1. links after relaunching a single-language site
2. links to the primary language site version that had no language prefix before the relaunch
3. links to a language version that no longer exists after the relaunch

The current logic handles all cases. If we drop (3) the logic is simpler.

apepper: I would drop (3) to keep it more simple.

src/config/scrivitoSites.ts
```diff
@@ -84,24 +84,33 @@ export function isNoSitePresent(): boolean {
...
+  }
+
+  const site = await load(getPreferredSite)
+  if (!site) return
+
+  const siteUrl = await load(() => urlFor(site))
+  const { pathname, search, hash } = window.location
+  const path = pathname === '/' ? '' : pathname
+
+  window.location.assign(`${siteUrl}${path}${search}${hash}`)
```
apepper, line 101: Will `siteUrl` always contain a trailing `/`? In your prior code you checked for it to end with `/` or not.

apepper (commented):
FEEDBACK ([e9dbe0d](https://github.com/Scrivito/scrivito-portal-app/pull/708/commits/e9dbe0d7994eb0d91eb3b8752246566785d6caeb))

Really nice. I left one smaller comment.

src/config/scrivitoSites.ts
dcsaszar: This time it's the other way around: pathname will always **start** with a `/` (except if it's empty)

apepper: But with `siteUrl` ends with a `/` and `path` starts with a `/` you will have `//` in your URL, right?

dcsaszar: Yes, but how can you make `siteUrl` end with `/`?

apepper (approved):
LGTM ([e9dbe0d](https://github.com/Scrivito/scrivito-portal-app/pull/708/commits/e9dbe0d7994eb0d91eb3b8752246566785d6caeb))

PR #710, opened 2024-12-11
No longer use the term CMS in front facing UI

apepper:
The readme stays as is for now.


src/Components/NotFoundErrorPage.tsx
```diff
@@ -85,7 +85,7 @@ const GetStartedButton = connect(
         className="btn btn-primary"
         href={`https://edit.scrivito.com/${location.href}`}
       >
-        Get started with your CMS
+        Get started with your Interface Builder
```
dcsaszar, line 88: * [ ] In our planning discussion Kris suggested something like "Get started with your website" or "Get started building your website", which I find more neutral and also clearer.

PR #711, opened 2024-12-12
Custom colors: calculate lighter/darker using native CSS

apepper:
See https://github.com/Scrivito/scrivito-portal-app/pull/469#discussion_r1651037476 for initial discussion.

Since then at least Firefox added support. I'm still not 100% sure if the browser support is widespread enough (see https://caniuse.com/?search=relative%20color).

TODO:
* [x] Clear existing `siteColorPrimaryLighten` etc. attributes (if applicable). _EDIT:_ Not needed - since there is no color set in the default content.


src/Components/DesignAdjustments.tsx
```diff
@@ -10,25 +10,21 @@ export const DesignAdjustments = connect(
...
-    const secondary = root.get('siteColorSecondary')
-    if (secondary) styles.push(`--bs-secondary: ${secondary};`)
-
-    const secondaryLighten = root.get('siteColorSecondaryLighten')
-    if (secondaryLighten) {
-      styles.push(`--bs-secondary-lighten: ${secondaryLighten};`)
+    if (primary) {
+      styles.push(`--bs-primary: ${primary};`)
+      styles.push(`--bs-primary-lighten: hsl(from ${primary} h s calc(l + 8));`)
+      styles.push(`--bs-primary-darken: hsl(from ${primary} h s calc(l - 8));`)
```
dcsaszar, line 16: * [ ] We should not duplicate the CSS code here. The current reason just using the CSS rules don't work is that the CSS applies to `:root` while the styles here only apply to `body` (see line 40). I suggest to apply the styles to `:root`. E.g.:
```tsx
  <Helmet style={[{ cssText: `:root{${styles.join('')}}` }]} />
```

dcsaszar (changes requested):
Nice. One thing to improve, see comment.

dcsaszar (approved):
LGTM

PR #713, opened 2024-12-17
Restore original Vite chunk size limit

dcsaszar:
See https://v3.vitejs.dev/guide/build.html#chunking-strategy

I picked a chunking strategy where we chunk two large groups of vendor dependencies: React based and Scrivito based packages.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -26,7 +26,12 @@ export default defineConfig(({ mode }) => {
         },
         output: {
           manualChunks: {
-            react: ['react-bootstrap', 'react-helmet-async', 'react-toastify'],
+            react: [
+              'react-bootstrap',
+              'react-dropzone',
+              'react-helmet-async',
+              'react-toastify',
+            ],
             scrivito: ['scrivito-neoletter-form-widgets'],
           },
         },
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -18,12 +18,17 @@ export default defineConfig(({ mode }) => {
 
   return {
     build: {
-      chunkSizeWarningLimit: 900,
       outDir,
       rollupOptions: {
         input: {
           main: resolve(__dirname, 'index.html'),
           _scrivito_extensions: resolve(__dirname, '_scrivito_extensions.html'),
+        },
+        output: {
+          manualChunks: {
+            react: ['react-bootstrap', 'react-helmet-async', 'react-toastify'],
+            scrivito: ['scrivito-neoletter-form-widgets'],
+          },
         },
       },
     },


vite.config.ts
```diff
@@ -18,13 +18,18 @@ export default defineConfig(({ mode }) => {
...
       outDir,
       rollupOptions: {
         input: {
           main: resolve(__dirname, 'index.html'),
           _scrivito_extensions: resolve(__dirname, '_scrivito_extensions.html'),
         },
+        output: {
+          manualChunks: {
+            react: ['react-bootstrap', 'react-helmet-async', 'react-toastify'],
+            scrivito: ['scrivito-neoletter-form-widgets'],
```
apepper, line 30: FMI: Please help my understand, what this `scrivito` chunk means? My (maybe naive) understanding would be, that their is a chunk named `scrivito`, but it only contains `scrivito-neoletter-form-widgets`. But if I look at the resulting size it shows this:

```
dist/assets/scrivito-CFqY22Np.js                                     465.18 kB │ gzip: 144.80 kB
```

But `scrivito-neoletter-form-widgets` is not that big (should be ~13 kB gzipped). I assume, that it also compiles the npm package `scrivito` in there.

vite.config.ts
```diff
@@ -18,13 +18,18 @@ export default defineConfig(({ mode }) => {
...
-      chunkSizeWarningLimit: 900,
       outDir,
       rollupOptions: {
         input: {
           main: resolve(__dirname, 'index.html'),
           _scrivito_extensions: resolve(__dirname, '_scrivito_extensions.html'),
         },
+        output: {
+          manualChunks: {
+            react: ['react-bootstrap', 'react-helmet-async', 'react-toastify'],
```
apepper, line 29: `'react-dropzone'` is also a react dependency, that is used in `DataFormUploadWidget`. Please add it to the list.

vite.config.ts
```diff
@@ -18,13 +18,18 @@ export default defineConfig(({ mode }) => {
...
     build: {
-      chunkSizeWarningLimit: 900,
       outDir,
       rollupOptions: {
         input: {
           main: resolve(__dirname, 'index.html'),
           _scrivito_extensions: resolve(__dirname, '_scrivito_extensions.html'),
         },
+        output: {
+          manualChunks: {
```
apepper, line 28: Besides the newly created `react` and `scrivito` chunk there are still two js files in the end: `main-<hex>.js` and `index-<hex>.js` - since `main-<hex>.js` is quite small, can they be combined into one file?

apepper, line 28: Any metrics that point to having the same (or even better) performance with manual chunks?

apepper (commented):
FEEDBACK ([8531bb6](https://github.com/Scrivito/scrivito-portal-app/pull/713/commits/8531bb61cc5f778e6a36c261d247da6d8fd92dfe))

Nice 🚀. I added some questions for discussion.

vite.config.ts
dcsaszar: Yes, it includes dependencies. Same goes for the react-* packages, since they depend on `react` and `react-dom`. We could make it explicit, but since subdependencies are also included, it doesn't really make things clearer.

vite.config.ts
dcsaszar: `react-dropzone` is only used in editing. I'd like to avoid having editing-only dependencies in manual chunks.

vite.config.ts
dcsaszar: According to manual Lighthouse runs it's the same. Differences are all in the < 10 ms range.

vite.config.ts
apepper: No, this package is used for the regular `DataFormUploadWidget`:

https://github.com/Scrivito/scrivito-portal-app/blob/e073bfb082f2fa25c9c49a16d2b1bb436fbf9b81/src/Widgets/DataFormUploadWidget/DataFormUploadWidgetComponent.tsx#L26-L30

See it in action on https://main.scrivito-portal-app.pages.dev/en/create-ticket-f909e9691db27360.

dcsaszar: True, I wrongly assumed dropzone goes together with draggable. 😄

vite.config.ts
```diff
@@ -18,13 +18,23 @@ export default defineConfig(({ mode }) => {
 
   return {
     build: {
-      chunkSizeWarningLimit: 900,
       outDir,
       rollupOptions: {
         input: {
-          main: resolve(__dirname, 'index.html'),
+          index: resolve(__dirname, 'index.html'),
```
apepper, line 24: If I'm not mistaken, this only changes the name, but not _merge_ the two files:

```
dist/assets/index-BZn5sF-c.js                                          6.56 kB │ gzip:   2.74 kB
dist/assets/editingConfigs-WubQ1JNa.js                                26.47 kB │ gzip:   8.25 kB
dist/assets/ScrivitoBootstrapIconEditor-DTXMz-zT.js                   47.51 kB │ gzip:  15.12 kB
dist/assets/index-DOwQ58pJ.js                                        107.26 kB │ gzip:  30.25 kB
```

Compare to one previous commit:

```
dist/assets/main-Bfrvei51.js                                           6.56 kB │ gzip:   2.74 kB
dist/assets/editingConfigs-WubQ1JNa.js                                26.47 kB │ gzip:   8.25 kB
dist/assets/ScrivitoBootstrapIconEditor-DTXMz-zT.js                   47.51 kB │ gzip:  15.12 kB
dist/assets/index-DOwQ58pJ.js                                        107.26 kB │ gzip:  30.25 kB
```

apepper (commented):
FEEDBACK ([aa92913](https://github.com/Scrivito/scrivito-portal-app/pull/713/commits/aa9291341ddfe43e5d254b43bbb343f84225c97b))

Almost there! Please see comments.

vite.config.ts
```diff
@@ -18,13 +18,24 @@ export default defineConfig(({ mode }) => {
...
           _scrivito_extensions: resolve(__dirname, '_scrivito_extensions.html'),
         },
+        output: {
+          manualChunks: {
+            react: [
```
apepper, line 29: Editing no longer works 😿 . How to reproduce:

* Visit https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dave-default-chunk-size-limi.scrivito-portal-app.pages.dev/en
* Switch to edit mode.
* Open the widget properties e.g. of the Columns Widget or an icon widget

Expected:
The column/icon editor is shown and usable.

Actual:
No editor is shown and the js console prints the following:

```
Uncaught Error: Minified React error #299; visit https://reactjs.org/docs/error-decoder.html?invariant=299 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at Rg.gt.createRoot (react-Drplhdjg.js:40:55773)
    at Pt (index-BHeeQ0Eb.js:15:3960)
    at index-BHeeQ0Eb.js:15:3938
```

apepper (changes requested):
FEEDBACK ([ad8f70a](https://github.com/Scrivito/scrivito-portal-app/pull/713/commits/ad8f70aa4d046c4de9a5994045586657245d7a1f))

Looks really nice - but editing is partly broken 😿. See comments.

vite.config.ts
dcsaszar: > can they be combined into one file

No.

After checking the latest feedback I found out that this can't be done without compromises.
The reason is that `main.tsx` and `scrivitoExtensions.tsx` both share a lot of dependencies which will be in the same chunk (`index`).

* If we merge `index` and `main`, `scrivitoExtensions.tsx` (which depends on `index`) will execute the `main.tsx` code and we would have to add conditional execution there.
* Having two independent chunk hierarchies with duplicated `index` is not easily possible, see https://github.com/vitejs/vite/discussions/14728.

Thus I'd remove the last two commits and live with `main` + `index`.
Revert or force-push?

apepper: A force-push is fine with me.

dcsaszar: Done, review plz.

apepper (approved):
LGTM ([08fe47a](https://github.com/Scrivito/scrivito-portal-app/pull/713/commits/08fe47ac7859ff97155045048a9e0a773e720178))

PR #718, opened 2024-12-20
searchLocalStorageDataClasses

apepper:
This is a preparation of the upcoming CRM search.

Intended use https://github.com/Scrivito/scrivito-portal-app/compare/searchLocalStorageDataClasses...poc_crm_search . 

Example deployment: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://poc-crm-search.scrivito-portal-app.pages.dev/en/search-02fffe06aa9f09d9?_scrivito_workspace_id=p6d384dda43ac613&_scrivito_display_mode=view

### DIFF
--- a/src/Data/provideLocalStorageDataClass.ts
+++ b/src/Data/provideLocalStorageDataClass.ts
@@ -145,8 +145,12 @@ export function provideLocalStorageDataClass(
 export function searchLocalStorageDataClasses(
   search: string,
   classNames: string[],
-): Array<{ _id: string; className: string; title: string }> {
-  const results: Array<{ _id: string; className: string; title: string }> = []
+): Array<{ _id: string; className: string; rawItem: Record<string, unknown> }> {
+  const results: Array<{
+    _id: string
+    className: string
+    rawItem: Record<string, unknown>
+  }> = []
 
   const lowerCaseSearchTerm = search.toLowerCase()
   const matchesSearchTerm = (value: unknown) =>
@@ -155,12 +159,12 @@ export function searchLocalStorageDataClasses(
 
   classNames.forEach((className) => {
     Object.entries(restoreRecord(recordKeyForClassName(className))).forEach(
-      ([_id, item]) => {
-        if (Object.values(item).some(matchesSearchTerm)) {
+      ([_id, rawItem]) => {
+        if (Object.values(rawItem).some(matchesSearchTerm)) {
           results.push({
             _id,
             className,
-            title: ensureString(item.title) || ensureString(item.keyword),
+            rawItem,
           })
         }
       },
--- a/src/Data/provideLocalStorageDataClass.ts
+++ b/src/Data/provideLocalStorageDataClass.ts
@@ -10,8 +10,6 @@ interface RawDataItem {
   [key: string]: unknown
 }
 
-const recordKeys: Set<{ recordKey: string; className: string }> = new Set()
-
 export function provideLocalStorageDataClass(
   className: string,
   {
@@ -26,8 +24,7 @@ export function provideLocalStorageDataClass(
     attributes?: ReadonlyDataClassAttributes
   } = {},
 ) {
-  const recordKey = `localDataClass-${scrivitoTenantId()}-${className}`
-  recordKeys.add({ className, recordKey })
+  const recordKey = recordKeyForClassName(className)
 
   if (initialContent) initializeContent(initialContent)
 
@@ -147,6 +144,7 @@ export function provideLocalStorageDataClass(
 
 export function searchLocalStorageDataClasses(
   search: string,
+  classNames: string[],
 ): Array<{ _id: string; className: string; title: string }> {
   const results: Array<{ _id: string; className: string; title: string }> = []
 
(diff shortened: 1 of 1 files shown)


src/Data/provideLocalStorageDataClass.ts
```diff
@@ -142,6 +145,34 @@ export function provideLocalStorageDataClass(
...
+  blackListEntities: string[] = [],
+): Array<{ _id: string; entity: string; title: string }> {
+  const results: Array<{ _id: string; entity: string; title: string }> = []
+
+  const lowerCaseSearchTerm = search.toLowerCase()
+  const matchesSearchTerm = (value: unknown) =>
+    typeof value === 'string' &&
+    value.toLowerCase().includes(lowerCaseSearchTerm)
+
+  recordKeys.forEach(({ className: entity, recordKey }) => {
```
dcsaszar, line 159: * [ ] FYI we're mixing and renaming `className` and `entity`/`entities` for the same thing. Can we stick to either one or the other for better readability?

src/Data/provideLocalStorageDataClass.ts
```diff
@@ -10,6 +10,8 @@ interface RawDataItem {
   [key: string]: unknown
 }
 
+const recordKeys: Set<{ recordKey: string; className: string }> = new Set()
```
dcsaszar, line 13: * [ ] FYI Making this a Set is misleading b/c the seemingly same key, e.g. `recordKeys.add({ recordKey: "foo", className: "bar" })` will be added multiple times to the set. How about making it an Array instead?

src/Data/provideLocalStorageDataClass.ts
```diff
@@ -25,6 +27,7 @@ export function provideLocalStorageDataClass(
   } = {},
 ) {
   const recordKey = `localDataClass-${scrivitoTenantId()}-${className}`
+  recordKeys.add({ className, recordKey })
```
dcsaszar, line 30: * [ ] FYI The `recordKey` can be recomputed easily from the `className`. How about reducing redundancy and just storing the list of class names?
* [ ] FYI Taking it one step further: Looking at the intended usage the caller of `searchLocalStorageDataClasses` [knows](https://github.com/Scrivito/scrivito-portal-app/commit/b3501f232fb4b4ead42ada2787f8376f59dbdf99#diff-f6b709a5b9754aed0e22f78271f8af54bb316ee2f4a4be41e2106558e39398b4R12-R20) all class names. If this is true for all callers we could  rid of the list of keys completely by taking the list of class names as an argument.

src/Data/provideLocalStorageDataClass.ts
```diff
@@ -142,6 +145,34 @@ export function provideLocalStorageDataClass(
...
+    typeof value === 'string' &&
+    value.toLowerCase().includes(lowerCaseSearchTerm)
+
+  recordKeys.forEach(({ className: entity, recordKey }) => {
+    if (blackListEntities.includes(entity)) return
+
+    Object.entries(restoreRecord(recordKey)).forEach(([_id, item]) => {
+      if (Object.values(item).some(matchesSearchTerm)) {
+        results.push({
+          _id,
```
dcsaszar, line 165: * [ ] FYI We may end up with multiple results sharing the same `_id`, e.g., after copy-pasting to add a new local-storage data class. Not sure which implications this will have later on.

src/Data/provideLocalStorageDataClass.ts
```diff
@@ -142,6 +145,34 @@ export function provideLocalStorageDataClass(
   }
 }
 
+export function searchLocalStorageDataClasses(
+  search: string,
+  blackListEntities: string[] = [],
+): Array<{ _id: string; entity: string; title: string }> {
+  const results: Array<{ _id: string; entity: string; title: string }> = []
```
dcsaszar, line 152: * [ ] FYI For a reader with a functional programming background something like this could be more readable (actually this is what I translate the logic of the function to in my head when reading the code):

```tsx
  return entities.flatMap((entity) =>
    Object.entries(restoreRecord(recordKeyForClassName(entity)))
      .filter(([, item]) => Object.values(item).some(matchesSearchTerm))
      .map(([_id, item]) => ({
        _id,
        entity,
        title: ensureString(item.title) || ensureString(item.keyword),
      })),
  )
```

src/Data/provideLocalStorageDataClass.ts
```diff
@@ -142,6 +145,34 @@ export function provideLocalStorageDataClass(
...
+        results.push({
+          _id,
+          entity,
+          title: ensureString(item.title) || ensureString(item.keyword),
+        })
+      }
+    })
+  })
+
+  return results
```
dcsaszar, line 173: * [ ] FYI The results will be "sorted" by the runtime order of entities. Assuming an actual backend would sort by relevance, could we apply some sorting, e.g., by the number of matches per item?

dcsaszar (approved):
Looks good so far. Depending on where we're heading with this, see FYI comments.

src/Data/provideLocalStorageDataClass.ts
apepper: FYI: I'll make it a `Record`.

src/Data/provideLocalStorageDataClass.ts
apepper: Very good point. I'll ask about it in the search concept.

src/Data/provideLocalStorageDataClass.ts
apepper: I think the resulting code complexity is not worth it. And it would most likely still respond with a different sorting order than the real backend.

dcsaszar: Thinking about this again it came to me that ordering is built into data locators. The actual change is not as complex as we thought. Could be follow-up though and is just an idea:

```ts
      .map(([_id, rawItem]) => ({
        _id,
        className,
        rawItem,
        relevance: Object.values(rawItem).filter(matchesSearchTerm).length, // <-- added
      })),
```

dcsaszar (approved):

src/Data/provideLocalStorageDataClass.ts
apepper: Looking at the SDK code I can only find `relevance` in the context of menu sorting in the content browser - not for index results. Am I missing something?

dcsaszar: The idea was to include `relevance` as a property of search result data items. What I forgot is that ordering of items is not done by the SDK but only passed on to the `index` callback. So my idea doesn't work.
However, with a `sortBy` helper the complexity of sorting by relevance should still be in the one-liner area.

apepper: I don't want to invest more into a nicer localstorage search experience. This is only done, so it's easier to develop without a PisaSales server running.

The search of `provideLocalStorageDataClass` also does not do any relevance sorting.

PR #721, opened 2025-01-06
Unified delete message

apepper:
Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://unified-delete-message.scrivito-portal-app.pages.dev/en/cart-a8ede3121df9e2f6?_scrivito_workspace_id=ub67fc762303f1b1&_scrivito_display_mode=diff

Please note, that the content is only visible, if one adds some products to the cart.

How to reproduce:
* Add a product to the cart via the product page
* Remove a product from the cart via the product page
* Add the product again to the cart (via the product page)
* Switch to the cart
* Remove the product again from the cart (via the cart page)

Before:
Screenshot 2025-01-06 at 11 46 00

Now:
Screenshot 2025-01-06 at 11 45 19

Found while reviewing https://github.com/Scrivito/scrivito-portal-app/pull/720.

### DIFF
--- a/src/Objs/Product/ProductComponent.tsx
+++ b/src/Objs/Product/ProductComponent.tsx
@@ -273,7 +273,7 @@ const LOCALIZERS = {
   de: {
     cartAddedMessage: '__product__ wurde dem Warenkorb hinzugefügt.',
     cartAddLabel: 'In den Warenkorb',
-    cartRemovedMessage: '__product__ wurde aus dem Warenkorb entfernt.',
+    cartRemovedMessage: 'Artikel aus dem Warenkorb entfernt.',
     cartRemoveLabel: 'Aus dem Warenkorb entfernen',
     cartLoginLabel: 'Anmelden',
     cartUnavailableMessage:
@@ -286,7 +286,7 @@ const LOCALIZERS = {
   en: {
     cartAddedMessage: 'Added __product__ to cart.',
     cartAddLabel: 'Add to cart',
-    cartRemovedMessage: 'Removed __product__ from cart.',
+    cartRemovedMessage: 'Removed item from cart.',
     cartRemoveLabel: 'Remove from cart',
     cartLoginLabel: 'Log in',
     cartUnavailableMessage: 'Please log in to add __product__ to cart.',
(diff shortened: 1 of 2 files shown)


src/Objs/Product/ProductComponent.tsx
```diff
@@ -286,7 +286,7 @@ const LOCALIZERS = {
   en: {
     cartAddedMessage: 'Added __product__ to cart.',
     cartAddLabel: 'Add to cart',
-    cartRemovedMessage: 'Removed __product__ from cart.',
+    cartRemovedMessage: 'Removed item from cart.',
```
dcsaszar, line 289: * [ ] FYI this actually decreases usability. From the screenshots you can see how clear the add/remove messages are if the product name is included. Ideally we would always show the product name instead of never.

dcsaszar (approved):

src/Objs/Product/ProductComponent.tsx
apepper: > Ideally we would always show the product name instead of never.

But as far as I know you don't have access to the product name in the `DataDeleteButtonWidget` scenario.

apepper: I tried adding a copy of the title of the product to `CartItem`.

With that I could use the following delete message: `Removed __CartItem.title__ from cart.`. If I render this in a `Scrivito.ContentTag` all is fine. But if I try to read it via `const deletedMessage = widget.get('deletedMessage')` I don't get the resolve value, but the raw `Removed __CartItem.title__ from cart.`.

Any idea how I can get a string with the resolved value from Scrivito?

dcsaszar: This can already be done by moving the delete button into the Product data group and wrapping it by a CartItem data group. With this setup, `Removed __Product.title__ from cart.` works.

apepper: 👍 for `useResolvedHtmlValue`!

Adressed in https://github.com/Scrivito/scrivito-portal-app/pull/722.

PR #726, opened 2025-01-13
Enable template expression type check

dcsaszar:
Some of the code changes were actual issues discovered thanks to the new rule.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -78,6 +78,17 @@ export default [
           argsIgnorePattern: '^_',
         },
       ],
+
+      '@typescript-eslint/restrict-template-expressions': [
+        'error',
+        {
+          allowAny: false,
+          allowBoolean: false,
+          allowNever: false,
+          allowNullish: false,
+          allowRegExp: false,
+        },
+      ],
     },
   },
   {
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -37,6 +37,10 @@ export default [
 
     languageOptions: {
       parser: tsParser,
+      parserOptions: {
+        projectService: true,
+        tsconfigRootDir: import.meta.dirname,
+      },
     },
 
     settings: {
--- a/src/Components/ScrivitoExtensions/SocialCardsTab.tsx
+++ b/src/Components/ScrivitoExtensions/SocialCardsTab.tsx
@@ -1,3 +1,5 @@
+// TODO: remove workaround for #11474
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
 import {
   Obj,
   uiContext,
--- a/src/Data/CartItem/Cart.ts
+++ b/src/Data/CartItem/Cart.ts
@@ -76,7 +76,7 @@ export async function checkoutCart(): Promise<DataItem> {
 async function getTitle() {
   const name = await load(() => currentUser()?.name())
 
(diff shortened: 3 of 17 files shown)


src/Data/CartItem/Cart.ts
```diff
@@ -72,6 +76,8 @@ export async function checkoutCart(): Promise<DataItem> {
 async function getTitle() {
   const name = await load(() => currentUser()?.name())
 
+  if (name === undefined) return
```
apepper, line 79: If I'm not mistaken this would create an Opportunity with `undefined` for the `keyword`.

This looks more like an exceptional case: Either throw an error or provide a fallback keyword.

src/Widgets/VideoWidget/VideoWidgetComponent.tsx
```diff
@@ -22,7 +22,7 @@ provideComponent(VideoWidget, ({ widget }) => {
   return (
     <ContentTag
       tag="video"
-      key={`${src}${posterUrl}`}
+      key={`${src || ''}${posterUrl || ''}`}
```
apepper, line 25: I think the code would be improved if `urlFromBinaryObj` always returns a string - an empty string if no URL could be generated.

src/Components/ScrivitoExtensions/SocialCardsTab.tsx
```diff
@@ -1,3 +1,5 @@
+// TODO: remove workaround for #11474
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
```
apepper, line 2: Why?

eslint.config.mjs
```diff
@@ -74,6 +78,17 @@ export default [
           argsIgnorePattern: '^_',
         },
       ],
+
+      '@typescript-eslint/restrict-template-expressions': [
```
apepper, line 82: According to https://typescript-eslint.io/rules/restrict-template-expressions/ this is part of `plugin:@typescript-eslint/recommended-type-checked`. Would it be much work, to use the more general plugin?

apepper (commented):
FEEDBACK ([9907799](https://github.com/Scrivito/scrivito-portal-app/pull/726/commits/99077992f3a694ca74a36b212ac09bbe2fa3e959))

Nice 🚀! Thank you for elevating the code quality of the portal app!

Please see my comments for potential issues/improvements.

src/Components/ScrivitoExtensions/SocialCardsTab.tsx
dcsaszar: From the docs:
> The severity is a string denoting the severity of the validation result and can have one of the following values: "error", "warning" and "info".

The current SDK will return `undefined` under some circumstances (and is typed accordingly), but this will be fixed upstream.

eslint.config.mjs
dcsaszar: Both the `defaultOptionsRecommended` and `defaultOptionsStrict` are not suited for us.
* recommended: allows a lot of crazy stuff
* strict: doesn't allow numbers, which we currently accept

apepper (approved):
LGTM ([ac9c153](https://github.com/Scrivito/scrivito-portal-app/pull/726/commits/ac9c153658b051cc60b416bfd387cc14bc7d00b3))

Nice 🚀!

PR #728, opened 2025-01-28
TopNavigationWidget: Allow to configure brand link

apepper:
Defaults to `root` (as previously).

### DIFF
--- a/src/Widgets/TopNavigationWidget/SubComponents/Brand.tsx
+++ b/src/Widgets/TopNavigationWidget/SubComponents/Brand.tsx
@@ -1,17 +1,26 @@
-import { connect, InPlaceEditingOff, LinkTag, ImageTag } from 'scrivito'
+import {
+  connect,
+  InPlaceEditingOff,
+  LinkTag,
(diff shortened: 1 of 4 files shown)


src/Widgets/TopNavigationWidget/SubComponents/Brand.tsx
```diff
@@ -1,17 +1,26 @@
...
+  ImageTag,
+  Obj,
+  Link,
+} from 'scrivito'
 import { HomepageInstance } from '../../../Objs/Homepage/HomepageObjClass'
 import { alternativeTextForObj } from '../../../utils/alternativeTextForObj'
 
 export const Brand = connect(function Brand({
+  linkTo,
   root,
```
dcsaszar, line 14: * [ ] FYI it looks weird to pass in `root` when the component could ask the SDK for it, e.g.,
```tsx
  const root = Obj.root()
  if (!isHomepage(root)) return null
```

dcsaszar (approved):
LGTM

PR #729, opened 2025-01-28
Unified endpoint

apepper:
Requires Cloudflare Pages changes (see chat) (not yet applied).


src/Data/CallbackRequest/callbackRequestParamsFallback.ts
```diff
@@ -9,6 +9,9 @@ async function attributes(): Promise<DataAttributeDefinitions> {
   }
 }
 
-export function localStorageCallbackRequestDataClass() {
-  return provideLocalStorageDataClass('CallbackRequest', { attributes })
+export function callbackRequestParamsFallback() {
```
dcsaszar, line 12: * [ ] FYI Naming: `*DataClass*` was clearer than `*Params*`. Params are just a wildcard and can mean anything, while a data class is meant to be passed to `provideDataClass`. How about `*FallbackDataClass` or `*DataClassFallback`?

src/Data/Contract/ContractDataClass.ts
```diff
@@ -1,6 +1,14 @@
...
 
-export const Contract = import.meta.env.ENABLE_PISA
-  ? pisaContractDataClass()
-  : localStorageContractDataClass()
+export const Contract = provideDataClass(
+  'Contract',
+  (async () => {
+    const restApi = await pisaConfig('contract')
+    if (!restApi) {
+      return (await import('./contractParamsFallback')).contractParamsFallback()
```
dcsaszar, line 9: * [ ] FYI I see that we're trying to save by dynamic imports, but the outcome is not very promising and thus the complexity may not be worth the effort or in the wrong spot.

### Non-editing assets

kB  | PR | main
-- | -- | --
overall | 394.78 | 383.64
index | 277.4 | 298.45

Yes, we save ~22 k for the non-local-storage case.
But we end up with ~11 k more (and many requests) for the local-storage-case.

dcsaszar (approved):
LGTM. Two FYI for consideration.

PR #734, opened 2025-01-30
ProductCategory: Merge "description" into "metaDataDescription"

apepper:
Both attributes are very similar and "metaDataDescription" also ready has a title. Better only use one of the two.

Content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://merge-description-into-metad.scrivito-portal-app.pages.dev/en/products-749ee4e15707e4ff?_scrivito_workspace_id=h530cb393e8ca441&_scrivito_display_mode=diff

This is a breaking change: The app will only show the correct values on the "Products" page, if both the app and the content are "in sync" => don't just merge.

### DIFF
--- a/src/Objs/ProductCategory/ProductCategoryEditingConfig.ts
+++ b/src/Objs/ProductCategory/ProductCategoryEditingConfig.ts
@@ -13,7 +13,7 @@ provideEditingConfig(ProductCategory, {
   title: 'Product Category',
   thumbnail: Thumbnail,
   attributes: defaultPageEditingConfigAttributes,
-  properties: [...defaultPageProperties, 'description', 'image'],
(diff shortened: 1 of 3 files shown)


apepper:
Alternatively we can leave two attributes (`description` and `metaDataDescription`) for ProductCategory.

In this case we need to find a good name for `description`, since `metaDataDescription` already has the title `Page description`.

dcsaszar (approved):
According to the suggested usage of meta description [here](https://developers.google.com/search/docs/appearance/snippet#meta-descriptions) LGTM.

apepper:
Obsolete with https://github.com/Scrivito/scrivito-portal-app/pull/788.

PR #747, opened 2025-02-10
ContentSearchResult

apepper:
Moves search logic of https://github.com/Scrivito/scrivito-portal-app/blob/main/src/Widgets/SearchResultsWidget/SearchResultsWidgetComponent.tsx into a data class. Overall goal: Get rid of `SearchResultsWidget` and use data search primitives.

Working copy to test it out: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://contentsearchresult.scrivito-portal-app.pages.dev/en/search-results?q=Test&_scrivito_workspace_id=r228d708e9d6524d&_scrivito_display_mode=view

Please not that other PRs are still needed before the content can go live.

### DIFF
--- a/src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
+++ b/src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
@@ -73,8 +73,8 @@ function objToResult(obj: Obj) {
   const imageId = isImage(imageReference) ? imageReference.id() : null
 
   const snippet =
-    extractText(obj, { length: 300 }) ||
-    ensureString(obj.get('metaDataDescription'))
+    ensureString(obj.get('metaDataDescription')) ||
+    extractText(obj, { length: 300 })
 
   return {
     _id: obj.id(),
--- a/src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
+++ b/src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
@@ -48,7 +48,7 @@ export const ContentSearchResult = provideDataClass('ContentSearchResult', {
 
       const objSearch = Obj.whereFullTextOf('*', 'matches', search)
         .andNot('_objClass', 'equals', BLACKLIST_OBJ_CLASSES)
-        .and('_dataParam', 'equals', null) // Ignore data details pages
+        .and('_dataParam', 'equals', null) // Ignore data details pages. See issue #11592.
         .andNot('excludeFromSearch', 'equals', true)
 
       const limit = params.limit()
--- a/src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
+++ b/src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
@@ -70,7 +70,7 @@ export const ContentSearchResult = provideDataClass('ContentSearchResult', {
 
 function objToResult(obj: Obj) {
   const imageReference = obj.get('image')
-  const image = isImage(imageReference) ? imageReference.id() : null
+  const imageId = isImage(imageReference) ? imageReference.id() : null
 
   const snippet =
     extractText(obj, { length: 300 }) ||
@@ -78,7 +78,7 @@ function objToResult(obj: Obj) {
 
   return {
     _id: obj.id(),
-    image,
+    image: imageId,
     snippet: removePlaceholders(snippet),
     title: removePlaceholders(objTitle(obj)),
     url: urlFor(obj),
--- a/src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
+++ b/src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
@@ -31,16 +31,16 @@ export const ContentSearchResult = provideDataClass('ContentSearchResult', {
   },
   title: async () =>
     (await load(currentLanguage)) === 'de'
-      ? 'Inhalts-Suchergebnis'
-      : 'Content Search Result',
+      ? 'Suchtreffer (Content)'
+      : 'Search result (content)',
   connection: {
     async index(params) {
       if (Object.keys(params.filters()).length > 0) {
-        throw new Error('Filtering is not supported for CMS Search Results.')
+        throw new Error('Search result (content) does not support filtering.')
       }
 
       if (params.order().length > 0) {
-        throw new Error('Sorting is not supported for CMS Search Results.')
+        throw new Error('Search result (content) does not support sorting.')
       }
 
       const search = params.search()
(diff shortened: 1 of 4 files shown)


src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
```diff
@@ -0,0 +1,89 @@
...
+export const ContentSearchResult = provideDataClass('ContentSearchResult', {
+  attributes: async () => {
+    const lang = await load(currentLanguage)
+
+    return {
+      image: [
+        'reference',
+        { title: lang === 'de' ? 'Bild' : 'Image', to: 'Image' },
+      ],
+      snippet: ['string', { title: lang === 'de' ? 'Schnipsel' : 'Snippet' }],
```
dcsaszar, line 24: * [ ] Maybe more formal "Textausschnitt"?

src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
```diff
@@ -0,0 +1,89 @@
...
+      title: ['string', { title: lang === 'de' ? 'Titel' : 'Title' }],
+      url: ['string', { title: 'URL' }],
+    }
+  },
+  title: async () =>
+    (await load(currentLanguage)) === 'de'
+      ? 'Inhalts-Suchergebnis'
```
dcsaszar, line 31: * [ ] FYI To make this more readable maybe it's sufficient to have just "search results" and (later on) "data search results". Alternatively maybe sth like "Suchtreffer (Content)"?

src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
```diff
@@ -0,0 +1,89 @@
...
+      : 'Content Search Result',
+  connection: {
+    async index(params) {
+      if (Object.keys(params.filters()).length > 0) {
+        throw new Error('Filtering is not supported for CMS Search Results.')
```
dcsaszar, line 36: * [ ] FYI strange capitalization in "Search Results".

src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
```diff
@@ -0,0 +1,89 @@
...
+      if (params.order().length > 0) {
+        throw new Error('Sorting is not supported for CMS Search Results.')
+      }
+
+      const search = params.search()
+      if (!search) return { results: [], count: 0 }
+
+      const objSearch = Obj.whereFullTextOf('*', 'matches', search)
+        .andNot('_objClass', 'equals', BLACKLIST_OBJ_CLASSES)
+        .and('_dataParam', 'equals', null) // Ignore data details pages
```
dcsaszar, line 48: * [ ] Undocumented feature. Is there an issue for that? Maybe add a bit more about how/why/for how long this works to the comment? How about removing this magic and setting `excludeFromSearch: true` for these pages?

src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
```diff
@@ -0,0 +1,89 @@
...
+  },
+})
+
+function objToResult(obj: Obj) {
+  const imageReference = obj.get('image')
+  const image = isImage(imageReference) ? imageReference.id() : null
+
+  const snippet =
+    extractText(obj, { length: 300 }) ||
+    ensureString(obj.get('metaDataDescription'))
```
dcsaszar, line 74: * [ ] FMI how often/for which pages/objects is the fallback code path taken?

src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
```diff
@@ -0,0 +1,89 @@
...
+      const continuation = limit < count ? limit.toString() : undefined
+
+      return { continuation, count, results }
+    },
+  },
+})
+
+function objToResult(obj: Obj) {
+  const imageReference = obj.get('image')
+  const image = isImage(imageReference) ? imageReference.id() : null
```
dcsaszar, line 70: * [ ] FYI maybe use `imageId` here for more clarity about the data in question.

dcsaszar (changes requested):
Cool. It's obvious that we lose the live updating by this approach but I guess there currently is no way around it. An invalidation mechanism would be nice to have 😜 .

See minor feedback in comments.

src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
apepper: I don't think that there is an issue for that. It is also not mentioned in the original introduction in https://github.com/Scrivito/scrivito-portal-app/commit/0f854f830452aab04c738753e588e90c595f75f2.

src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
apepper: Currently this is only used for "overview" pages, such as:

```
e8f5370d6569478a GDPR consent
e94896f45d8b5699 Parts
ec9a22125e28ddcc Quotes archive
f7da3de531e46fe9 Quotes
fb302b23c46fea41 Events
```

src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
apepper: I'm hesitant to manually set `excludeFromSearch: true` for detail pages. This could be a trap, since the editor would have to think about this whenever he or she creates a new details page. I think it's better to filter them out by default.

apepper:
REVIEW ([9983d16](https://github.com/Scrivito/scrivito-portal-app/pull/747/commits/9983d169b648c2e596fdc925200bae95d03166d0))

src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
dcsaszar: * [ ] FYI makes sense, but looking at these cases I think it would make more sense to switch precendence. Otherwise a simple edit, e.g., an added sub-headline, can make the search result way less helpful.

```ts
ensureString(obj.get('metaDataDescription')) || extractText(obj, { length: 300 })
```

dcsaszar (approved):

src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
apepper: I created a new issue of this: https://github.com/infopark/scrivito_js/issues/11592

apepper:
REVIEW ([59bf0a6](https://github.com/Scrivito/scrivito-portal-app/pull/747/commits/59bf0a67756cad2fff73f1d8426a49517070b857))

dcsaszar (approved):

PR #752, opened 2025-02-12
Added DataCountWidget

apepper:
Inspired by parts of [SearchResultsWidget](https://github.com/Scrivito/scrivito-portal-app/blob/f25429b0584c0913bb901a80d434b38bc85f8ae7/src/Widgets/SearchResultsWidget/SearchResultsWidgetComponent.tsx#L77-L79).

Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://datasearchcountwidget.scrivito-portal-app.pages.dev/en/tickets?_scrivito_workspace_id=me5502f21472a82e&_scrivito_display_mode=diff

PR #755, opened 2025-02-13
Subscription supporting filter

apepper:
Discovered while using https://github.com/Scrivito/scrivito-portal-app/pull/752. Motivation: Show a "x active subscriptions" counter on the subscriptions page. This is only possible, if filters for subscriptions are supported.

While I was already at it, I supported most of the other options as well - or failed them on purpose.

### DIFF
--- a/src/Data/Subscription/SubscriptionDataClass.ts
+++ b/src/Data/Subscription/SubscriptionDataClass.ts
@@ -35,7 +35,10 @@ export const Subscription = provideDataClass('Subscription', {
         unfilteredSubscriptions,
       )
 
-      return { results: subscriptions, count: subscriptions.length }
+      return {
+        results: subscriptions.slice(0, params.limit()),
+        count: subscriptions.length,
+      }
     },
     async get(id: string) {
       return (await fetchSubscriptions()).find((sub) => sub._id === id) || null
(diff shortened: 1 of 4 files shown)


src/Data/filterDataItems.ts
```diff
@@ -0,0 +1,61 @@
+import { DataConnectionFilters, DataConnectionResultItem } from 'scrivito'
+
+export function filterDataItems(
```
dcsaszar, line 3: * [ ] FYI The location and name of the file looks innocent but may lead to code with bad performance. A reader may not be aware that client-side filtering is a bad idea in most cases and only makes sense for external data sets with a constant fetch effort. Maybe we can find a name that indicates the intended use better? Something like `filterDataItemsAfterFullFetch`? Maybe we could even ask the respective backend to support the filters we need?

src/Data/Subscription/SubscriptionDataClass.ts
```diff
@@ -35,7 +35,10 @@ export const Subscription = provideDataClass('Subscription', {
         unfilteredSubscriptions,
       )
 
-      return { results: subscriptions, count: subscriptions.length }
+      return {
+        results: subscriptions.slice(0, params.limit()),
```
dcsaszar: * [ ] Oops! Checking why limit is not respected already I saw that `fetchSubscriptions` doesn't properly handle `continuation`. We should fix that and handle `limit` early. However, this also means that client-side filtering is not a good idea.

dcsaszar (commented):
I stopped the review after 0a20ffd4dc02cade2bc18fdd8a0905ce22654394. Needs rethinking. See comments for why.

PR #757, opened 2025-02-14
Extract and use JWT token for pisa Authorization

apepper:


DanielO-JR (approved):
Looks good

src/main.tsx
```diff
@@ -54,3 +62,16 @@ if (isEditorLoggedIn()) {
...
 }
+
+setPisaAuthorization(authorization())
+
+function authorization(): string | null {
+  if (isUserLoggedIn()) return null
+
+  if (typeof window === 'undefined') return null
+
+  const urlParams = new URLSearchParams(window.location.search)
```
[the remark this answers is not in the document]
apepper: I think `currentPageParams()` is not available on boot time:

https://github.com/Scrivito/scrivito_sdk_js/blob/6df50547dde2e6688ada1f01f15fefb863de4a30/src/scrivito_sdk/app_support/current_page_data.ts#L60

src/main.tsx
```diff
@@ -54,3 +62,16 @@ if (isEditorLoggedIn()) {
   import('./Objs/editingConfigs')
   import('./Widgets/editingConfigs')
 }
+
+setPisaAuthorization(authorization())
+
+function authorization(): string | null {
+  if (isUserLoggedIn()) return null
```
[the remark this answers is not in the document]
apepper: Yes. This should also prevent inclear states, e.g. if someone uses a token URL during editing.

apepper:
REVIEW ([6a1cae6](https://github.com/Scrivito/scrivito-portal-app/pull/757/commits/6a1cae62fc796689c814534bb55605a662d2b43f))

src/main.tsx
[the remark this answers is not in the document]
dcsaszar: I see. To be sure:
* The token is not used to trigger a backend write immediately on a visit? I.e. no confirmation link etc?
* The target page will show who is logged in (e.g. in the nav bar) so the token visitor is made aware that their personal link is not showing their personal data because somebody else is still logged in?

src/main.tsx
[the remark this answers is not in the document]
dcsaszar: The caller already uses async data `currentLanguage()`, so it could await `currentPageParams()` as well.

dcsaszar (approved):
Nice. Only FYI/discussion left.

src/Data/pisaClient.ts
```diff
@@ -25,14 +31,37 @@ export async function pisaConfig(subPath: string) {
...
   }
 }
 
 function never() {
   return new Promise<never>(() => {})
 }
+
+let cachedAuthorization: string | null | undefined = undefined
+function getAuthorization(): string | null {
+  if (cachedAuthorization === undefined) cachedAuthorization = calculate()
```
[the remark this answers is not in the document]
apepper: I intentialy left this as an inner function to avoid bugs, where `calculateAuthorization()` is called again and again.

src/main.tsx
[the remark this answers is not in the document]
apepper: I don't know, but I find the code not to be easier to read:

```ts
let cachedAuthorization: Promise<string | null> | undefined = undefined
async function getAuthorization(): Promise<string | null> {
  if (cachedAuthorization === undefined) cachedAuthorization = calculate()

  return cachedAuthorization

  async function calculate() {
    if (isUserLoggedIn()) return null

    const pageParams = await load(() => currentPageParams())
    const token = pageParams.token

    return typeof token === 'string' && token ? `JWT ${token}` : null
  }
}
```

I would avoid this, if we can.

src/Data/pisaClient.ts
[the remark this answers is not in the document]
dcsaszar: From my point of view explanatory naming is the more common (and therefore better) way of avoiding expensive control flows.

apepper:
PRE-REVIEW ([4bf4973](https://github.com/Scrivito/scrivito-portal-app/pull/757/commits/4bf49736bc48ce4babbef3ca8ccd3693029cdedc))

I had to switch the approach due to #11616. While manually testing I noticed, that `pisaDataBinaryToUrl` is not ready yet (I'll work on it tomorrow). But I would love to get your feedback @dcsaszar already for the code so far.

apepper:
REVIEW ([be927dc](https://github.com/Scrivito/scrivito-portal-app/pull/757/commits/be927dc9b3f9c1f02e829d4e1c0fa97026cdbaae))

How to test:
* In PisaSales:
  * Log into PisaSales
  * Search for your account under "Kontakte => Personen"
  * On the very right of the screen click on the three dots and click "Einwilligungs E-Mail erzeugen": Screenshot 2025-02-26 at 14 16 18
  * In the opening dialog hover over "Yes, I would like to give my consent" and copy the link. It should look like this: `https://example.com/sales/dsgvo/consent.htm?token=eyJhb...`: Screenshot 2025-02-27 at 10 23 11
* On your local machine:
  * Start the dev server without `FORCE_LOCAL_STORAGE`.
  * In a private browser window visit `http://localhost:8080/d0a154d76edf2a7bd991fc658e700a1d/en?token=eyJhb...`
  * In the browser JS console execute:
    * `Scrivito.getDataClass('CurrentUser').all().take()[0].get('name')` to test the current user implementation.
    * `Scrivito.getDataClass('Gdpr').all().take()` to test the index function of Gdpr.
    * `Scrivito.getDataClass('Gdpr').all().take()[0].update({ active: true })` to test the update function of Gdpr.
    * You may have to rerun the commands due to `ScrivitoError: Content not yet loaded`

src/Data/Gdpr/GdprDataClass.ts
```diff
@@ -9,6 +10,77 @@ export const Gdpr = provideDataClass(
...
+          const response = await fetch(`${restApi.url}/${id}`, {
+            method: 'PATCH',
+            headers: { ...restApi.headers, 'Content-Type': 'application/json' },
+            body: JSON.stringify(data),
+          })
+          if (!response.ok) {
+            const error = new DataConnectionError(
+              'Failed to update GDPR consent',
+            )
+            errorToast('Unable to connect to PisaSales', error)
```
dcsaszar, line 62: * [ ] "Unable to connect to PisaSales" is nothing a visitor would understand. "Failed to update GDPR consent" is fine though.

src/Data/Gdpr/GdprDataClass.ts
```diff
@@ -9,6 +10,77 @@ export const Gdpr = provideDataClass(
...
+          if (continuation) urlParams.set('_continuation', continuation)
+
+          const url = `${restApi.url}?${urlParams.toString()}`
+
+          const response = await fetch(url, { headers: restApi.headers })
+          if (!response.ok) {
+            const error = new DataConnectionError(
+              'Failed to fetch GDPR consent',
+            )
+            errorToast('Unable to connect to PisaSales', error)
```
dcsaszar, line 46: * [ ] Both messages will not help the visitor understand what happened and how to proceed.

dcsaszar (changes requested):
Just quick feedback on code changes. I haven't tried it out yet b/c of time constraints.

* [ ] Q: One thing I'm still confused about is the potential mix of two identities, if Bob is logged into IAM and clicks on Alices link, wouldn't some parts of the portal (the non-Pisa-related ones) still get/write data for Bob?

src/Data/Gdpr/GdprDataClass.ts
dcsaszar: I think a good compromise would be to have both:
* An error message that helps the user understand they could contact support. "Unable to connect to PisaSales" is not good at explaining the next step. A visitor would probably go and search the web for "PisaSales" or "How can I connect to PisaSales", which would lead them nowhere.
* Error details a supporter can work with. Mentioning PisaSales is not necessarily helpful here, b/c the responsible backend is probably already known at the receiving end.

src/main.tsx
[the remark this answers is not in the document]
apepper: We settled on the following:
* If a user is logged in, the data from that user is taken. But the token is independently verified, if it belongs to the same user. If not an error toast is shown.

src/Data/CurrentUser/CurrentUserDataItem.ts
```diff
@@ -124,29 +131,53 @@ export const CurrentUser = provideDataItem('CurrentUser', {
   },
 })
 
+async function getWhoAmIOnlyUser(): Promise<{
```
dcsaszar, line 134: * [ ] FYI The naming is not helpful. I had to read the callers and implementation to understand what it does. Given the trivial nature and a single caller we could inline the code to avoid naming issues.

src/Data/CurrentUser/CurrentUserDataItem.ts
```diff
@@ -97,6 +102,8 @@ export const CurrentUser = provideDataItem('CurrentUser', {
       }
     },
     async update(params) {
+      if (getPisaAuthorization()) throw new Error('Update not supported.')
```
dcsaszar, line 105: * [ ] FYI Not sure if `DataConnectionError` is the better choice for such cases?

src/Data/CurrentUser/getWhoAmI.ts
```diff
@@ -0,0 +1,106 @@
...
+export async function verifySameWhoAmIUser(
+  email: string,
+  pisaAuthorization: string | null,
+) {
+  if (!pisaAuthorization) return
+
+  const baseUrl = await pisaUrl()
+  if (!baseUrl) return
+
+  const response = await fetch(`${baseUrl}/whoami`, {
```
dcsaszar, line 30: * [ ] FYI Should there be a TODO comment here to use the SDK helper when it supports custom auth?

src/Data/CurrentUser/getWhoAmI.ts
```diff
@@ -0,0 +1,106 @@
...
+    headers: { Authorization: pisaAuthorization },
+  })
+  if (response.status === 401) {
+    errorToast('URL token is unauthorized', new Error(''))
```
dcsaszar, line 34: * [ ] FYI The pattern of `new Error("")` is weird. How about allowing `errorToast` with one arg or using toast.info or toast.warn?

src/Data/CurrentUser/getWhoAmI.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  if (result.email === email) return
+
+  errorToast(
+    `Ignoring URL token for user ${result?.email ?? ''}.`,
+    new Error(''),
+  )
+}
+
+async function requestPisa(url: string, headers: Record<string, string>) {
```
dcsaszar, line 50: * [ ] FYI another helper that should be marked with `TODO delete me ASAP`?

src/Data/CurrentUser/getWhoAmI.ts
```diff
@@ -0,0 +1,106 @@
...
+  } = item as WhoAmI
+
+  return (
+    typeof _id === 'string' &&
+    isOptionalString(name) &&
+    isOptionalString(salutation) &&
+    isOptionalString(givenName) &&
+    isOptionalString(familyName) &&
+    isOptionalString(email) &&
+    typeof staff === 'boolean'
```
dcsaszar, line 104: * [ ] FMI Why exactly to we need these checks? Isn't the backend under our control?

dcsaszar (changes requested):
See comments. Overall the logic we apply is relatively simple, but the code looks very complicated. I see there are TODO parts but even with these removed I think we could make the behavior clearer.

Here's how I understand the logic:

If there is a token
1. and there is no user logged in, the current user is populated from a special source, token-based.
2. the GDPR endpoint uses token-based auth.
3. and the logged-in user and token-based user do not match, show a toast.

src/Data/CurrentUser/CurrentUserDataItem.ts
```diff
@@ -124,29 +131,53 @@ export const CurrentUser = provideDataItem('CurrentUser', {
...
+  name: string
+  phoneNumber: string
+  picture: string
+  pisaUserId: string
+  salesUserId: string | null
+  salutation: string
+  serviceUserId: string | null
+} | null> {
+  const whoAmI = await getWhoAmI()
+  if (!whoAmI) return null // TODO: Throw error instead?
```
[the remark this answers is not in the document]
apepper: An expired token would already throw within `getWhoAmI` => removing the TODO.

src/Data/CurrentUser/CurrentUserDataItem.ts
```diff
@@ -64,7 +65,11 @@ export const CurrentUser = provideDataItem('CurrentUser', {
   connection: {
     async get() {
       const user = await load(currentUser)
-      if (!user) return null
+      if (!user) {
+        return getPisaAuthorization() ? getWhoAmIOnlyUser() : null
+      }
+
+      verifySameWhoAmIUser(user.email(), getPisaAuthorization())
```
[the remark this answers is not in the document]
apepper: Currently the logic is like this: Errors thrown while fetching data is only shown to logged in users:

https://github.com/Scrivito/scrivito-portal-app/blob/335868d17f26289c10a9d5cfe6fff84ee48ac281/src/Components/DataErrorEditorNote.tsx#L3-L10

So non-editors will not see those errors, but empty divs. To avoid that I opted for errorToasts in the `CurrentUserDataItem`.

I'm open for suggestions on how to avoid `errorToasts` in `CurrentUserDataItem` completely. But I see this more as a follow up refactoring.

src/Data/CurrentUser/getWhoAmI.ts
apepper: They are not strictly needed, but added for completeness sake.

apepper: I removed them.

apepper:
REVIEW ([9e43c68](https://github.com/Scrivito/scrivito-portal-app/pull/757/commits/9e43c68d8c4fd6a15db62c23f68aaf1f3e5b3511))

Sorry for the back and forth with all those function. It is probably best to review the complete diff instead of reviewing one commit at a time.

src/Data/CurrentUser/verifySameWhoAmIUser.ts
```diff
@@ -0,0 +1,25 @@
...
+  })
+  if (response.status === 401) simpleErrorToast('URL token is unauthorized')
+
+  if (!response.ok) return
+
+  const result = (await response.json()) as WhoAmI
+
+  if (result.email === email) return
+
+  simpleErrorToast(`Ignoring URL token for user ${result.email ?? ''}.`)
```
dcsaszar, line 24: * [ ] FYI The message should explain the issue and how to solve it. Something like "The link you followed is for Bob but you are currently logged in as Alice/viewing Alice's settings. In order to see Bob's settings, please log out first."

src/Data/CurrentUser/verifySameWhoAmIUser.ts
```diff
@@ -0,0 +1,25 @@
...
+  email: string,
+  Authorization: string,
+) {
+  const baseUrl = await pisaUrl()
+  if (!baseUrl) return
+
+  const response = await fetch(`${baseUrl}/whoami`, {
+    headers: { Authorization },
+  })
+  if (response.status === 401) simpleErrorToast('URL token is unauthorized')
```
dcsaszar, line 16: * [ ] FYI Do we know more about the reason in order to make this more useful to the visitor? Something like "The link you followed is invalid or has expired."?

src/Data/CurrentUser/CurrentUserDataItem.ts
```diff
@@ -64,16 +68,23 @@ export const CurrentUser = provideDataItem('CurrentUser', {
   connection: {
     async get() {
       const user = await load(currentUser)
+
+      const Authorization = getTokenAuthorization()
+      if (Authorization) {
+        if (!user) return getTokenBasedCurrentUser(Authorization)
+        else verifySameWhoAmIUser(user.email(), Authorization)
```
dcsaszar, line 75: * [ ] FYI The fact that we're only checking this in the current user data item, and not in the GDPR data class, reinforces my previous feedback about an architectural issue. It would be better to move this check to the app boot process.

src/Data/CurrentUser/errorToast.tsx
```diff
@@ -22,3 +22,9 @@ export function errorToast(title: string, error: unknown): void {
     },
   )
 }
+
+export function simpleErrorToast(title: string): void {
+  console.error('simpleErrorToast', title)
```
dcsaszar, line 27: * [ ] FYI having `console.error` in the `errorToast` helpers may be unexpected for callers. The helpers look like a simple convenience for `toast.error` on the first sight. But actually they may trigger error reporting. Maybe we should rename them to `reportError` or so. Or remove the console logging.

src/Data/getTokenAuthorization.ts
```diff
@@ -0,0 +1,17 @@
+let cachedAuthorization: string | null | undefined = undefined
```
dcsaszar, line 1: * [ ] FYI `= undefined` is implicit and could be removed

src/Data/Gdpr/GdprDataClass.ts
```diff
@@ -9,6 +10,86 @@ export const Gdpr = provideDataClass(
       return (await import('./gdprParamsFallback')).gdprParamsFallback()
     }
 
+    if (!isUserLoggedIn()) {
+      const Authorization = getTokenAuthorization()
+      if (Authorization) {
+        const { url, headers: baseHeaders } = restApi
+        const headers = { ...baseHeaders, Authorization }
+
+        // TODO: Return { url, headers }, once #11616 is resolved
```
dcsaszar, line 19: * [ ] FYI: If I understand this comment correctly we would also drop the `does not support` checks. If this is correct, how about removing them right now?

dcsaszar (approved):
This version is readable and looks promising for after the TODOs.
See comments. All issues could be solved in follow-ups.

src/Data/CurrentUser/CurrentUserDataItem.ts
```diff
@@ -134,19 +155,57 @@ async function pisaIds() {
...
+    email: whoAmI.email ?? '',
+    familyName: whoAmI.familyName ?? '',
+    givenName: whoAmI.givenName ?? '',
+    image: whoAmI.image ?? null,
+    name: whoAmI.name ?? '',
+    position: whoAmI.position ?? '',
+    salesUserId: whoAmI.salesUserId ?? null,
+    salutation: whoAmI.salutation ?? '',
+    serviceUserId: whoAmI.serviceUserId ?? null,
+    staff: whoAmI.staff === true,
```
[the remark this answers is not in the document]
apepper: I would like to keep it "as consistent as possible".

src/Data/Gdpr/GdprDataClass.ts
apepper: One #11616 is available `filters`, `order` and `search` is supported by the backend. But since I don't want to implement them in the "manual fetch" way I opted for `does not support` messages.

DanielO-JR:
I have tested the branch locally and everything works fine, but I've discovered, that the name dosen't appear in the non editor environment. @apepper, maybe you can take a look on it.

Here the link, that can be used for tests locally:
http://localhost:8080/en/gdpr-consent-form?token=token=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJQaXNhU2FsZXMiLCJzdWIiOiJHRFBSIiwiYXVkIjoiUkVTVCBBUEkiLCJqdGkiOiIwQjk2NUNGNzM5QUE0MTdFQjI2NTY1RDM0RDcxMzM5NyIsImlhdCI6MTc0MTU5ODc0MywiZXhwIjoxNzY3NTE4NzQzLCJjb250YWN0IjoiOEQxNTA5QkRFNEFDNDE5NkEyNjg2MEIwMDc2NDI3OTYifQ.EHHEO1K9p1xf7NR6eiQFK4hlgOqAGhydYzKKdwrDo5t7VekfppHaZnvSvN6ghG67OuDxhG8Cm5rvPEJb6Dq-QD-vmbXqCV_tfTlDQvAQn8K_8ZYdej_9YRposktUL9zEtwdr6Hh9XvcsfZ-KFH5UT-sA5IMUrVcSh5AjypNQqntuo7S5btYD7VkXX7tWGClKaWzH-lIBiClEdUV8uZrV-btq5bsjLvelkzZZID4212tvMJpHPkDvn3tevUSCAM80OWMMSd_0IYXSTowAN5EkGPgbK5kdlFV_QtX3z0FiwFyYcaIWu0YA0Q8r7qPO0v65gsIVatpmzDN-ocshAoC0Hw

apepper:
>  have tested the branch locally and everything works fine, but I've discovered, that the name dosen't appear in the non editor environment. @apepper, maybe you can take a look on it.

Hey @DanielO-JR. Thanks for testing it again. I just stumbled over this independently of you and I'm working on a fix.

dcsaszar (approved):
Code LGTM. Since both of you have tried out the functionality I only looked at the code.

DanielO-JR:
Name appears now in an incognito window and otherwise everything works as expected. Thanks!

PR #762, opened 2025-02-19
Image widget with selectable object-fit

dcsaszar:
Potential follow-ups:
* [x] Disable the object-fit editing if width and height are unset
* [ ] Improve ARIA of the width/height inputs


src/Widgets/ImageWidget/DimensionsEditor.scss
```diff
@@ -84,12 +84,84 @@
...
+      border-left: 30px solid #f0f0f0;
+      box-shadow: 0 0 4px 1px rgba(0, 0, 0, 0.1);
+      padding: 4px 8px;
+      line-height: 15px;
+      display: flex;
+      flex-flow: column;
+      background: #fafafa;
+      color: #555;
+      font-size: 13px;
+      font-weight: normal;
```
apepper, line 113: FYI: Sorting attributes would probably help with the overview.

src/Widgets/ImageWidget/DimensionsEditor.scss
```diff
@@ -84,12 +84,84 @@
...
+    }
+
+    button:not(:disabled):hover {
+      background: #eee;
+      cursor: pointer;
+    }
+
+    button.enum_attribute_active {
+      border-left: 30px solid var(--jr-blue) !important;
+      box-shadow: 0 0 0 2px var(--jr-blue) inset;
```
apepper, line 143: Where is `var(--jr-blue)` defined? I can only see `var(--jr-blue-light)` in `scrivitoExtensions.scss`.

Answering my own question: In `scrivito_editing.css`.

I'm not sure, if we want to rely on css-variables from the SDK.

src/Widgets/ImageWidget/DimensionsEditor.scss
```diff
@@ -111,5 +183,35 @@
...
+      button {
+        border-left: 30px solid rgba(0, 0, 0, 0.1);
+        background: rgba(222, 222, 222, 0.2);
+        color: #eee;
+        span {
+          color: #eee;
+        }
+      }
+
+      button.enum_attribute_active {
```
apepper, line 197: FYI: With `&.enum_attribute_active` this could be inlined into the `button {` section (as done below for `:disabled` as well).

src/Widgets/ImageWidget/DimensionsEditor.tsx
```diff
@@ -97,3 +98,49 @@ const DimensionEditor = connect(function DimensionEditor({
     if (value !== '') widget.update({ [attribute]: `${value}${newUnit}` })
   }
 })
+
+const ObjectFit = connect(function ObjectFit({
+  widget,
+}: {
+  widget: ImageWidgetInstance
+}) {
+  const readOnly = !canEdit(widget.obj()) || isComparisonActive()
```
apepper, line 107: `readOnly` now gets calculated twice. I suggest to move it one component up and pass it on as a prop.

src/Widgets/ImageWidget/DimensionsEditor.tsx
```diff
@@ -97,3 +98,49 @@ const DimensionEditor = connect(function DimensionEditor({
...
+            <span>Contain</span>
+          </button>
+          <button
+            aria-current={objectFit === 'cover'}
+            className={
+              objectFit === 'cover' ? 'enum_attribute_active' : undefined
+            }
+            disabled={readOnly}
+            onClick={() => widget.update({ objectFit: 'cover' })}
+            title="The image is sized to maintain its aspect ratio while filling the element’s entire content box. The image will be clipped to fit."
```
apepper, line 136: This reads quite technical. How about this?

```
The image is scaled to fit within the element’s content box.
```

And

```
The image fills the entire content box, cropping if necessary.
```

apepper (approved):
LGTM ([4e54c50](https://github.com/Scrivito/scrivito-portal-app/pull/762/commits/4e54c50dd00bfdd533ddf45d8a034771d65e0497))

Nice 🚀! I left some improvement suggestions, but nothing blocking.

src/Widgets/ImageWidget/DimensionsEditor.tsx
```diff
@@ -8,23 +8,33 @@ export function DimensionsEditor({ widget }: { widget: ImageWidgetInstance }) {
...
           <DimensionEditor
             widget={widget}
             attribute="height"
             label="Height"
             pxOnly
+            readOnly={readOnly}
           />
         </div>
       </div>
+      <ObjectFit widget={widget} readOnly={readOnly || muteObjectFit} />
```
apepper, line 37: Please work on this in a follow up:

* I think it would be better, if `ObjectFit` always shows "Contains", regardless of the `objectFit` value, if no height is provided
* The widget rednering component should not add a `object-fit` property, if `height` is empty.

apepper (approved):
LGTM ([3e47214](https://github.com/Scrivito/scrivito-portal-app/pull/762/commits/3e47214f6b5a19bfa9b7b93e01c8c372512a0f7e))

Nice 🚀 - one follow up left.

src/Widgets/ImageWidget/DimensionsEditor.tsx
dcsaszar: > I think it would be better, if ObjectFit always shows "Contains", regardless of the objectFit value, if no height is provided

This would be inconsistent with other conditionally disabled controls, e.g.:
image

Since without height the behavior is the same regardless of the object-fit setting, it has even less potential for confusion.

However, with object-fit being last in the tab and having no meaning w/o height, how about hiding it completely in this case?

PR #764, opened 2025-02-19
Only apply cover if height is set

dcsaszar:
See https://github.com/Scrivito/scrivito-portal-app/pull/762#discussion_r1961704769


apepper (approved):
LGTM ([5ed7b37](https://github.com/Scrivito/scrivito-portal-app/pull/764/commits/5ed7b374343ff76ebb8020f45a3d6a3edba599a0))

I added the same code to `DataImageWidget` as well.

PR #766, opened 2025-02-20
Remove SearchResultsWidget and Highlighter

apepper:
⚠️ ⚠️ Do not merge before the release of v6 ⚠️ ⚠️ 

Otherwise existing instances (e.g. demo5 or play tenants) with old content will no longer work properly.


dcsaszar (approved):
Nice. Looking forward to reintroduce highlighting at some point.

PR #768, opened 2025-02-21
Rerender <source> elements on change

### LINKED ISSUE #767: Video background sometimes not working correctly
### How to reproduce

* Visit the Homepage in edit mode
* Open the widget properties of the slider widget
* Open the first "Slide"
* Click on "Background image or video", select a video and click "select"

*Expected*
The selected video plays in the background

*Actual*
No video is shown - just the blank background color

### Background
After some debugging it looks like, that the full video URL is never loaded. It seams, that the video player will not try to load a different video URL if a previous one (the loading placeholder) was set.

apepper:
Fixes #767.

As far as I can see, if one changes the `src` of the `<source>` dom element, the browser (at least Chromium) will ignore this `<source>` and pretend that it does not exist.

With the explicit `key` we tell React to recreate a new `<video>` dom element, if the content URL or it type changes.

It is also a better solution for #4767 (in Scrivito)

### DIFF
--- a/src/Widgets/SectionWidget/SectionWidgetComponent.tsx
+++ b/src/Widgets/SectionWidget/SectionWidgetComponent.tsx
@@ -50,23 +50,46 @@ const ImageOrVideo = connect(function ImageOrVideo({
   const classNames = ['img-background']
   if (widget.get('backgroundAnimateOnHover')) classNames.push('img-zoom')
 
-  return (
-    // Check is a working around for issue #4767
-    // TODO: remove work around
-    background.contentType().startsWith('video/') &&
-      background.contentUrl().startsWith('https://') ? (
-      <video className={classNames.join(' ')} autoPlay loop muted playsInline>
-        <source src={background.contentUrl()} type={background.contentType()} />
-      </video>
-    ) : (
-      <InPlaceEditingOff>
-        <ImageTag
-          content={widget}
-          attribute="backgroundImage"
-          className={classNames.join(' ')}
-          alt=""
-        />
-      </InPlaceEditingOff>
+  if (background.contentType().startsWith('video/')) {
+    return (
+      <Video
+        className={classNames.join(' ')}
+        contentUrl={background.contentUrl()}
+        contentType={background.contentType()}
+        key={[
+          'SectionWidget',
+          widget.id(),
+          'Video',
+          background.contentUrl(),
+          background.contentType(),
+        ].join('-')} // Ensure the video is reloaded when the content changes
+      />
     )
+  }
+  return (
+    <InPlaceEditingOff>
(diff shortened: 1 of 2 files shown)


src/Widgets/SectionWidget/SectionWidgetComponent.tsx
```diff
@@ -50,23 +50,46 @@ const ImageOrVideo = connect(function ImageOrVideo({
...
+        contentUrl={background.contentUrl()}
+        contentType={background.contentType()}
+        key={[
+          'SectionWidget',
+          widget.id(),
+          'Video',
+          background.contentUrl(),
+          background.contentType(),
+        ].join('-')} // Ensure the video is reloaded when the content changes
+      />
```
dcsaszar, line 66: In the VideoWidgetComponent we use a more straightforward approach:
```tsx
key={`${src}${posterUrl}`}
```
Here we don't even have a `posterUrl`, and `contentType` will usually only change if `src` also changes, so how about:
```tsx
<video className={className} autoPlay loop muted playsInline key={src}>
  <source src={src} type={contentType} />
</video>
```
* [ ] Try to simplify

src/Widgets/SectionWidget/SectionWidgetComponent.tsx
```diff
@@ -50,23 +50,46 @@ const ImageOrVideo = connect(function ImageOrVideo({
...
+        className={classNames.join(' ')}
+        contentUrl={background.contentUrl()}
+        contentType={background.contentType()}
+        key={[
+          'SectionWidget',
+          widget.id(),
+          'Video',
+          background.contentUrl(),
+          background.contentType(),
+        ].join('-')} // Ensure the video is reloaded when the content changes
```
dcsaszar, line 65: * [ ] FYI having the forced redraw in the container instead of the `<video>` which is actually the reason for the workaround obfuscates the logic.

dcsaszar (changes requested):
See comments for simplification.

src/Widgets/SectionWidget/SectionWidgetComponent.tsx
apepper: I just clarified my own confusion. So far I was not sure, if a `key` is global in react or not. But they are _not_ global:

> ### Note
>
> Remember that keys are not globally unique. They only specify the position _within the parent_.

Source: https://react.dev/learn/preserving-and-resetting-state#option-2-resetting-state-with-a-key

apepper:
REVIEW ([2ed9c48](https://github.com/Scrivito/scrivito-portal-app/pull/768/commits/2ed9c489aa737d4193eec486c33d5052275a988b))

PR #769, opened 2025-02-24
SliderWidget: Added autoplayInterval option

apepper:
Previously it was hardcoded to 5 seconds.

Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://slider-duration.scrivito-portal-app.pages.dev/en?_scrivito_workspace_id=ud8f8d872a34c34e&_scrivito_display_mode=diff

### DIFF
--- a/src/Widgets/SliderWidget/SliderWidgetClass.ts
+++ b/src/Widgets/SliderWidget/SliderWidgetClass.ts
@@ -3,6 +3,7 @@ import { provideWidgetClass } from 'scrivito'
 export const SliderWidget = provideWidgetClass('SliderWidget', {
   attributes: {
     autoplay: 'boolean',
+    autoplayInterval: 'integer',
     controls: 'boolean',
     margin: [
       'enum',
--- a/src/Widgets/SliderWidget/SliderWidgetClass.ts
+++ b/src/Widgets/SliderWidget/SliderWidgetClass.ts
@@ -3,7 +3,7 @@ import { provideWidgetClass } from 'scrivito'
 export const SliderWidget = provideWidgetClass('SliderWidget', {
   attributes: {
     autoplay: 'boolean',
-    autoplayInterval: 'integer',
+    autoplayInterval: 'float',
     controls: 'boolean',
     margin: [
       'enum',
--- a/src/Widgets/SliderWidget/SliderWidgetComponent.tsx
+++ b/src/Widgets/SliderWidget/SliderWidgetComponent.tsx
@@ -15,7 +15,9 @@ import './SliderWidget.scss'
 
 provideComponent(SliderWidget, ({ widget }) => {
   const showControls = widget.get('autoplay') ? widget.get('controls') : true
-  const autoplayInterval = (widget.get('autoplayInterval') ?? 5) * 1000
+  const autoplayInterval = Math.round(
+    (widget.get('autoplayInterval') ?? 5) * 1000,
+  )
 
   return (
     <Carousel
--- a/src/Widgets/SliderWidget/SliderWidgetComponent.tsx
+++ b/src/Widgets/SliderWidget/SliderWidgetComponent.tsx
@@ -15,16 +15,14 @@ import './SliderWidget.scss'
 
 provideComponent(SliderWidget, ({ widget }) => {
   const showControls = widget.get('autoplay') ? widget.get('controls') : true
-  const autoplayInterval = Math.round(
-    (widget.get('autoplayInterval') ?? 5) * 1000,
-  )
+  const intervalMs = Math.round((widget.get('autoplayInterval') ?? 5) * 1000)
 
   return (
     <Carousel
       className={`slider-widget ${widget.get('margin') || 'mb-4'}`}
       controls={showControls}
       indicators={showControls}
-      interval={widget.get('autoplay') ? autoplayInterval : null}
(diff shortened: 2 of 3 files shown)


src/Widgets/SliderWidget/SliderWidgetEditingConfig.ts
```diff
@@ -11,6 +11,10 @@ provideEditingConfig(SliderWidget, {
       title: 'Autoplay?',
       description: 'Default: No',
     },
+    autoplayInterval: {
+      title: 'Autoplay interval (in s)',
```
dcsaszar, line 15: * [ ] FYI we could write out "seconds" here for more clarity
* [ ] FYI Maybe there's a more editor-friendly wording instead of the technical term "autoplay interval"? "Time between slides"?

src/Widgets/SliderWidget/SliderWidgetEditingConfig.ts
```diff
@@ -32,10 +36,12 @@ provideEditingConfig(SliderWidget, {
     'slides',
     'autoplay',
     ['controls', { enabled: widget.get('autoplay') }],
+    ['autoplayInterval', { enabled: widget.get('autoplay') }],
```
dcsaszar, line 39: * [ ] FYI We could move the three autoplay options ("show controls" effectively is an autoplay option, too) to a dedicated properties group.
* [ ] FYI With two disabled options the UI looks a bit stuffed. If autoplay is turned off often: Maybe we should hide the controls and interval options

src/Widgets/SliderWidget/SliderWidgetClass.ts
```diff
@@ -3,6 +3,7 @@ import { provideWidgetClass } from 'scrivito'
 export const SliderWidget = provideWidgetClass('SliderWidget', {
   attributes: {
     autoplay: 'boolean',
+    autoplayInterval: 'integer',
```
dcsaszar, line 6: * [ ] FYI I don't see why we should not allow float. I guess if the [recommendation](https://ux.stackexchange.com/a/107400) says 5 - 10, it could also be 7.5. Through testing marketing people may arrive at a non-integer sweet spot.

dcsaszar (approved):
Nice addition. See FYI for suggestions.

src/Widgets/SliderWidget/SliderWidgetClass.ts
apepper: I chose seconds as a unit intentionally. The current used library supports milliseconds, but we might have to switch the used library at some point. Finding a slider library that supports at least seconds will be easier to do, instead of milliseconds.

dcsaszar: I'm not saying we should support milliseconds, but fractional seconds.

apepper: > [...], but fractional seconds.

How would you enforce this? If you take a float of seconds, one could set 0.001 seconds, which is identical to setting it to 1 millisecond. But float would not stop there but "allow" 0.0001 seconds or smaller - which would have no effect.

dcsaszar: > which would have no effect.

It would be rounded to 0 ms, which in turn will lead to a delay close to 0 ms, which is exactly what the user specified. But it's garbage input.
I'm not talking about garbage, but realistic input like 7.5 seconds.

apepper:
REVIEW ([8304b41](https://github.com/Scrivito/scrivito-portal-app/pull/769/commits/8304b41963f459aa9f6141541214ff3825e68fdb))

src/Widgets/SliderWidget/SliderWidgetEditingConfig.ts
```diff
@@ -12,8 +12,8 @@ provideEditingConfig(SliderWidget, {
       description: 'Default: No',
     },
     autoplayInterval: {
-      title: 'Autoplay interval (in s)',
-      description: 'Default: 5 seconds',
+      title: 'Time between slides (in seconds)',
+      description: 'Default: 5.0 seconds',
```
dcsaszar, line 16: * [ ] FYI "Default: 5 seconds" was better. No unnecessary accuracy and matches what the input would show for 5.0.

src/Widgets/SliderWidget/SliderWidgetComponent.tsx
```diff
@@ -15,7 +15,9 @@ import './SliderWidget.scss'
 
 provideComponent(SliderWidget, ({ widget }) => {
   const showControls = widget.get('autoplay') ? widget.get('controls') : true
-  const autoplayInterval = (widget.get('autoplayInterval') ?? 5) * 1000
+  const autoplayInterval = Math.round(
```
dcsaszar, line 18: * [ ] FYI renaming the variable from the attribute can make sense here since the value is something else. Maybe `intervalMs` or so?

dcsaszar (approved):
Nice. Some more FYI

apepper:
REVIEW ([141a2f5](https://github.com/Scrivito/scrivito-portal-app/pull/769/commits/141a2f5e1b2926380ce048b3958a42d69eb902f8))

dcsaszar (approved):
👍 📹

PR #772, opened 2025-02-28
Dropdown: Show 404 instead of redirect

apepper:
Only for regular visitors, not editors.

Motivation: Don't train users that the auto-redirect URLs work.


dcsaszar (approved):
Nice.
* [x] FYI Now that bots no longer see the soft redirect, maybe set Dropdowns to `noindex` automatically or by default (`initialContent`), in order to prevent accidentally published permalinks from being added to the index?

PR #778, opened 2025-03-03
DataFormBooleanWidget: Replace label with on/off label

apepper:
Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://data-form-boolean-widget-on.scrivito-portal-app.pages.dev/en/subscriptions?_scrivito_workspace_id=vafd51492ab75da9&_scrivito_display_mode=diff

Requested in https://infopark.slack.com/archives/C05FSSUAKUP/p1740737989472949?thread_ts=1738677774.437709&cid=C05FSSUAKUP.


dcsaszar (approved):
Nice. Regarding content:

* [x] I'd prefer "Not subscribed" (as in German) because a channel can only be unsubscribed by an active user choice, which is not necessarily the case here.

PR #783, opened 2025-03-07
HomepageFooterWidget

apepper:
Goal: Replace the the showAsLandingPage option in TopNavigationWidget (done in another PR).


src/Objs/defaultPageAttributes.ts
```diff
@@ -4,7 +4,7 @@ export const defaultPageAttributes = {
   // Layout widgets
   layoutHeader: 'widgetlist',
   layoutShowHeader: 'boolean',
-  layoutFooter: ['widgetlist', { only: 'SectionWidget' }],
+  layoutFooter: ['widgetlist', { only: ['FooterWidget', 'SectionWidget'] }],
```
dcsaszar, line 7: * [ ] Wouldn't a FooterWidget in the layoutFooter (of the homepage) lead to an endless recursion?
* [ ] FYI I'm not sure about the intended use. From what I can see, the new widget only makes sense on a standalone page with no path. For a standalone page, the layoutFooter makes not much sense, as putting a footer at the end of the main content has the same effect. If my guess about the intended use is correct, we could repurpose the existing `layoutShowFooter` to show the homepage footer instead of introducing a widget for that.

src/Widgets/FooterWidget/FooterWidgetComponent.tsx
```diff
@@ -0,0 +1,15 @@
+import { ContentTag, InPlaceEditingOff, Obj, provideComponent } from 'scrivito'
+import { FooterWidget } from './FooterWidgetClass'
+import { isHomepage } from '../../Objs/Homepage/HomepageObjClass'
+
+provideComponent(FooterWidget, () => {
+  const root = Obj.root()
+  if (!isHomepage(root)) return null
+  if (!root.get('layoutShowFooter')) return null
```
dcsaszar, line 8: * [ ] FYI From just reading the descriptions in the UI an editor could assume that the footer widget also works if `layoutShowFooter` is off on the homepage.

src/Widgets/FooterWidget/FooterWidgetEditingConfig.ts
```diff
@@ -0,0 +1,9 @@
+import { provideEditingConfig } from 'scrivito'
+import { FooterWidget } from './FooterWidgetClass'
+import Thumbnail from './thumbnail.svg'
+
+provideEditingConfig(FooterWidget, {
+  description: 'Shows the footer of the homepage.',
+  thumbnail: Thumbnail,
+  title: 'Footer',
```
dcsaszar, line 8: * [ ] FYI Maybe "Homepage Footer" would make the purpose clearer and we could drop the description.

dcsaszar (changes requested):
Needs some clarification. See comments.

apepper:
REVIEW ([ae49c70](https://github.com/Scrivito/scrivito-portal-app/pull/783/commits/ae49c708a314562a7d28ecd96950fe50104d7a86))

dcsaszar (approved):

PR #788, opened 2025-03-11
metaDataDescription => description

apepper:
`metaDataDescription` was originally introduced in https://github.com/Scrivito/scrivito_example_app_js/commit/a9f29d18b011be82630bd8c7e55ebb18f517f370#diff-6f02b4692074b5c3984b8d01f9bc00693e47e6185df112d8e5524dcc6e8420a9R3. But the name is quite old and can be simplified. Requested by Kris. Since there are not many places, that use `metaDataDescription` I think this is a good request.

Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://rip-meta-data-description.scrivito-portal-app.pages.dev/en/my-profile?_scrivito_workspace_id=ubbd3298f18c6b0f&_scrivito_display_mode=diff

### DIFF
--- a/src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
+++ b/src/Data/ContentSearchResult/ContentSearchResultDataClass.ts
@@ -73,8 +73,7 @@ function objToResult(obj: Obj) {
   const imageId = isImage(imageReference) ? imageReference.id() : null
 
   const snippet =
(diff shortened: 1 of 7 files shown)


dcsaszar (approved):
Very nice.

PR #794, opened 2025-03-18
Verify content format

apepper:
Concept: https://docs.google.com/document/d/1qF_jeawRCV4wVeE0f-E3A9jeGRX6AKvX7GhG53qiFys/

Also provides new JrPlatform flag and initial content URL.

### DIFF
--- a/src/privateJrPlatform/getJrPlatformApp.ts
+++ b/src/privateJrPlatform/getJrPlatformApp.ts
@@ -29,6 +29,8 @@ async function isValidContentFormat(): Promise<boolean> {
   const redirectOrigin = KNOWN_CONTENT_FORMATS[siteContentFormat]
   if (!redirectOrigin) return false
 
+  if (location.origin === redirectOrigin) return false
+
   const url = location.href.replace(location.origin, redirectOrigin)
   location.replace(url)
 
--- a/src/privateJrPlatform/getJrPlatformApp.ts
+++ b/src/privateJrPlatform/getJrPlatformApp.ts
@@ -1,0 +1,40 @@
+import { load, Obj } from 'scrivito'
+import { App } from '../App'
+import { WrongContentFormat } from './Components/WrongContentFormat'
+
+export async function getJrPlatformApp() {
+  return (await isValidContentFormat()) ? App : WrongContentFormat
+}
+
+const CONTENT_FORMAT = 'portal-app:5'
+const KNOWN_CONTENT_FORMATS: Record<string, string> = {
+  'portal-app:5': 'https://scrivito-portal-app.pages.dev',
+
+  /* Once v6 is released, the list of KNOWN_CONTENT_FORMATS looks like this: */
+  // 'portal-app:5': 'https://v5.scrivito-portal-app.pages.dev',
+  // 'portal-app:6': 'https://scrivito-portal-app.pages.dev',
+}
+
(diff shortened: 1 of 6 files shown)


src/privateJrPlatform/getJrPlatformApp.ts
```diff
@@ -0,0 +1,40 @@
...
+  const siteContentFormat = root.get('contentFormat')
+  if (siteContentFormat === CONTENT_FORMAT) return true
+
+  if (typeof siteContentFormat !== 'string') return false
+  if (!siteContentFormat) return false
+
+  const redirectOrigin = KNOWN_CONTENT_FORMATS[siteContentFormat]
+  if (!redirectOrigin) return false
+
+  const url = location.href.replace(location.origin, redirectOrigin)
```
dcsaszar, line 32: * [ ] FYI in local testing this can lead to an endless redirect loop. It would be great if this case (redirect to same origin?) would show the `WrongContentFormat` instead. What I did:


* Set the content to v6
* Added a v6 redirect to http://v6.localhost:8080
* The redirect from http://localhost:8080 worked as expected but
* Since I left the app on v5, which was required for the test, there was an endless loop

vite.config.ts
```diff
@@ -15,6 +15,7 @@ export default defineConfig(({ mode }) => {
   const outDir = 'dist'
 
   const forceLocalStorage = env.FORCE_LOCAL_STORAGE === 'true'
+  const privateJrPlatform = env.PRIVATE_JR_PLATFORM === 'true'
```
dcsaszar, line 18: * [ ] FYI I naively tried with `PRIVATE_JR_PLATFORM=1` and wondered why it didn't work 🙈

src/privateJrPlatform/getJrPlatformApp.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+const CONTENT_FORMAT = 'portal-app:5'
+const KNOWN_CONTENT_FORMATS: Record<string, string> = {
+  'portal-app:5': 'https://scrivito-portal-app.pages.dev',
+
+  /* Once v6 is released, the list of KNOWN_CONTENT_FORMATS looks like this: */
+  // 'portal-app:5': 'https://v5.scrivito-portal-app.pages.dev',
+  // 'portal-app:6': 'https://scrivito-portal-app.pages.dev',
+}
```
dcsaszar, line 16: * [ ] FYI From the comment it's not 100% clear if a v5 app needs an update when/after/before v6 comes out. Maybe the comment can make clearer, which versions need which updates on a new content version.

dcsaszar (approved):
Only FYI in comments.

src/privateJrPlatform/getJrPlatformApp.ts
apepper: `KNOWN_CONTENT_FORMATS` should be identical between `v5` and `v6`.

When `v6` get's released for the first time it can be deployed first and `v5` soon after.

apepper:
REVIEW ([ebb5981](https://github.com/Scrivito/scrivito-portal-app/pull/794/commits/ebb5981e7b2949862575d75b64af5a0b89d50a19))

dcsaszar (approved):

apepper:
FYI: Added `PRIVATE_JR_PLATFORM` with `true` in Cloudflare.

PR #800, opened 2025-03-19
JR Platform: Extract multi tenancy

apepper:
+ Add alias lookup.

One goal of this PR: Incoporated https://github.com/Scrivito/scrivito-portal-app/compare/main...v5 into `main`.

### DIFF
--- a/src/config/scrivito.ts
+++ b/src/config/scrivito.ts
@@ -4,23 +4,26 @@ import { getJrPlatformConfig } from '../privateJrPlatform/getJrPlatformConfig'
 import { getJrPlatformInstanceId } from '../privateJrPlatform/multiTenancy'
 
 export function configureScrivito(options?: { priority?: 'background' }) {
+  const tenant = scrivitoTenantId()
+  if (!tenant) return // Other layers (like vite.config.ts) should have thrown an error.
+
   configure({
     activateDataIntegration: true,
     adoptUi: true,
     autoConvertAttributes: true,
     baseUrlForSite,
     contentTagsForEmptyAttributes: false,
-    extensionsUrl: `/_scrivito_extensions.html?tenantId=${scrivitoTenantId()}`,
+    extensionsUrl: `/_scrivito_extensions.html?tenantId=${tenant}`,
     optimizedWidgetLoading: true,
     siteForUrl,
     strictSearchOperators: true,
-    tenant: scrivitoTenantId(),
+    tenant,
     ...(import.meta.env.PRIVATE_JR_PLATFORM ? getJrPlatformConfig() : {}),
     ...options,
   })
 }
 
-function scrivitoTenantId(): string {
+function scrivitoTenantId(): string | null {
   if (import.meta.env.PRIVATE_JR_PLATFORM) return getJrPlatformInstanceId()
 
   return import.meta.env.SCRIVITO_TENANT
--- a/src/config/scrivito.ts
+++ b/src/config/scrivito.ts
@@ -1,7 +1,7 @@
 import { configure } from 'scrivito'
 import { baseUrlForSite, siteForUrl } from './scrivitoSites'
-import { scrivitoTenantId } from './scrivitoTenants'
 import { getJrPlatformConfig } from '../privateJrPlatform/getJrPlatformConfig'
+import { getJrPlatformInstanceId } from '../privateJrPlatform/multiTenancy'
 
 export function configureScrivito(options?: { priority?: 'background' }) {
   configure({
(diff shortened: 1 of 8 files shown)


src/privateJrPlatform/multiTenancy.ts
```diff
@@ -0,0 +1,37 @@
...
+  const tenantFromQuery = new URLSearchParams(location.search).get('tenantId')
+  const tenant = tenantFromUrl || tenantFromQuery
+
+  if (!tenant) throw new Error('Could not determine tenant!')
+
+  return tenant
+}
+
+export function getJrPlatformBaseAppUrl(): string {
+  if (!origin) throw new Error('No origin defined!')
```
dcsaszar, line 27: * [ ] `PRIVATE_JR_PLATFORM=true npm run prerender` fails here b/c of the missing `window !== 'undefined'` fallback to SCRIVITO_ORIGIN.

src/privateJrPlatform/getJrPlatformApp.ts
```diff
@@ -23,6 +24,9 @@ async function isValidContentFormat(): Promise<boolean> {
   const siteContentFormat = root.get('contentFormat')
   if (siteContentFormat === CONTENT_FORMAT) return true
 
+  // For alias instances it's better to show a half-broken app, then to show a full error page.
```
dcsaszar, line 30: * [ ] FYI since this behavior changes the documented behavior of the function it would be better to move the comment to the function docs.

src/privateJrPlatform/getJrPlatformApp.ts
```diff
@@ -23,6 +24,9 @@ async function isValidContentFormat(): Promise<boolean> {
...
+  if (instanceFromHostname()) return true
```
dcsaszar, line 31: * [ ] FYI Could be moved to the top of the function for more clarity

vite.config.ts
```diff
@@ -17,6 +17,14 @@ export default defineConfig(({ mode }) => {
   const forceLocalStorage = env.FORCE_LOCAL_STORAGE === 'true'
   const privateJrPlatform = env.PRIVATE_JR_PLATFORM === 'true'
 
+  if (scrivitoTenantIsMissing(env)) {
+    throw new Error(
+      'Environment variable "SCRIVITO_TENANT" is not defined!' +
+        ' Check if the ".env" file with a proper SCRIVITO_TENANT is set.' +
+        ' See ".env.example" for an example.',
+    )
```
dcsaszar, line 25: * [ ] FYI The error message is not accurate given the implementation.

vite.config.ts
```diff
@@ -71,6 +79,12 @@ export default defineConfig(({ mode }) => {
   }
 })
 
+function scrivitoTenantIsMissing(env: Record<string, string>): boolean {
```
dcsaszar, line 82: * [ ] FYI Splitting the check and the error between the top and bottom of the module is hard to read. We could use an assert function that throws by itself.

src/config/scrivito.ts
```diff
@@ -4,23 +4,26 @@ import { getJrPlatformConfig } from '../privateJrPlatform/getJrPlatformConfig'
 import { getJrPlatformInstanceId } from '../privateJrPlatform/multiTenancy'
 
 export function configureScrivito(options?: { priority?: 'background' }) {
+  const tenant = scrivitoTenantId()
+  if (!tenant) return // Other layers (like vite.config.ts) should have thrown an error.
```
dcsaszar, line 8: * [ ] FYI The comment couples this implementation with other implementations ("layers") for no good reason. Remove? Throw?

dcsaszar (changes requested):
Mostly FYI. One prerender error.

vite.config.ts
apepper: Do you mean, because it can also look into `.env.local`?

src/config/scrivito.ts
apepper: Throw is not an option, because it would stop the control flow and would never execute `getRootComponent()` from `main.tsx`.

apepper:
REVIEW ([cf015b7](https://github.com/Scrivito/scrivito-portal-app/pull/800/commits/cf015b7d9deaeafbe32ed97bb5f30d6ebb3898b5))

vite.config.ts
dcsaszar: Instead of not passing SCRIVITO_TENANT maybe the user wanted to set PRIVATE_JR_PLATFORM?

apepper: No - `PRIVATE_JR_PLATFORM` is a private variable, that is only intended for "us" scrivito developer who are running a multi tenancy app. No-one outside of our team should use `PRIVATE_JR_PLATFORM`.

PR #806, opened 2025-03-21
LanguageSwitch: Fix hover styling

apepper:


src/assets/stylesheets/index.scss
```diff
@@ -1886,6 +1886,9 @@ header .navbar.navbar-expand-lg {
     box-shadow: 0 0 5px rgba(0, 0, 0, 0.2);
     border-radius: 5px;
     overflow: hidden;
+    .dropdown-item {
+      text-decoration: none !important;
+    }
```
dcsaszar, line 1891: * [ ] FYI In the previous commit we consolidated the selectors for no-decoration. Why not use that and add `.dropdown-item` to the list?

dcsaszar (approved):

PR #803, opened 2025-03-20
LanguageSwitchWidget

apepper:
Needs content: https://edit.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://language-switch-widget.scrivito-portal-app.pages.dev/en/gdpr-consent-form?_scrivito_workspace_id=ob60fb08cf41b148&_scrivito_display_mode=view

PR #809, opened 2025-03-25
LanguageSwitchWidget: Only offer "same content id" languages

apepper:


src/Widgets/TopNavigationWidget/SubComponents/LanguageSwitch.tsx
```diff
@@ -28,8 +30,13 @@ export const LanguageSwitch = connect(function LanguageSwitch({
       }
     })
     .sort((a, b) => a.label.localeCompare(b.label, 'en'))
+  if (!allVersions) return null
 
-  if (!versions || versions.length < 2) return null
+  const versions = onlyCurrentPageVersions
+    ? allVersions.filter(({ version }) => version)
+    : allVersions
```
dcsaszar, line 37: * [ ] I think it's better to use https://docs.scrivito.com/versionsonallsites-166f6477906beee1 in the `onlyCurrentPageVersions` case. This function does exactly what we need. The current change overcomplicates this by generating an unrelated list (page versions with root fallbacks for missing language versions) first and then reducing it back to what `versionsOnAllSites()` returns anyway with even more app logic.

apepper:
Piggy-backing refactoring for https://github.com/Scrivito/scrivito-portal-app/pull/803#discussion_r2007472322.

dcsaszar (changes requested):
See comment.

src/Widgets/TopNavigationWidget/SubComponents/LanguageSwitch.tsx
apepper: But I do need the label and `root` based on the site.

Now the code would look like this:

```ts
const rawVersions = onlyCurrentPageVersions
    ? currentPage()
        ?.versionsOnAllSites()
        .map((version) => {
          const siteId = version.siteId()
          const site = siteId && Obj.onSite(siteId).root()

          return {
            label: site ? displayName(site.language()) : '',
            version,
            root: site as HomepageInstance,
          }
        })
    : Obj.root()
        ?.versionsOnAllSites()
        .map((site) => {
          const siteId = site.siteId()
          const version = siteId && currentPage()?.versionOnSite(siteId)
          return {
            label: displayName(site.language()),
            version,
            root: site as HomepageInstance,
          }
        })
  if (!rawVersions) return null

  const versions = rawVersions.sort((a, b) =>
    a.label.localeCompare(b.label, 'en'),
  )
```

Not sure, if this is easier to understand and follow.

dcsaszar: Yes we should have as much shared logic as possible. By restructuring we could reduce the differences to the first line:
```tsx
  const versionsSource = onlyCurrentPageVersions ? currentPage() : Obj.root()
  const siteIds = versionsSource
    ?.versionsOnAllSites()
    ?.map((obj) => obj.siteId())
    .filter((v) => v !== null)
  // ... go from here based on the relevant site IDs
```

This needs an additional `Obj.onSite(siteId).root()` later on. But for optimizing this component the expanded state is a better candidate anyway.

apepper:
REVIEW ([2922a63](https://github.com/Scrivito/scrivito-portal-app/pull/809/commits/2922a63920bd3a42dab1bf5650613b45be5cb44d))

src/Widgets/TopNavigationWidget/SubComponents/LanguageSwitch.tsx
```diff
@@ -65,17 +88,18 @@ export const LanguageSwitch = connect(function LanguageSwitch({
 })
 
 const LanguageLabel = connect(function LanguageLabel({
+  label,
+  language,
   root,
   showTextLabel,
 }: {
+  label: string
+  language: string | undefined
```
dcsaszar, line 97: * [ ] FYI Passing in the language blurs the source of truth. Since `root` is mandatory, and we already get the `siteLanguageIcon` from there, we should also use `root().language()`.
* [ ] FYI I see how recycling `label` makes things cheaper and also a label is not necessarily tied to the site, so passing in `label` feels more acceptable than `language`.

dcsaszar (approved):
Some nitpicking in comments.

PR #814, opened 2025-04-01
Add QST Builder package

aea-JR:
PiSA URL needs to be changed by customers, otherwise the questionnaire can not be created or answered.
Nothing will break if there is no connection to PiSA or the url has not been set.


src/Widgets/index.ts
```diff
@@ -1,7 +1,13 @@
...
+
 import.meta.glob(['./**/*WidgetClass.ts', './**/*WidgetComponent.tsx'], {
   eager: true,
 })
 
 initNeoletterFormWidgets()
-export {}
+
+// Replace "https://your-pisa-api-url" with the actual PisaSales REST API URL
+initPisaQuestionnaireWidgets({ pisaUrl: "https://your-pisa-api-url" })
```
apepper, line 11: Please use `pisaUrl()` from `src/Data/pisaClient.ts` for this.

Maybe it makes also sense, to not call `loadQuestionnaireEditingConfigs()`, if `pisaUrl()` resolves to `null`.

apepper (changes requested):
SNEAK-PEAK FEEDBACK ([58d1c53](https://github.com/Scrivito/scrivito-portal-app/pull/814/commits/58d1c5348a2dec6b3ed6bc5405d32d45d49d0cf7))

Nice 🚀! Thank you for your contribution. Please see my comments on how to move this PR forward. Please also fix all CI reported issues.

src/Widgets/index.ts
aea-JR: This is not possible, the url needs to be accessible to the editingConfig

src/Widgets/index.ts
```diff
@@ -1,7 +1,13 @@
...
+import { initPisaQuestionnaireWidgets } from 'scrivito-pisasales-questionnaire-builder'
+import { pisaSalesApiUrl } from '../Data/pisaClient'
+
 import.meta.glob(['./**/*WidgetClass.ts', './**/*WidgetComponent.tsx'], {
   eager: true,
 })
 
 initNeoletterFormWidgets()
+
+initPisaQuestionnaireWidgets({ pisaApiUrl: pisaSalesApiUrl() })
```
apepper, line 11: If `pisaSalesApiUrl()` resolves to `null` (e.g. with the flag `FORCE_LOCAL_STORAGE`) the loading indicator of a question widget never disappears:
Screenshot 2025-07-15 at 12 59 58

apepper, line 11: This package is quite heavy. Adding this line the compiled size of `index-<some hex>.js` increases from:

```
dist/assets/index-BbviU9ra.js                                        129.23 kB │ gzip:  35.80 kB
```

to:

```
dist/assets/index-gSfHT-7i.js                                        176.50 kB │ gzip:  50.25 kB
```

If you add `scrivito-pisasales-questionnaire-builder` to the list of 

https://github.com/Scrivito/scrivito-portal-app/blob/3b61c56b5f0ceaa908db1b07220968df6d44293f/vite.config.ts#L54

The `index-<some hex>.js` stays "lean", but the resulting `scrivito-<some hex>.js` grows by almost 10 %.

From:

```
dist/assets/scrivito-CyM8ByBS.js                                     458.67 kB │ gzip: 143.28 kB
```

to
```
dist/assets/scrivito-FkHUnfW6.js                                     507.48 kB │ gzip: 157.02 kB
```

This is quite a lot for a package, that is only used in very specific cases.

apepper (changes requested):
FEEDBACK ([8f15c6d](https://github.com/Scrivito/scrivito-portal-app/pull/814/commits/8f15c6d66eed7b5fd7597add00ef3a9fff17df79))

Nice progress. Please see my comments.

src/Widgets/index.ts
apepper: I think it should be possible to reduce this number from ~ 47K to less then 10K minified, not gzipped. Gzipped the result could be like 3K, not 14K.

How? By changing the code of all components: `import { initPisaQuestionnaireWidgets } from 'scrivito-pisasales-questionnaire-builder'` would load all obj classes and also "slim" components. Those "slim" components will start loading the code to render the full components, only if the widget is actually rendered (keywork "dynamic importsW using `import()`.

aea-JR (commented):
Hello

src/Widgets/index.ts
apepper: Here is a pointer on how this could be implemented: https://react.dev/reference/react/lazy

aea-JR: Loading behaviour improved, a loading bar will be shown until the proimse gets resolved, afterwards the questionnaire will be shown.

aea-JR: No solution yet for this one, all tries failed.

apepper:
After longer consideration we decided to close this for now. For now the package should be included manually.

PR #821, opened 2025-04-09
Data form widgets: Add validation, that the widget is part of a data …

apepper:
…form container

Otherwise the widget will look like "all is good", where as no data is update/created.

Inspired by https://github.com/Scrivito/scrivito_example_app_js/blob/master/src/Widgets/FormContainerWidget/utils/validations/insideFormContainerValidation.js


dcsaszar (approved):
LGTM

PR #826, opened 2025-04-28
Data breadcrumbs

dcsaszar:
Initial draft as a basis for discussion.

Content changes: https://edit-beta.scrivito.com/d0a154d76edf2a7bd991fc658e700a1d~https://dave-data-hierarchy.scrivito-portal-app.pages.dev/en/equipment-details-1e241bae572507e8?_scrivito_workspace_id=k634a4865846fd52&_scrivito_display_mode=diff


src/Widgets/DataBreadcrumbWidget/DataBreadcrumbWidgetComponent.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const parentAttributeName = useDataLocator(widget.get('parentData'))
+    .dataItemAttribute()
+    ?.attributeName()
+  const labelAttributeName = useDataLocator(widget.get('data'))
+    .dataItemAttribute()
+    ?.attributeName()
+
+  if (!currentItem) {
+    return <nav aria-label="breadcrumb" />
+  }
```
apepper, line 22: FYI: This can be a one-liner:

```suggestion
  if (!currentItem) return <nav aria-label="breadcrumb" />
```

src/Widgets/DataBreadcrumbWidget/DataBreadcrumbWidgetComponent.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <ol className="breadcrumb m-1">
+        {breadcrumbItems.map((item) => (
+          <BreadcrumbItem
+            key={item.id()}
+            isActive={item.id() === activeId}
+            item={item}
+            label={
+              ensureString(
+                labelAttributeName && item.get(labelAttributeName),
+              ) || '<untitled>'
```
apepper, line 50: While looking at this, I noticed, that we have `<untitled>` in other places as well. Could you provide a function with different translations? E.g. `function localizeUntitle(): string`.

src/Widgets/DataBreadcrumbWidget/DataBreadcrumbWidgetComponent.tsx
```diff
@@ -0,0 +1,73 @@
...
+}: {
+  isActive: boolean
+  item: DataItem
+  label: string
+}) {
+  return (
+    <li className={isActive ? 'breadcrumb-item active' : 'breadcrumb-item'}>
+      {isActive ? label : <LinkTag to={item}>{label}</LinkTag>}
+    </li>
+  )
```
apepper, line 72: FYI: Instead of twice checking `isActive` I suggest the following code-flow instead:

```suggestion
  if (isActive) return <li className="breadcrumb-item active">{label}</li>

  return (
    <li className="breadcrumb-item">
      <LinkTag to={item}>{label}</LinkTag>
    </li>
  )
```

I think it also helps readability.

src/Widgets/DataBreadcrumbWidget/DataBreadcrumbWidgetComponent.tsx
```diff
@@ -0,0 +1,73 @@
...
+  for (
+    let ancestorItem: unknown = currentItem, loopGuardIds = new Set<string>();
+    ancestorItem instanceof DataItem && !loopGuardIds.has(ancestorItem.id());
+    ancestorItem = parentAttributeName
+      ? ancestorItem.get(parentAttributeName)
+      : null
+  ) {
+    breadcrumbItems.unshift(ancestorItem)
+    loopGuardIds.add(ancestorItem.id())
+  }
```
apepper, line 35: I find this for loop hard to understand. I would prefer a more verbose variant, that is (hopefully) easier to understand, e.g.

```ts
function breadcrumbItems(
  currentItem: DataItem | null,
  parentAttributeName?: string,
): DataItem[] {
  if (!currentItem) return []
  if (!parentAttributeName) return [currentItem]

  const breadcrumbItems: DataItem[] = []
  const loopGuardIds = new Set()

  let ancestorItem: unknown = currentItem
  while (
    ancestorItem instanceof DataItem &&
    !loopGuardIds.has(ancestorItem.id())
  ) {
    breadcrumbItems.unshift(ancestorItem)
    loopGuardIds.add(ancestorItem.id())
    ancestorItem = ancestorItem.get(parentAttributeName)
  }

  return breadcrumbItems
}
```

src/Widgets/DataBreadcrumbWidget/DataBreadcrumbWidgetClass.ts
```diff
@@ -0,0 +1,8 @@
+import { provideWidgetClass } from 'scrivito'
+
+export const DataBreadcrumbWidget = provideWidgetClass('DataBreadcrumbWidget', {
+  attributes: {
+    data: 'datalocator',
+    parentData: 'datalocator',
```
apepper, line 6: As an editor I'm confused, what `data` actually refers to. If we already have two `'datalocator'` how about also two named `'datalocator'`? This way we can also name them appropriately (e.g. `labelData: 'datalocator',`) and give them a good editor title as well.

apepper (changes requested):
FEEDBACK ([a41914b](https://github.com/Scrivito/scrivito-portal-app/pull/826/commits/a41914b1710c6861edfed4ce470203ea9cd39b84))

Great start 🚀! Please see my comments for improvements or bigger questions.

apepper (changes requested):
FEEDBACK ([0fd2b3e](https://github.com/Scrivito/scrivito-portal-app/pull/826/commits/0fd2b3e9d54f87f747c90b1d00bcabea7599ea17))

Nice 🚀. This is moving in a good direction.

src/Widgets/DataBreadcrumbWidget/DataBreadcrumbWidgetComponent.tsx
```diff
@@ -0,0 +1,77 @@
...
+    .dataItemAttribute()
+    ?.attributeName()
+
+  if (!currentItem) return <nav aria-label="breadcrumb" />
+
+  const breadcrumbItems = getBreadcrumbItems(currentItem)
+  const activeId = currentItem.id()
+
+  return (
+    <nav aria-label="breadcrumb" className="py-2">
```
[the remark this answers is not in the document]
dcsaszar: A good solution would 
* allow the editor to add/remove margin and padding
* have consistent defaults

For now I went with your suggestion, b/c putting regular and data breadcrumbs next to each other is not really looking good design-wise anyway, but can still be accomplished by bottom-aligned columns.

apepper (approved):
LGTM ([5cfecf2](https://github.com/Scrivito/scrivito-portal-app/pull/826/commits/5cfecf2b6a73384f8afb1893f4c831f6531bae28))

Very nice 🚀!

PR #838, opened 2025-05-09
GDPR: Make use of explicit Authorization option

apepper:


src/Data/CurrentUser/fetchWhoAmIWithToken.ts
```diff
@@ -16,29 +21,22 @@ export async function fetchWhoAmIWithToken(): Promise<WhoAmI | null> {
...
 
   const headers = { ...baseHeaders, Authorization: tokenAuthorization }
+  const client = createRestApiClient(url, { headers })
 
-  // TODO: Replace fetch with pisaClient, once #11616 is resolved
-  let response: Response
   try {
-    response = await fetch(url, { method: 'GET', headers })
+    const response = await client.get('')
+    return response as WhoAmI
```
dcsaszar, line 28: * [ ] FYI instead of `await` we could return the promise directly. `return client.get('') as Promise<WhoAmI>`

apepper: No 😸. Because this `await` is important for the `try/catch` block.

PR #841, opened 2025-05-13
Optimized SVGs for space usage

apepper:
Command used: `svgo src/**/*.svg` with `svgo --version` being `3.3.2`.


dcsaszar (approved):
LGTM

Maybe instead of making this part of the maintenance chores, this could be a built-in feature, e.g., with https://github.com/r3dDoX/vite-plugin-svgo.

apepper:
> Maybe instead of making this part of the maintenance chores, this could be a built-in feature, e.g., with https://github.com/r3dDoX/vite-plugin-svgo.

This feels like a waste of resources. Why compute the optimized SVG over and over again?

dcsaszar:
> Why compute the optimized SVG over and over again?

To save the maintainer(s) of a(ll) project(s) based on the Portal App from having to constantly think about it after a new SVG has been added.
If resources matter it could be part of the prod build only.

But I understand that it's a tradeoff and there's no "right" or "wrong".

PR #843, opened 2025-05-14
Extract `pisaBaseUrl` function

apepper:
This is closes to the upcoming "Common URL path for all PisaSales REST APIs"

### DIFF
--- a/src/Data/pisaClient.ts
+++ b/src/Data/pisaClient.ts
@@ -1,15 +1,33 @@
-import { Obj, createRestApiClient, currentLanguage, load } from 'scrivito'
+import {
+  Obj,
+  createRestApiClient,
+  currentLanguage,
+  currentUser,
+  load,
+} from 'scrivito'
 import { isHomepage } from '../Objs/Homepage/HomepageObjClass'
 
-export async function pisaUrl(): Promise<string | null> {
+async function pisaBaseUrl(): Promise<string | null> {
   if (import.meta.env.FORCE_LOCAL_STORAGE) return null
+  if (!(await load(() => currentUser()))) return null
 
+  // TODO: Read out `pisa_sales_api_url` from AMS instance configuration once available
   const defaultRoot = await load(() =>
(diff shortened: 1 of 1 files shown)


src/Data/pisaClient.ts
```diff
@@ -1,15 +1,33 @@
...
+  currentLanguage,
+  currentUser,
+  load,
+} from 'scrivito'
 import { isHomepage } from '../Objs/Homepage/HomepageObjClass'
 
-export async function pisaUrl(): Promise<string | null> {
+async function pisaBaseUrl(): Promise<string | null> {
   if (import.meta.env.FORCE_LOCAL_STORAGE) return null
+  if (!(await load(() => currentUser()))) return null
```
dcsaszar, line 12: * [ ] FYI `isUserLoggedIn` is the simpler alternative here, see https://docs.scrivito.com/isuserloggedin-a84bd9ef936b26f6#isUserLoggedIn-vs-currentUser

src/Data/pisaClient.ts
```diff
@@ -1,15 +1,33 @@
...
 
+  // TODO: Read out `pisa_sales_api_url` from AMS instance configuration once available
```
dcsaszar, line 14: * [ ] FYI This TODO can probably only be understood by the one who wrote it. Remove?

src/Data/pisaClient.ts
apepper: But `isUserLoggedIn` only works, if `Scrivito.configure` was called previously. This is not always the case with `pisaBaseUrl()`.

dcsaszar: If this is true either the docs are wrong or the SDK has a bug:

> the application logic never has to wait for the result of isUserLoggedIn, and is not required to handle the case that the value of isUserLoggedIn may change.

apepper:
This will break this existing GDPR JWT token solution. Closing for now, until we found a better way.

PR #845, opened 2025-05-16
Timeout notification for PisaSales

apepper:
Shows a timeout toast after 5 seconds. Toast disappears, if connection gets established later.

Screenshot 2025-05-16 at 12 31 05

### DIFF
--- a/src/Data/notifyOnConnectionTimeout.tsx
+++ b/src/Data/notifyOnConnectionTimeout.tsx
@@ -1,0 +1,54 @@
+import { currentLanguage } from 'scrivito'
+import { pseudoRandom32CharHex } from '../utils/pseudoRandom32CharHex'
+import { toast } from 'react-toastify'
+
+export async function notifyOnConnectionTimeout<T>(connection: Promise<T>) {
+  const connectionSucceeded = await Promise.race([
+    connection.then(() => true),
+    waitFor(5).then(() => false),
+  ])
+  if (connectionSucceeded) return
+
+  const toastId = pseudoRandom32CharHex()
+
+  toast.warning(
+    <div>
+      <div>{localizeUnableToConnect()}</div>
+      <small>{localizePisaSalesServerRunning()}</small>
+    </div>,
+    { autoClose: false, toastId },
+  )
+
+  await connection
+  toast.dismiss(toastId)
+}
+
+async function waitFor(seconds: number) {
+  return new Promise((resolve) => setTimeout(resolve, seconds * 1000))
+}
+
+function localizeUnableToConnect() {
+  switch (currentLanguage()) {
+    case 'de':
+      return 'Keine Backend-Verbindung.'
+    case 'fr':
+      return 'Pas de connexion au backend.'
+    case 'pl':
+      return 'Brak połączenia z backendem.'
+    default:
+      return 'No backend connection.'
+  }
+}
+
+function localizePisaSalesServerRunning() {
+  switch (currentLanguage()) {
+    case 'de':
+      return 'Läuft der PisaSales-Server?'
+    case 'fr':
+      return "Le serveur PisaSales est-il en cours d'exécution ?"
+    case 'pl':
+      return 'Czy serwer PisaSales działa?'
+    default:
+      return 'Is the PisaSales server running?'
+  }
+}
--- a/src/Data/CurrentUser/CurrentUserDataItem.ts
+++ b/src/Data/CurrentUser/CurrentUserDataItem.ts
@@ -15,7 +15,7 @@ import { getTokenAuthorization } from '../getTokenAuthorization'
 import { errorToast } from './errorToast'
 import { pisaClient } from '../pisaClient'
 import { fetchWhoAmIWithToken } from './fetchWhoAmIWithToken'
-import { notifyOnConnectionTimeout } from '../notifyOnConnectionTimeout'
+import { notifyOnConnectionTimeout } from './notifyOnConnectionTimeout'
 
 async function attributes(): Promise<DataAttributeDefinitions> {
   const lang = await load(currentLanguage)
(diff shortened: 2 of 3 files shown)


src/Data/notifyOnConnectionTimeout.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { toast } from 'react-toastify'
+
+export async function notifyOnConnectionTimeout<T>(connection: Promise<T>) {
+  const connectionSucceeded = await Promise.race([
+    connection.then(() => true),
+    waitFor(5).then(() => false),
+  ])
+  if (connectionSucceeded) return
+
+  const toastId = pseudoRandom32CharHex()
```
dcsaszar, line 12: * [ ] As can be seen in https://fkhadra.github.io/react-toastify/update-toast#basic-example we get the ID as the result of the API call. No need to generate IDs -> simpler code.

src/Data/notifyOnConnectionTimeout.tsx
```diff
@@ -0,0 +1,54 @@
...
+      return 'No backend connection.'
+  }
+}
+
+function localizePisaSalesServerRunning() {
+  switch (currentLanguage()) {
+    case 'de':
+      return 'Läuft der PisaSales-Server?'
+    case 'fr':
+      return "Le serveur PisaSales est-il en cours d'exécution ?"
```
dcsaszar, line 48: * [ ] FYI Please prefer non-breaking space (`\u00a0`) to avoid line breaks between text and question mark.

src/Data/notifyOnConnectionTimeout.tsx
```diff
@@ -0,0 +1,54 @@
+import { currentLanguage } from 'scrivito'
```
dcsaszar, line 1: * [ ] FYI Regarding the location of the module: The only caller is  `CurrentUserDataItem`. There should be no other caller (otherwise we'd show multiple toasts). Thus I think the module should live in `src/Data/CurrentUser/`.

src/Data/notifyOnConnectionTimeout.tsx
```diff
@@ -0,0 +1,54 @@
...
+    { autoClose: false, toastId },
+  )
+
+  await connection
+  toast.dismiss(toastId)
+}
+
+async function waitFor(seconds: number) {
+  return new Promise((resolve) => setTimeout(resolve, seconds * 1000))
+}
```
dcsaszar, line 28: * [ ] FYI react-toastify has built-in support for promises, so we could simplify our code. Prototype:

```tsx
  toast.promise(waitFor(6), {
    pending: {
      delay: 5000,
      icon: Icons.warning,
      type: 'warning',
      render: () => (
        <div>
          <div>{localizeUnableToConnect()}</div>
          <small>{localizePisaSalesServerRunning()}</small>
        </div>
      ),
      closeButton: true,
    },
    success: { delay: 0, autoClose: 1 },
  })
```

src/Data/notifyOnConnectionTimeout.tsx
```diff
@@ -0,0 +1,54 @@
...
+  toast.warning(
+    <div>
+      <div>{localizeUnableToConnect()}</div>
+      <small>{localizePisaSalesServerRunning()}</small>
+    </div>,
+    { autoClose: false, toastId },
+  )
+
+  await connection
+  toast.dismiss(toastId)
```
dcsaszar, line 23: * [ ] FYI Maybe change the behavior to how `toast.promise` behaves by default: If the promise toast resolves, there is a success message. This will make the state change clearer than just hiding the toast. For example, the visitor may not have had enough time to read the message of the "pending" toast.

dcsaszar (changes requested):
Nice, but has a lot of room for simplification.

src/Data/notifyOnConnectionTimeout.tsx
apepper: Nice 🚀. Good pointer.

apepper: Sadly with the following code the pending will always be shown, regardless of the state of `connection`:

```tsx
import { currentLanguage } from 'scrivito'
import { toast, Icons } from 'react-toastify'

export async function notifyOnConnectionTimeout<T>(connection: Promise<T>) {
  toast.promise(connection, {
    pending: {
      delay: 5000,
      icon: Icons.warning,
      type: 'warning',
      render() {
        return (
          <div>
            <div>{localizeUnableToConnect()}</div>
            <small>{localizePisaSalesServerRunning()}</small>
          </div>
        )
      },
      closeButton: true,
    },
  })
}

function localizeUnableToConnect() {
  switch (currentLanguage()) {
    case 'de':
      return 'Keine Backend-Verbindung.'
    case 'fr':
      return 'Pas de connexion au backend.'
    case 'pl':
      return 'Brak połączenia z backendem.'
    default:
      return 'No backend connection.'
  }
}

function localizePisaSalesServerRunning() {
  switch (currentLanguage()) {
    case 'de':
      return 'Läuft der PisaSales-Server?'
    case 'fr':
      return "Le serveur PisaSales est-il en cours d'exécution\u00a0?"
    case 'pl':
      return 'Czy serwer PisaSales działa?'
    default:
      return 'Is the PisaSales server running?'
  }
}

```

dcsaszar: Maybe if `success` is missing `toast.promise` doesn't know what to do? What happens if we add `success`?

apepper: I tried this as well - still the same behavior.

apepper:
REVIEW ([c4e3675](https://github.com/Scrivito/scrivito-portal-app/pull/845/commits/c4e367559e561a00d8cb0fc4dd9ec34bce78d0c5))

dcsaszar (approved):
LGTM

PR #848, opened 2025-05-22
Include Honeybadger right away

apepper:
This way boot time exceptions are also reliably logged to Honeybadger.

The stub is provided to reduce payload, if Honeybadger is not needed.


dcsaszar (approved):
LGTM

PR #868, opened 2025-07-01
Font: Consider `application/font-woff` to be a font

apepper:
This is the old mime-type still in use by some systems (e.g. recent MacOS). See https://www.iana.org/assignments/media-types/media-types.xhtml for reference.

How to reproduce: Download `woff` version from https://de.bestfonts.pro/font/arbutus-slab and try to use it as a custom font.

Originally reported by @agessler.


dcsaszar (approved):
Weird stuff.

PR #884, opened 2025-07-04
Better "Skip to content" button

apepper:
Screenshot 2025-07-04 at 12 09 50

How to see this in action:
* Visit https://improved-skip-to-content-but.scrivito-portal-app.pages.dev/en
* Hit "tab" button

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -55,6 +55,6 @@ function localizeSkipToContent(): string {
     case 'pl':
       return 'Przejdź do treści'
     default:
-      return 'Skip to Content'
+      return 'Skip to content'
   }
 }


src/App.tsx
```diff
@@ -35,3 +33,24 @@ export function App({
...
+function localizeSkipToContent(): string {
+  switch (currentLanguage()) {
+    case 'de':
+      return 'Zum Inhalt springen'
+    case 'fr':
+      return 'Aller au contenu'
+    case 'pl':
+      return 'Przejdź do treści'
+    default:
+      return 'Skip to Content'
```
dcsaszar, line 58: * [ ] FYI Not sure which would be the most consistent, but I quickly found `localizeLogOutLabel` which uses sentence case. I actually expected sentence case here, not title case.

dcsaszar (approved):
Fancy. One FYI.

src/App.tsx
apepper: Nice catch!

PR #887, opened 2025-07-08
PageTitleWidget: Use level h1

apepper:
Most PageTitleWidgets will be placed at the very top of the page. Therefore a h1 is quite fitting. Styling is not affected.


dcsaszar (approved):
Sounds reasonable

PR #898, opened 2025-07-24
CardWidget allows video background

apepper:
[Concept](https://docs.google.com/document/d/1q4RqsL8alSl34ulBr2WRdOhKJMfQtz-fFnzSRXYp4fY/edit?tab=t.0#heading=h.d1qjg5rpzwfl)

Since the data class changes needs to be approved independently this PR is only send to energy during the vacation season.

PR #904, opened 2025-07-28
ImageOrVideo: Pause autoplay video on click

apepper:
[Concept](https://docs.google.com/document/d/1q4RqsL8alSl34ulBr2WRdOhKJMfQtz-fFnzSRXYp4fY/edit?tab=t.0#heading=h.gyss5p15hl8)

https://github.com/user-attachments/assets/45cdbbee-2967-46b7-99bb-c74dc90eea4b

Since this change builds on the refactoring included in https://github.com/Scrivito/scrivito-portal-app/pull/898 I send this PR to energy.


agessler:
Is this needed? It looks ugly the play button there.

agessler:
@apepper https://github.com/Scrivito/scrivito-portal-app/pull/909

apepper:
REVIEW ([a7776f0](https://github.com/Scrivito/scrivito-portal-app/pull/904/commits/a7776f0eb0ccea3629dd918352fb9e700f865e1f))

src/Widgets/CardWidget/CardWidgetComponent.tsx
```diff
@@ -30,13 +32,17 @@ provideComponent(CardWidget, ({ widget }) => {
   if (widget.get('cardExtended')) cardClassNames.push('card-extended')
 
   return (
-    <WidgetTag className={cardClassNames.join(' ')}>
+    <WidgetTag
+      className={cardClassNames.join(' ')}
+      onClick={(e) => togglePlayPauseRef.current?.togglePlayPause(e)}
+    >
```
dcsaszar, line 38: * [ ] FYI Having this logic in all parent components has a certain complexity. Since we handle clicks anywhere, how about using `onMouseUp` instead of `onClick`? The mouse events bubble by default.

src/Components/ImageOrVideo.tsx
```diff
@@ -26,8 +26,22 @@ export const ImageOrVideo = connect(function ImageOrVideo<T extends string>({
   }>
 }) {
   const [isPaused, setIsPaused] = useState(false)
+  const [shouldAutoplay, setShouldAutoplay] = useState(true)
   const videoRef = useRef<HTMLVideoElement>(null)
 
+  useEffect(() => {
+    const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)')
+    setShouldAutoplay(!mediaQuery.matches)
+    if (mediaQuery.matches) setIsPaused(true)
```
dcsaszar, line 35: * [ ] We should definitely not put the play button on top. For reduced motion it may cover elements the user is supposed to interact with by default.

dcsaszar (changes requested):
See comments for better usability and reduced code complexity

src/Widgets/CardWidget/CardWidgetComponent.tsx
dcsaszar: I see. How about making ImageOrVideo a wrapper component? So instead of
```tsx
<WidgetTag>
  <ImageOrVideo />
  { stuff }
</WidgetTag>
```

we'd go with the event bubbling version:
```tsx
<WidgetTag>
  <ImageOrVideo>
    { stuff }
  </ImageOrVideo>
</WidgetTag>
```
We could rename it to `<BackgroundImageOrVideo>`.
We need an additional wrapper, but this can correctly handle click events on children.
If the added wrapper causes CSS issues, we could try it with `display: contents;`.

src/Components/ImageOrVideo.tsx
dcsaszar: > Adding a second video tag feels like a hack to me

The root cause is that we're using a setting that was introduced for CSS to control a DOM element. Hiding/showing DOM elements is one of the few things CSS can do to "manipulate" the DOM.
With this in mind, JS media queries also feel like a hack.
Currently our hack renders a non-autoplaying video as autoplaying in prerendered HTML (and on navigating, for a brief moment), because of how we initialize the state and the use of `useEffect` instead of `useLayoutEffect`. If we stick with the JS version we should at least

* [ ] Render non-playing initially

src/Components/ImageOrVideo.scss
```diff
@@ -15,6 +15,14 @@
   border: none;
   cursor: pointer;
   transition: all ease-in-out 0.4s;
+  opacity: 0;
+  pointer-events: none;
+
+  &.is-paused {
+    opacity: 1;
+    pointer-events: auto;
```
dcsaszar, line 22: * [ ] Still interferes with user input for people with reduced motion preference. How about only setting this style if motion is preferred? It would no longer animate on hover in the "reduced motion" case, but this is actually good.

src/Components/ImageOrVideo.scss
```diff
@@ -11,7 +11,6 @@
   display: flex;
   align-items: center;
   justify-content: center;
-  z-index: 1111;
```
dcsaszar: * [ ] FYI doesn't work for me. On Chrome `z-index: -1` appears to do the trick?

src/Components/ImageOrVideo.tsx
```diff
@@ -31,15 +31,21 @@ export const ImageOrVideo = connect(function ImageOrVideo<T extends string>({
 
   useEffect(() => {
     const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)')
-    setShouldAutoplay(!mediaQuery.matches)
-    if (mediaQuery.matches) setIsPaused(true)
-
-    const handleChange = () => {
-      setShouldAutoplay(!mediaQuery.matches)
+    const updatePreferences = () => {
+      if (mediaQuery.matches) {
```
dcsaszar, line 35: * [ ] FYI Naming the `mediaQuery` after what it queries would improve readability. E.g. `if (prefersReducedMotion.matches)`.

dcsaszar (changes requested):
Good progress. I like how the button now always nicely fades in and out.

src/Widgets/CardWidget/CardWidgetComponent.tsx
apepper: I gave it a try, but there is quite a lot of css that assumes, that the background image/video is a direct child of the card. I would leave it as is for now.

apepper:
REVIEW ([27790b3](https://github.com/Scrivito/scrivito-portal-app/pull/904/commits/27790b3f37d4dcd60f340dd70ae87838dfbd28a6))

src/Components/ImageOrVideo.tsx
```diff
@@ -26,10 +26,10 @@ export const ImageOrVideo = connect(function ImageOrVideo<T extends string>({
   }>
 }) {
   const [isPaused, setIsPaused] = useState(false)
-  const [shouldAutoplay, setShouldAutoplay] = useState(true)
+  const [shouldAutoplay, setShouldAutoplay] = useState(false)
```
dcsaszar, line 29: * [ ] FYI Looking at the reducedMotion logic, the initial state of `isPaused` looks a bit contradictory now.

apepper: It was set on false on purpose: With this I wanted to avoid a slight "button available" flicker, when the page is initially rendered.

PR #907, opened 2025-07-28
Support multi-site in addition to multi-language

dcsaszar:
It is pretty hard to keep the many possible URL formats and fallbacks in mind when changing the code.
Maybe we can come up with a better code structure in a follow-up. I've identified at least these responsibilities
- Known URL schemes and their precedence (including multi-tenancy, content-ID-based, default site, via baseUrl)
- Recognizing sites
- Redirecting to fallback URLs

### DIFF
--- a/src/config/scrivitoSites.ts
+++ b/src/config/scrivitoSites.ts
@@ -7,10 +7,7 @@ import {
   urlFor,
 } from 'scrivito'
 import { ensureString } from '../utils/ensureString'
-import {
-  getJrPlatformInstanceBaseUrl,
-  jrPlatformRedirectToSiteUrl,
-} from '../privateJrPlatform/multiTenancy'
+import { getJrPlatformInstanceBaseUrl } from '../privateJrPlatform/multiTenancy'
 
 const origin =
   typeof window !== 'undefined'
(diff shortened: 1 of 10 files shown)


src/config/scrivitoSites.ts
```diff
@@ -85,7 +121,7 @@ function baseUrlsFor(site: Obj) {
 }
 
 export function isNoSitePresent(): boolean {
-  return !appWebsites()?.length
+  return !getLanguageVersions()?.length
 }
 
 export async function ensureSiteIsPresent() {
```
apepper, line 127: As discussed in a voice call: How about extracting `ensureSiteIsPresent` to a dedicated module, that should only know about finding a good language version for a detected content ID?

apepper (changes requested):
FEEDBACK ([30211f6](https://github.com/Scrivito/scrivito-portal-app/pull/907/commits/30211f63ca69ee7e60ddd5216364ff31a7862e20))

Great start 🚀!

Please see my comments on how to move this PR forward.

src/config/scrivitoSites.ts
```diff
@@ -43,23 +44,35 @@ export function siteForUrl(
...
+  return { contentId, language, siteId }
+}
+
+function findSiteIdBy(query: { contentId: string; language: string }) {
+  return Obj.onAllSites()
+    .where('_path', 'equals', '/')
+    .and('_contentId', 'equals', query.contentId)
+    .and('_language', 'equals', query.language)
+    .toArray()[0]
+    ?.siteId()
```
apepper, line 75: FYI: In a follow up we can try to improve the loading behavior, so that we only execute the minimum amount of search requests related to site finding/picking.

apepper (approved):
LGTM ([080ce00](https://github.com/Scrivito/scrivito-portal-app/pull/907/commits/080ce0037d9e0143fac66aeb07bcadb4cc5cd3b3))

Nice 🚀!

I left some follow up comments to improve loading speed.

Issue #901, opened 2025-07-25
App focus style bleeds into extensions


dcsaszar:
### Actual

<img width="185" height="124" alt="Image" src="https://github.com/user-attachments/assets/6641f0bb-9f4e-4798-8eb3-26bfbe165de0" />

### Expected

A focus stlye like here:

<img width="185" height="88" alt="Image" src="https://github.com/user-attachments/assets/fecdf6bb-d905-4f5c-8e4d-d75b7cd944f3" />

PR #910, opened 2025-07-30
disable styles for focus visible state in UI

agessler:
FIX for https://github.com/Scrivito/scrivito-portal-app/issues/901
 
 I have disbales all focus-visible styles for data-attribute [data-scrivito-editing]. Was the only thing I found to address.

### DIFF
--- a/src/assets/stylesheets/index.scss
+++ b/src/assets/stylesheets/index.scss
@@ -109,6 +109,14 @@ a:focus-visible,
   transition: all ease-in-out 0.3s;
 }
 
+[data-scrivito-editing] {
+  a:focus-visible,
+  .btn:focus-visible,
+  :focus-visible {
+    outline: none !important;
+  }
+}
+
 .btn {
   --bs-btn-border-color: transparent;
   --bs-btn-hover-border-color: transparent;


src/assets/stylesheets/index.scss
```diff
@@ -109,6 +109,14 @@ a:focus-visible,
...
 }
 
+[data-scrivito-editing] {
+  a:focus-visible,
+  .btn:focus-visible,
+  :focus-visible {
+    outline: none !important;
+  }
+}
+
```
dcsaszar, line 119: * [ ] Rules in index.scss also affect the app and because of this WYSIWYG is broken in edit mode. Extension-only styles can be put into: `src/assets/stylesheets/scrivitoExtensions.scss`.

src/assets/stylesheets/index.scss
```diff
@@ -109,6 +109,14 @@ a:focus-visible,
   transition: all ease-in-out 0.3s;
 }
 
+[data-scrivito-editing] {
+  a:focus-visible,
+  .btn:focus-visible,
+  :focus-visible {
+    outline: none !important;
```
dcsaszar, line 116: * [ ] This disables the focus style completely, which defeats the color picker input styles. If this is intended, we should delete these styles.

dcsaszar (changes requested):
Found some issues.

Maybe we should apply the special focus styles more selectively, because even in the app it appears sometimes inappropriate, e.g., on the slider back/forward nav?

* [ ] FYI However a quick solution would be wrapping the app styles in
```css
#app-wrapper {
}
```

PR #913, opened 2025-07-30
Fix card background border radius in editing

dcsaszar:


agessler (approved):

agessler (commented):
I thought an important was necessary?
 
` border-radius: inherit !important;`

PR #922, opened 2025-08-04
Homepage: Validate correct language

apepper:
Follow up of https://github.com/Scrivito/scrivito-portal-app/pull/907.

Otherwise the multi-site setup might be broken.

The UX is probably not ideal, but at least you can't publish broken sites:

No language
Multiple languages

### DIFF
--- a/src/Objs/Homepage/HomepageEditingConfig.ts
+++ b/src/Objs/Homepage/HomepageEditingConfig.ts
@@ -1,4 +1,4 @@
-import { provideEditingConfig, Obj } from 'scrivito'
+import { provideEditingConfig } from 'scrivito'
 import { Homepage } from './HomepageObjClass'
 import { SiteColorsPicker } from './SiteColorsPicker'
 import {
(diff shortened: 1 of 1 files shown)


src/Objs/Homepage/HomepageEditingConfig.ts
```diff
@@ -128,5 +128,32 @@ provideEditingConfig(Homepage, {
...
+            severity: 'error',
+          }
+        }
+
+        const contentId = obj.contentId()
+        const duplicates = Obj.onAllSites()
+          .where('_objClass', 'equals', 'Homepage')
+          .and('_contentId', 'equals', contentId)
+          .and('_language', 'equals', language)
+          .count()
```
dcsaszar, line 148: * [ ] FYI How about iterating over `versionsOnAllSites()`? This may look less efficient but the result is already in memory b/c of the language switch logic.

dcsaszar (approved):
LGTM. One performance suggestion that may even improve readability.

apepper:
REVIEW ([bf2bac6](https://github.com/Scrivito/scrivito-portal-app/pull/922/commits/bf2bac622af1af61d8f1f2f8a507614c3b80938d))

dcsaszar (approved):

PR #928, opened 2025-08-06
Site rounded corners only px

apepper:
Follow up of https://github.com/Scrivito/scrivito-portal-app/pull/900.

The [concept](https://docs.google.com/document/d/1q4RqsL8alSl34ulBr2WRdOhKJMfQtz-fFnzSRXYp4fY/edit?tab=t.0#heading=h.12d9pvcch0xk) was updated to only support `px` as a unit. This PR removes `rem` as the unit.


dcsaszar (approved):
* [ ] FYI Now that it's basically unit-less, we could make it an `integer` attribute and merge it into site settings. Basically one step back to what we had before when it was a boolean attribute.

apepper:
> FYI Now that it's basically unit-less, we could make it an integer attribute and merge it into site settings.

I want the format to be closer to dartagnan and have a "easy" path forward, if we decide to support other units as well.

PR #934, opened 2025-08-11
CSS: Bugfix - Fix "No semantic tag" styles of HeadlineWidget

apepper:
They should look identical to having a h1 to h6 associated.


dcsaszar (approved):
LGTM

PR #936, opened 2025-08-12
Scrivito 1.53.0

apepper:


src/Widgets/HeadlineWidget/HeadlineWidgetComponent.tsx
```diff
@@ -28,6 +28,7 @@ provideComponent(HeadlineWidget, ({ widget }) => {
       attribute="headline"
       className={classNames.join(' ')}
       tag={tag(widget.get('level'), style)}
+      // TODO: replace `speakingUrl` with our slugify package once available
```
dcsaszar, line 31: * [ ] FYI Not a big fan of this type of code comment. Maybe move this to a todo list?

src/Data/CartItem/CartItemDataClass.ts
```diff
@@ -21,11 +21,11 @@ async function attributes(): Promise<DataAttributeDefinitions> {
   }
 }
 
-export const CartItem = provideDataClass('CartItem', {
+export const CartItem = provideDataClass('CartItem', () => ({
```
dcsaszar, line 24: * [ ] FYI This is now more complicated than before. Is there a reason for that?

dcsaszar (approved):
LGTM

src/Data/CartItem/CartItemDataClass.ts
apepper: This is needed, since `localStorageDataConnection` requires a instance ID to be configured. If you visit the multi-tenancy app without an instance ID the `CartItem` params would be executed right away. With making it a callback it will only be executed, if `CartItem` is actually needed.

PR #948, opened 2025-09-02
[blobToBinary] Bugfix - allow octet-stream base64 strings

apepper:
Sometimes the prefix of dataUrl of the FileReader output, does not match `blob.type`. In this case it is better to allow the generic mime type `application/octet-stream` instead.

Fixes https://app.honeybadger.io/projects/132043/faults/123009878.

### DIFF
--- a/src/utils/blobToBinary.ts
+++ b/src/utils/blobToBinary.ts
@@ -50,7 +50,7 @@ async function blobToBase64(blob: Blob): Promise<string> {
       }
 
       const parts = dataUrl.split(';base64,')
-      const [_firstPart, secondPart, ...rest] = parts
+      const [, secondPart, ...rest] = parts
       if (!secondPart || rest.length > 0) {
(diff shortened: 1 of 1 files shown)


dcsaszar (commented):
I've read through most of https://stackoverflow.com/questions/18650168/convert-blob-to-base64 and I can't understand why our implementation is so long. It would be great if the fixed code would become shorter instead of longer.

* [ ] Could we go with a simple approach to get the base64 and then have straightforward blob.type -> data-url -> error fallbacks for the content type?
If not, why? Do we have a set of test files so I can reproduce the known issues?

apepper:
REVIEW ([e2c9939](https://github.com/Scrivito/scrivito-portal-app/pull/948/commits/e2c993981962c08880af25630560f68c4a65d247))

`blobToBase64` is now independent of `blob.type`.

PR #955, opened 2025-09-03
[typescript] Upgrade "moduleResolution" to more modern "bundler"

apepper:
See https://www.typescriptlang.org/docs/handbook/modules/reference.html#node10-formerly-known-as-node for background.


apepper:
REVIEW ([5fefa34](https://github.com/Scrivito/scrivito-portal-app/pull/955/commits/5fefa3471f6e1eee140335beeca84d2c30385f31))

PR #957, opened 2025-09-09
Replace SCRIVITO_ROOT_OBJ_ID with SCRIVITO_DEFAULT_CONTENT_ID

apepper:
Motivation: Currently the setup relies on a "magic obj ID". When this obj is existing all is good, but as soon as this obj is deleted we would have to document created a specific Obj with a specific `_id`. Instead one now has to "only" set a special `_contentId` for the root obj.

So https://docs.scrivito.com/js-sdk-cheat-sheet#deleting-all-website-content would look like this _after_ this PR:

```
// objs = await ...

Scrivito.getClass("Homepage").create({
  _path: "/",
  _language: "en",
  _contentId: "c2a0aab78be05a4e",
});

// objs.forEach((o) => { ...
```

### DIFF
--- a/src/Components/NotFoundErrorPage.tsx
+++ b/src/Components/NotFoundErrorPage.tsx
@@ -10,17 +10,14 @@ import {
   NotFoundErrorPage as ScrivitoNotFoundErrorPage,
 } from 'scrivito'
 import { Loading } from './Loading'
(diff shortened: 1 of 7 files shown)


dcsaszar (changes requested):
* [ ] I haven't looked at the changes yet but saw the missing extensions running the branch:

Image

PR #969, opened 2025-09-25
R.I.P. react-helmet

apepper:
Or more specifically `@dr.pogodin/react-helmet`. Thanks for the ride so far. But with React 19 most of the utilities are no longer needed. This saves ~5kb gzipped payload.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,6 +7,7 @@ import { NotFoundErrorPage } from './Components/NotFoundErrorPage'
 import { Toasts } from './Components/Toasts'
 import { DesignAdjustments } from './Components/DesignAdjustments'
 import { SinglePageSite } from './Components/SinglePageSite'
+import { Favicon } from './Components/Favicon'
 
 export const helmetContext: { helmet?: HelmetServerState } = {}
 
@@ -24,6 +25,7 @@ export function App({
             <SinglePageSite>
               <CurrentPage />
               <NotFoundErrorPage />
+              <Favicon />
               <CurrentPageMetadata />
             </SinglePageSite>
           </DesignAdjustments>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
 import { connect, currentLanguage, CurrentPage } from 'scrivito'
-import { HelmetProvider, HelmetServerState } from '@dr.pogodin/react-helmet'
 
 import { CurrentPageMetadata } from './Components/CurrentPageMetadata'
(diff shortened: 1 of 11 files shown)


src/prerenderContent/prerenderObj.tsx
```diff
@@ -46,3 +47,14 @@ export async function prerenderObj(
     },
   ]
 }
+
+function extractTitle(html: string): {
+  title: string
+  bodyContent: string
+} {
+  const titleMatch = html.match(/<title[^>]*>.*?<\/title>/i)
```
dcsaszar, line 50: * [ ] ReDoS vulnerable

src/App.tsx
```diff
@@ -24,6 +25,7 @@ export function App({
             <SinglePageSite>
               <CurrentPage />
               <NotFoundErrorPage />
+              <Favicon />
```
dcsaszar: * [ ] The `link` position leads to an HTML validation error:

> Error: A link element must not appear as a descendant of a body element unless the link element has an itemprop attribute or has a rel attribute whose value contains dns-prefetch, modulepreload, pingback, preconnect, prefetch, preload, prerender, or stylesheet.

src/Components/CurrentPageMetadata.tsx
```diff
@@ -14,8 +14,15 @@ export const CurrentPageMetadata = connect(() => {
         htmlAttributes={{
           lang: page?.language() || 'en',
         }}
-        meta={getMetadata(page)}
       />
+      {getMetadata(page).map(({ name, property, content }) => (
+        <meta
```
dcsaszar, line 19: * [ ] Validation errors b/c of wrong position:

> Attribute name not allowed on element [meta](https://html.spec.whatwg.org/multipage/#the-meta-element) at this point.

> Element [meta](https://html.spec.whatwg.org/multipage/#the-meta-element) is missing one or more of the following attributes: itemprop, property.

dcsaszar (changes requested):
https://validator.w3.org/ is complaining about the current outcome. Also, ReDoS. See comments.

apepper:
I'm going to close this for now. So far I was not able to produce code, that

(a) prerenders nicely without hydration errors.
(b) produces valid html (according to https://validator.w3.org/nu/).
(c) does not produce console.log warning during regular development.

One or more of those points where in the way with different approaches .

Maybe I'll revisit this in half a year or so to see, if my skills improved or the React eco-system at large.

PR #971, opened 2025-10-01
easier to read color adaption

agessler:
https://codepen.io/chriscoyier/pen/ZEZKqvq

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix

### DIFF
--- a/src/assets/stylesheets/index.scss
+++ b/src/assets/stylesheets/index.scss
@@ -216,8 +216,8 @@ fieldset:disabled .btn {
 // Base button variant styles using CSS custom properties
 .btn-primary,
 .btn-secondary {
-  --btn-color-lighten: hsl(from var(--btn-color) h s calc(l + 8));
-  --btn-color-darken: hsl(from var(--btn-color) h s calc(l - 8));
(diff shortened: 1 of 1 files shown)


src/assets/stylesheets/index.scss
```diff
@@ -216,8 +216,8 @@ fieldset:disabled .btn {
 // Base button variant styles using CSS custom properties
 .btn-primary,
 .btn-secondary {
-  --btn-color-lighten: hsl(from var(--btn-color) h s calc(l + 8));
-  --btn-color-darken: hsl(from var(--btn-color) h s calc(l - 8));
+  --btn-color-lighten: color-mix(in oklch, var(--btn-color), white 18%);
```
apepper, line 219: There are at least four more cases of `hsl` usage in `index.scss`. I would prefer to always use the same syntax.

apepper (commented):
FIRST-LOOK FEEDBACK ([2f1874f](https://github.com/Scrivito/scrivito-portal-app/pull/971/commits/2f1874fde644824e03c47f15141b0f0b56dc5d38))

PR #1067, opened 2026-03-27
Add Country data class

apepper:
This is only for completeness sake. PisaSales has it in its "portal/_manifest" file included. AFAIK country is used during the cpq configuration in paulas journey. Please note, that the initialContent is only a small subset on what an actual PisaSale server will respond.


src/Data/Country/countryParamsFallback.ts
```diff
@@ -0,0 +1,87 @@
+import { currentLanguage, DataAttributeDefinitions, load } from 'scrivito'
+import { localStorageDataConnection } from '../localStorageDataConnection'
+
+async function attributes(): Promise<DataAttributeDefinitions> {
+  const lang = await load(() => currentLanguage())
+
+  return {
+    countryCode: [
+      'string',
+      { title: lang === 'de' ? 'Telefonvorwahl' : 'Telephone country code' },
```
dcsaszar: * [x] FYI strictly and technically the "Vorwahl" is "+49", while the "Code" is "49". Looking at the usage we include the "+", so a less ambiguous EN wording could be "Country calling code" or "International dialing code".

dcsaszar (approved):
LGTM

src/Data/Country/countryParamsFallback.ts
apepper: Ronald changed it to `Landesvorwahl`/`Country calling code`.

dcsaszar (approved):
👯

PR #1068, opened 2026-03-27
Bugfix - Prerender: Make design adjustment !important

apepper:
~Otherwise `<link rel="stylesheet" ...` placed by vite will have higher priority.~

Test with:
* npm run prerender
* npm run prerender-preview
* Visit http://localhost:8080/71ec0827969c2bd8/en.html within your browser - disable javascript.

Expected: Background color is correct (yellowish)
Actual (prior to this commit): Background color was wrong (bluish)


src/prerenderContent/generateHtml.ts
```diff
@@ -11,11 +11,13 @@ export async function generateHtml(
...
   },
 ): Promise<string> {
-  return keys.reduce((html, key) => {
+  const html = keys.reduce((html, key) => {
     const re = replacements[key]
     if (html.match(re)) return html.replace(re, () => data[key])
     throw new Error(`Missing placeholder for ${key} in index.html.`)
   }, baseHtmlTemplate)
+
+  return html.replace('</head>', `${data.style}</head>`)
```
dcsaszar: * [ ] We now have two ways of modifying the `index.html`, which makes the "correct" approach less clear.

index.html
dcsaszar: * [x] FYI This creates the illusion of an expected order, which Vite doesn't have by default. I suggest to remove the stylesheet link here and import in `main.tsx` to avoid surprises.

dcsaszar (changes requested):
Learnt something new about Vite 🎉 . See comments.

* [x] My alternative suggestion: Tag the overriding vars in `DesignAdjustments` as `!important`. Why does this work?
  1. The idea is to have a default (in CSS) and an override (by the app), `!important` makes this explicit.
  1. The vars are "global", i.e. not intended to be overridden by deeper DOM elements, so `!important` can't hurt nested selectors.

apepper:
REVIEW ([a9cc1e3](https://github.com/Scrivito/scrivito-portal-app/pull/1068/commits/a9cc1e3d4a5874628a4e4c5bd6e223f83d14869d))

FYI: I forced pushed a different version.

src/Components/DesignAdjustments.tsx
```diff
@@ -104,7 +108,7 @@ const FontFace = connect(
             font-family: '${fontFamily}';
             font-display: swap;
             src: url('${encodeURI(fontUrl)}');
-            ${weight ? `font-weight: ${weight};` : ''}
+            ${weight ? `font-weight: ${weight} !important;` : ''}
```
dcsaszar, line 107: * [x] FYI in order to avoid subtle typos etc. by not being DRY enough, we could collect the values in "plain" form and add the `" !important;"` suffix where we join. Update: comment is meant for line 60 and above.

dcsaszar (approved):
LGTM

apepper:
REVIEW ([04b0273](https://github.com/Scrivito/scrivito-portal-app/pull/1068/commits/04b0273a524fa553011b695b0b47490a745ab46f))

dcsaszar (approved):
LGTM

PR #1088, opened 2026-04-24
Tailwind basics

apepper:


src/assets/stylesheets/index.css
```diff
@@ -8,3 +8,23 @@
   --breakpoint-xl: 75rem; /* 1200px */
   --breakpoint-2xl: 87.5rem; /* 1400px — Bootstrap's `xxl` */
 }
+
+/* Use bootstrap 5 container widths */
+@utility container {
```
dcsaszar, line 13: * [ ] FYI I'd prefer a solution that plays along with the Tailwind default instead of overriding it. Ideas:
  * get rid of the added padding
  * use padding classes (we already do this - so maybe just adjust the padding values)
  * use the documented way of default padding: https://v3.tailwindcss.com/docs/container#adding-horizontal-padding

src/assets/stylesheets/index.css
```diff
@@ -8,3 +8,23 @@
...
+  width: 100%;
+  @media (width >= theme(--breakpoint-sm)) {
+    max-width: 540px;
```
dcsaszar, line 16: * [x] Please do not mix rem and px. Otherwise the layout will behave in unexpected ways on screens with non-default rem size. Since the theme recommends rem let's stick with rem.

src/Widgets/SectionWidget/SectionWidgetComponent.tsx
```diff
@@ -12,11 +12,11 @@ provideComponent(SectionWidget, ({ widget }) => {
     sectionClassNames.push(`bg-${backgroundColor}`)
   }
 
-  if (widget.get('showPadding')) sectionClassNames.push('py-5')
+  if (widget.get('showPadding')) sectionClassNames.push('py-12!')
```
dcsaszar, line 15: * [ ] FMI: Why do we need `!`?
* [ ] FYI: Idea: Maybe we should enforce comments for every usage of `!`.

src/assets/stylesheets/index.css
```diff
@@ -28,3 +28,33 @@
     max-width: 1320px;
   }
 }
+
+/* skip-to-content link
+ ================================================== */
+
+.skippy:focus-within {
```
dcsaszar, line 35: * [x] FYI How about deleting styles we moved back from the old file to have some form of progress tracking?

src/assets/stylesheets/index.css
```diff
@@ -30,6 +31,12 @@
   }
 }
 
+body {
```
dcsaszar, line 34: * [x] FYI: Any chance of using the Tailwind approach to set the default font? https://v3.tailwindcss.com/docs/font-family#customizing-the-default-font (custom base style)?

src/assets/stylesheets/index.css
```diff
@@ -30,6 +31,12 @@
...
+  font-family: var(--bs-body-font-family);
```
dcsaszar, line 35: * [ ] Please no `bs-` prefix if we're not using Bootstrap. This is confusing in a Tailwind environment.

src/Widgets/DataImageWidget/DataImageWidgetComponent.tsx
```diff
@@ -51,7 +51,9 @@ const ImageComponent = connect(function ImageComponent({
 }: {
   widget: DataImageWidgetInstance
 }) {
-  const className = widget.get('roundCorners') ? 'rounded' : undefined
+  const className = widget.get('roundCorners')
+    ? 'rounded-(--bs-border-radius)'
```
dcsaszar, line 55: * [ ] Please no `bs-` prefix.

src/assets/stylesheets/index.css
```diff
@@ -31,6 +31,13 @@
   }
 }
 
+img,
+video {
+  display: inline-block;
+  max-width: 100%;
+  object-fit: contain;
```
dcsaszar, line 38: * [ ] FYI Could we try using @layer base and @apply as much as possible in order to be more Tailwindy? https://stackoverflow.com/a/74432943

dcsaszar (changes requested):
See comments. I'm also new to Tailwind so my suggestions may need further research.

src/assets/stylesheets/index.css
apepper: I'm confused. The `@utility container {` section is about widths of containers in different breakpoints. WHat does this have to do with padding?

Also your pointer to default padding links to an older version of tailwind. It changed with the recent v4 release: https://tailwindcss.com/docs/upgrade-guide#container-configuration

src/Widgets/SectionWidget/SectionWidgetComponent.tsx
apepper: The `!` makes `padding-block: calc(var(--spacing) * 12) !important;`. Without `!` it is just `padding-block: calc(var(--spacing) * 12);`.

Without the `!important` the padding in edit mode will be overwritten by `scrivito_editing.css` with:
```css
[data-scrivito-editing] [data-scrivito-private-structure-widget] {
    padding-top: 20px;
    padding-bottom: 20px
}
```

To my knowledge bootstrap used `!important`.

src/assets/stylesheets/index.css
apepper: I get your point. I want to normalizes this, once I introduce the other `:root` variables, that are color related (e.g. `--bs-success`).

apepper:
Closing for now and will reapply (in smaller chunks).

PR #1089, opened 2026-04-24
Replace Sass bestContrast with PostCSS plugin

dcsaszar:
This removes a dependency on Sass.

FYI `postcss` is a dependency of Vite anyway.


postcss.config.ts
```diff
@@ -0,0 +1,47 @@
...
+          color-mix(in srgb, ${liteText} 0%,
+            color(from ${bg} xyz none none none / calc(1 - round(y + ${PIVOT_Y_OFFSET})))))
+        srgb r g b / 1)`
+    },
+  }
+}
+
+bestContrast.postcss = true
+
+export default { plugins: [bestContrast()] }
```
apepper, line 47: FMI: This works for vite out of the box: https://vite.dev/guide/features#postcss.

package.json
```diff
@@ -67,6 +67,7 @@
     "jsontoxml": "^1.0.1",
     "knip": "^6.3.1",
     "local-web-server": "^5.4.0",
+    "postcss": "^8.5.12",
```
apepper, line 70: FYI: I resolve the conflict already.

tsconfig.node.json
```diff
@@ -5,5 +5,5 @@
     "module": "ESNext",
     "moduleResolution": "bundler"
   },
-  "include": ["vite.config.ts", "headers.config.ts"]
+  "include": ["vite.config.ts", "headers.config.ts", "postcss.config.ts"]
```
apepper, line 9: It looks like, that `npm run types` does _not_ check the types of `postcss.config.ts`. E.g. if I write `return null` within `bestContrast` I can see my editor complaining, but `npm run types` does not fail.

apepper (changes requested):
FEEDBACK ([3946e47](https://github.com/Scrivito/scrivito-portal-app/pull/1089/commits/3946e472f563ae11a74eca797d616a211c6a554a))

Nice - thank you for looking into this. I only found one smaller issue - see below.

.gitignore
```diff
@@ -14,6 +14,7 @@ dist-ssr
 *.local
 *.local.json
 .lighthouseci
+*.tsbuildinfo
```
apepper: I would prefer to store the file in a "hidden" place.

E.g. add the following to tsconfig.node.json:

```ts
"tsBuildInfoFile": "node_modules/.cache/tsconfig.node.tsbuildinfo"
```

tsconfig.node.json
apepper: With [470bea7](https://github.com/Scrivito/scrivito-portal-app/pull/1089/commits/470bea716e67b22c0d28e38d03a8b71f40eed941) `npm run types` still passes, without complaining.

apepper (commented):
FEEDBACK ([470bea7](https://github.com/Scrivito/scrivito-portal-app/pull/1089/commits/470bea716e67b22c0d28e38d03a8b71f40eed941))

`npm run types` still does not "catch" typescript errors in postcss.config.ts

tsconfig.node.json
dcsaszar: If you make an actual type error in `postcss.config.ts` (something flagged red by the IDE) `npm run types` fails. At least it does for me. Am I missing some edge case?

apepper: The following complains in VS-Code, but not with `npm run types`:

```ts
import type { PluginCreator } from 'postcss'

/**
 * PostCSS plugin to expand best-contrast() function calls.
 *
 * Transforms: best-contrast($bg, $darkText, $liteText)
 * Into: Full CSS calculation using XYZ color space luminance
 */
const bestContrast: PluginCreator<void> = () => {
  // Pivot point for luminance-based text color selection.
  // Backgrounds #a3a3a3 (Y = 0.366253) and #a3a3a4 (Y = 0.366613) were empirically
  // determined as the flip point between black/white text using https://colorcontrast.app/
  const PIVOT_Y = (0.366253 + 0.366613) / 2
  const PIVOT_Y_OFFSET = 0.5 - PIVOT_Y

  return null

  // ...
}

bestContrast.postcss = true

export default { plugins: [bestContrast()] }

```

Interestingly, removing e.g. `postcssPlugin: 'postcss-best-contrast',` from the correct code does make `npm run types` fail.

apepper: I would say: good enough.

apepper (approved):
LGTM ([4cf505e](https://github.com/Scrivito/scrivito-portal-app/pull/1089/commits/4cf505e1ad35597bebd5a21459df121c576cdbe6))

Nice 🚀! Thank you for your help.

PR #1092, opened 2026-04-27
http-proxy-agent: Force update of @tootallnate/once

apepper:
Please note, that there are several versions of http-proxy-agent in use (7.0.2 and 5.0.0), but only version 5.0.0 uses "@tootallnate/once". I tried to scope the overwrite to `http-proxy-agent@5` and `http-proxy-agent@5.0.0`, but npm (version 11.12.1) had trouble to follow the instruction correctly.

Fixes https://github.com/Scrivito/scrivito-portal-app/security/dependabot/60.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "vite-node": "^6.0.0"
   },
   "overrides": {
+    "@tootallnate/once": "^3.0.1",
     "@lhci/cli": {
       "cookie": "^0.7.0",
       "rimraf": "4",
--- a/package-lock.json
+++ b/package-lock.json
@@ -3403,9 +3403,9 @@
       }
     },
     "node_modules/@tootallnate/once": {
-      "version": "2.0.0",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -75,6 +75,9 @@
     "vite-node": "^6.0.0"
   },
   "overrides": {
+    "http-proxy-agent": {
+      "@tootallnate/once": "^3.0.1"
+    },
```
dcsaszar: * [x] FYI Since we only have one dependency path anyway and cannot scope to `local-web-server`, how about a global override `"@tootallnate/once": "3"`?

dcsaszar (approved):

package.json
apepper: I did so and forced pushed the change.

apepper:
REVIEW ([29ad947](https://github.com/Scrivito/scrivito-portal-app/pull/1092/commits/29ad94709aa852eabc395719f38bcbcc133f074d))

dcsaszar (approved):

PR #1094, opened 2026-04-28
Use strict-dynamic CSP

dcsaszar:


src/prerenderContent/prerenderObj.tsx
```diff
@@ -23,20 +22,15 @@ export async function prerenderObj(
...
       style: helmet?.style.toString() || '',
       title: helmet?.title.toString() || '',
     }
   })
 
-  const preloadDumpFileContent = generatePreloadDump(preloadDump)
-  const preloadDumpContentHash = await contentHash(preloadDumpFileContent)
-  const preloadDumpFileName = `/assets/preloadDumps/${objId}.${preloadDumpContentHash}.js`
-  const preloadDumpScript = `<script type="module" src="${preloadDumpFileName}"></script>`
+  const preloadDumpScript = generatePreloadDumpScript(preloadDump)
```
apepper, line 31: Have you benchmarked this change? I don't remember specifically, but I think back in the days we decided against inlining the preload dump into the html.

src/prerenderContent/generateHtml.ts
```diff
@@ -19,9 +19,9 @@ export async function generateHtml(
 }
 
 const replacements: Record<(typeof keys)[number], RegExp> = {
-  bodyAttributes: /data-bodyAttributes/,
+  bodyAttributes: /data-body-attributes/,
   bodyContent: /\n *<!-- bodyContent start -->[^!]*<!-- bodyContent end -->\s*/,
```
apepper: I would prefer to use `kebab-style-casing` through out the replacements. So `body-content`, etc.

vite.config.ts
```diff
@@ -149,13 +150,31 @@ function scrivitoOrigin(env: Record<string, string>) {
...
+
+    buildStart() {
+      inlineScriptHashes = []
+    },
+
+    transformIndexHtml: {
+      order: 'post' as const,
+      handler(html: string) {
+        const strictCsp = new StrictCsp(html)
+        strictCsp.refactorSourcedScriptsForHashBasedCsp()
```
apepper: Where can I find documentation on `refactorSourcedScriptsForHashBasedCsp`? I tried https://github.com/search?q=repo%3Agoogle%2Fstrict-csp%20refactorSourcedScriptsForHashBasedCsp&type=code and couldn't find anything.

index.html
```diff
@@ -16,7 +16,7 @@
     <!-- preloadDumpScript -->
```
apepper: Now that `preloadDumpScript` contains no longer a link to js, but the whole dump, please move it _below_ or at the end of the body, so the html can already be displayed, while the html file is not yet fully loaded.

src/prerenderContent/generatePreloadDump.ts
```diff
@@ -1,13 +0,0 @@
...
-
-/**
- * JSON may contain line separators, which are invalid in string literals in older browsers.
- * See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#issue_with_plain_json.stringify_for_use_as_javascript
- * */
-function stringLiteral(string: string) {
-  return JSON.stringify(string)
-    .replace(/\u2028/g, '\\u2028')
-    .replace(/\u2029/g, '\\u2029')
-}
```
apepper, line 13: Is this still relevant? If I'm not mistaken the new code no longer normalizes the affected unicode chars.

headers.config.ts
```diff
@@ -49,14 +30,59 @@ function headers(environment: string) {
...
+      'https://*.etracker.com',
+      'https://*.etracker.de',
+      'https://api.scrivito.com',
+      'https://assets.scrivito.com',
+      // The package `@vitejs/plugin-react-swc` include an inline script into index.html (see [1]).
+      // [1] https://github.com/vitejs/vite-plugin-react/blob/7517103485081b26004e79f169efdd2d12a60946/packages/common/refresh-utils.ts#L7-L12
+      // In case it breaks please look into the JS console and search for "CSP".
+      // There you can find the current "sha256-x" value, which is to be copied over here.
+      "'sha256-Z2/iFzh9VMlVkEOar1f/oSHWwQk3ve1qk/C2WdsC4Xk='",
+    ],
```
apepper: Do I understand this correctly, that during development we still rely on the "old" URL-whitelist based approach? I think this will eventually bid us, e.g. if etracker uses more "advanced use-cases" with adding scripts later on.

Ideally both prod and development should use a sha based CSP rule.

apepper (commented):
FEEDBACK ([f6a70da](https://github.com/Scrivito/scrivito-portal-app/pull/1094/commits/f6a70daffe0c06a5e37b08f25a972d29ad5b29e1))

Nice start 🚀! Please see my comments for questions or potential issues.

src/prerenderContent/prerenderObj.tsx
dcsaszar: I rely on `npm run lighthouse` and see no changes there.

src/prerenderContent/generatePreloadDump.ts
dcsaszar: No longer relevant as we no longer do the `JSON.stringify`/`JSON.parse` dance, but store the string as `text/plain`.

headers.config.ts
dcsaszar: For this to work, we'd need a vite dev mode that is strict-dynamic compatible. I haven't done any research if this is possible at all. I'd suggest to make this a follow-up task.

apepper: Please at least research if it's possible to run vite dev with `strict-dynamic` headers possible. Otherwise I'm hesitant to merge `strict-dynamic` in the first place.

vite.config.ts
dcsaszar: https://www.npmjs.com/package/strict-csp#example-usage

apepper: Thanks for the link.

I dug a little deeper. It looks like, that `refactorSourcedScriptsForHashBasedCsp` was removed in the PR https://github.com/google/strict-csp/pull/69, but the package has not yet been released after the merge - around 8 months later.

I'm not so sure, if we should trust `strict-csp` with its experimental status. https://www.npmjs.com/package/vite-plugin-sri-gen is probably a more promising option.

apepper (changes requested):
FEEDBACK ([f6a70da](https://github.com/Scrivito/scrivito-portal-app/pull/1094/commits/f6a70daffe0c06a5e37b08f25a972d29ad5b29e1))

apepper (approved):
LGTM (8669d1981b4d6f40df31b55ddbecd2ead601c1c8)

Nice 🚀!

PR #1096, opened 2026-04-29
Tailwind: (Data)ImageWidget and custom radius

apepper:


src/assets/stylesheets/theme/box-comment.scss
```diff
@@ -6,7 +6,7 @@
     min-width: 76px;
     img {
       width: 60px;
-      border-radius: var(--bs-border-radius);
+      border-radius: var(--radius-jr);
       box-shadow: var(--jr-box-shadow);
```
dcsaszar: * [x] FYI since we're establishing a new naming convention anyway, we could use something less cryptic than `--radius-jr`. E.g. the more contextual `--radius-portal` or semantic `--radius-primary`.

dcsaszar (approved):
LGTM. See comment for naming feedback.

PR #1105, opened 2026-05-07
Convert contentDump/index.json to new objs manifest format

patrykboch:
Replaces the legacy `{ objIds: [...] }` manifest with the new `{ objs: [{ id, siteId?, contentId?, path? }] }`. 

Is the target branch correct? Indeed, this highlights the difference between `v6-content`, but it probably needs to be separated. We can create & change the target before merging.

See: https://github.com/infopark/scrivito_js/pull/13002

### DIFF
--- a/contentDump/index.json
+++ b/contentDump/index.json
@@ -1,190 +1,1004 @@
 {
-  "objIds": [
-    "fec151165aeb27da",
-    "fd573c40f9c733ef",
-    "fd456e1a38e85cfb",
-    "fcfbdded110d9f59",
-    "fb302b23c46fea41",
-    "fb161c52690aeb71",
-    "fa725f0b47fdc408",
-    "f909e9691db27360",
-    "f7da3de531e46fe9",
-    "f70ac27424b4ba20",
-    "f6f521cad64ae778",
-    "f688f1b1b19451ae",
-    "f5993d6de3609ac0",
-    "f35e3741a5662ea5",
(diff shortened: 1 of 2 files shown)


contentDump/index.json
```diff
@@ -1,190 +1,1004 @@
...
-    "0dda554dfe67da45",
-    "09a152b36247dc0a",
-    "060918d8e44ae54f",
-    "0513c02cb152a069",
-    "046cc19cf0f87d9d",
-    "03cd0d7704a1195a",
-    "02d09ad2dd7d2332",
-    "027be0e907d46d8e",
-    "0092ff67875b69f8"
+  "objs": [
```
apepper, line 2: I don't see changes to `src/privateJrPlatform/dumpContent.ts`. Where are they? Ideally have two commits: One with updating `dumpContent.ts` to generate the new format and one with executing the new script.

apepper (changes requested):
FEEDBACK ([60dc21e](https://github.com/Scrivito/scrivito-portal-app/pull/1105/commits/60dc21e1e428241ea831eed4d586a64d2388ee31))

I only took a rough look at the PR. Please see my comments.

apepper (approved):
LGTM ([c7e45ff](https://github.com/Scrivito/scrivito-portal-app/pull/1105/commits/c7e45ffeee0852583d97a4d7114c9fc5d83b6bb5))

Good stuff 🚀!

PR #1107, opened 2026-05-08
Tailwind primary and secondary color

apepper:
This PR is only a first step for the full color support. It focused on primary and secondary for now. But I would like to get your feedback and hopefully get this in.

### DIFF
--- a/src/assets/stylesheets/index.css
+++ b/src/assets/stylesheets/index.css
@@ -17,7 +17,9 @@
 
 :root {
   --dark-text-color: #454545;
+  --dark-headline-text-color: #3b3b3b;
   --light-text-color: #fff;
+  --light-headline-text-color: #fff;
 }
 
 @utility bg-portal-primary {
@@ -27,6 +29,11 @@
     var(--dark-text-color),
     var(--light-text-color)
   );
+  --headline-text-color: best-contrast(
+    var(--color-portal-primary),
+    var(--dark-headline-text-color),
+    var(--light-headline-text-color)
+  );
 }
 
 @utility bg-portal-secondary {
@@ -35,6 +42,11 @@
     var(--color-portal-secondary),
     var(--dark-text-color),
     var(--light-text-color)
+  );
+  --headline-text-color: best-contrast(
+    var(--color-portal-secondary),
+    var(--dark-headline-text-color),
+    var(--light-headline-text-color)
   );
 }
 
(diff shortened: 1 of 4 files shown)


src/assets/stylesheets/index.css
```diff
@@ -15,6 +15,29 @@
   --color-portal-secondary: #39a9eb;
 }
 
+:root {
+  --dark-text-color: #454545;
+  --light-text-color: #fff;
+}
+
+@utility bg-portal-primary {
+  background-color: var(--color-portal-primary);
```
dcsaszar: * [ ] This repeats what Tailwind already provides out of the box. We should Tailwind let do the job. Same for `bg-portal-secondary` background color.

src/assets/stylesheets/index.css
```diff
@@ -15,6 +15,29 @@
...
+  color: best-contrast(
+    var(--color-portal-primary),
+    var(--dark-text-color),
+    var(--light-text-color)
+  );
```
dcsaszar: * [ ] Not sure about coupling the fg color to the bg color class. Tailwind appears to separate bg and text colors as it provides distinct auto-generated classes for each. How about we specify these together: `bg-portal-primary text-portal-primary-foreground`. It would then be sufficient to provide a variable like so:
* [ ] The contrast color definition should be part of the theme, e.g.:
```css
@theme {
  /* .. */
  --color-portal-primary-foreground: best-contrast(
    var(--color-portal-primary),
    var(--dark-text-color),
    var(--light-text-color)
  );
}
```

src/assets/stylesheets/index.css
```diff
@@ -27,6 +29,11 @@
     var(--dark-text-color),
     var(--light-text-color)
   );
+  --headline-text-color: best-contrast(
```
dcsaszar: * [ ] A cleaner way to put this in `@theme`: define `--color-portal-primary-headline`, then use classname `text-portal-primary-headline`.
* [ ] FYI Suggestion from ChatGPT: 
  > semantic layering (cleaner long-term)
  > If you expect more complexity, separate surface roles:
```css
@theme {
  --color-surface-primary: /* ... */;
  --color-on-primary: /* ... */;
  --color-on-primary-strong: /* ... */;
}
```
```html
<section class="bg-surface-primary text-on-primary">
  <h1 class="text-on-primary-strong">
</section>
```

dcsaszar (changes requested):
Looks good so far. Still getting the hang of how Tailwind does things.

dcsaszar (approved):
The separation into static theme color and context-dependent colors now LGTM

PR #1112, opened 2026-05-18
Tailwind: More background colors

apepper:
Building on the same pattern of https://github.com/Scrivito/scrivito-portal-app/pull/1107 this PR add the remaining missing background colors.

It also harmonizes the backgroundColor implementation between all widgets. Please note, that only the background color aspect was addressed. Other aspects of the touched widgets probably need more tailwind migrations.

Also the Slide/Slider Widget was left out for now. They will be part of a bigger refactoring in a later PR.

The duplicate `backgroundClassName` functions are on purpose: One widget's component should (ideal) not rely on shared code.


agessler (approved):
👍

src/Objs/defaultPageLayoutComponent.tsx
```diff
@@ -55,9 +55,42 @@ const BackgroundWrapper = connect(function BackgroundWrapper({
...
+    case 'warning':
+      return 'bg-portal-warning text-on-portal-warning'
+    case 'danger':
+      return 'bg-portal-danger text-on-portal-danger'
+    case 'transparent':
+      return ''
+    default:
+      throw new Error(`Unknown backgroundColor: ${backgroundColor}`)
+  }
+}
```
markus-codewrangler, line 92: - [ ] transparent case is dead code – `BackgroundWrapper` already returns early before calling this function ~~(same for other components)~~
- [ ] FYI simplify (same for other components):

```suggestion
  function backgroundClassName(backgroundColor: string): string {
    const validColors = [
      'primary', 'secondary', 'white', 'light-grey', 'middle-grey',
      'dark-grey', 'success', 'info', 'warning', 'danger',
    ]
    if (!validColors.includes(backgroundColor)) {
      throw new Error(`Unknown backgroundColor: ${backgroundColor}`)
    }
    return `bg-portal-${backgroundColor} text-on-portal-${backgroundColor}`
  }
```

src/assets/stylesheets/index-old.scss
markus-codewrangler, line 1: FYI Claude says (please verify):

> PR deletes the .bg-primary, .bg-secondary, .bg-white, .bg-light-grey, .bg-danger CSS rules from index-old.scss, but these class names are still hardcoded in: ProductsOverviewComponent.tsx, ProductCategoryComponent.tsx, ProductPreviewComponent.tsx, ProductComponent.tsx, ErrorBoundary.tsx, JrPlatformMissingInstanceId.tsx, JrPlatformWrongContentFormat.tsx, DataPersonCardWidgetComponent.tsx, DataMessageWidgetComponent.tsx, DownloadCardWidgetComponent.tsx, MetaNavigation.tsx

src/Objs/defaultPageLayoutComponent.tsx
```diff
@@ -55,9 +55,42 @@ const BackgroundWrapper = connect(function BackgroundWrapper({
...
 
-  return <section className={`bg-${backgroundColor}`}>{children}</section>
+  return (
+    <section className={backgroundClassName(backgroundColor)}>
+      {children}
+    </section>
+  )
 })
 
+function backgroundClassName(backgroundColor: string): string {
```
markus-codewrangler, line 65: > The duplicate backgroundClassName functions are on purpose: One widget's component should (ideal) not rely on shared code.

FYI AFAICS we share many symbols across modules to avoid redundancy and deviations. so I would prefer DRYing this up 🙂

src/Widgets/SectionWidget/SectionWidgetComponent.tsx
```diff
@@ -51,3 +43,32 @@ provideComponent(SectionWidget, ({ widget }) => {
...
+      return 'bg-portal-info text-on-portal-info'
+    case 'warning':
+      return 'bg-portal-warning text-on-portal-warning'
+    case 'danger':
+      return 'bg-portal-danger text-on-portal-danger'
+    case 'transparent':
+      return ''
+    default:
+      throw new Error(`Unknown backgroundColor: ${backgroundColor}`)
+  }
```
dcsaszar, line 73: * [ ] FYI The old code used string interpolation, which is better for multiple reasons:
  * Less code, which is easier to read
  * Easier to extend (add/remove enum values)
  * No exception handling needed

src/assets/stylesheets/index.css
```diff
@@ -108,6 +204,38 @@
     --headline-text-color: var(--color-on-portal-secondary-heading);
   }
 
+  :where(.text-on-portal-white) {
+    --headline-text-color: var(--color-on-portal-white-heading);
```
dcsaszar, line 208: * [ ] FYI Not introduced here but the naming ambiguity heading/headline could be eliminated for more clarity

src/Widgets/SectionWidget/SectionWidgetComponent.tsx
apepper: String interpolation will not work with Tailwind. Tailwind needs to see the whole string. See https://tailwindcss.com/docs/detecting-classes-in-source-files#dynamic-class-names for more details.

dcsaszar: I see. Unfortunately this breaks the separation of concerns of widget config/rendering.
How about safelisting all `-portal-` names in the Tailwind config?

```ts
  safelist: [
    {
      pattern: /-portal-/,
    },
  ],
```

apepper: > Unfortunately this breaks the separation of concerns of widget config/rendering.

Can you elaborate? Which separation do you mean?

apepper: So far I'm not convinced of the safelist approach. The "tailwind has to read the whole string" rule is something that I think many devs will shortly become familiar if they start using "tailwind". On the contrary adding `safelist` for some class names could lead to more confusion: E.g. if the developer sees, that some class names are interpolated and tries the same with some other class names it will not work, if the other class name is not listed in the safelist.

src/assets/stylesheets/index-old.scss
apepper: Claude is correct. Those other components will be addressed in other PRs. The focus of this PR was to introduce the new colors in tailwind.

src/Objs/defaultPageLayoutComponent.tsx
apepper: Please continue the discussion regarding interpolated strings here: https://github.com/Scrivito/scrivito-portal-app/pull/1112#discussion_r3287034950

apepper: Regarding `'transparent'` being dead code: I only found this to be the case for `DefaultPageLayoutComponent`. Are there other components, that have the same issue?

src/Widgets/SectionWidget/SectionWidgetComponent.tsx
markus-codewrangler: continuing [discussion](https://github.com/Scrivito/scrivito-portal-app/pull/1112#discussion_r3287535003):

> String interpolation will not work with Tailwind. Tailwind needs to see the whole string.

ok, I see💡

but instead of the verbose `switch/case/return`, can we use a simple object (as the tailwind docs propose)?

```ts
  const colorVariants = {
    blue: "bg-blue-600 hover:bg-blue-500",
    red: "bg-red-600 hover:bg-red-500",
  };
```

src/Objs/defaultPageLayoutComponent.tsx
markus-codewrangler: > I only found this to be the case for `DefaultPageLayoutComponent`

correct, sorry! 👍

src/assets/stylesheets/index.css
apepper: => https://github.com/Scrivito/scrivito-portal-app/pull/1120

src/Widgets/SectionWidget/SectionWidgetComponent.tsx
apepper: With that the function would look like this:

```ts
const backgroundColorClassNames: Record<string, string> = {
  primary: 'bg-portal-primary text-on-portal-primary',
  secondary: 'bg-portal-secondary text-on-portal-secondary',
  white: 'bg-portal-white text-on-portal-white',
  'light-grey': 'bg-portal-light-grey text-on-portal-light-grey',
  'middle-grey': 'bg-portal-middle-grey text-on-portal-middle-grey',
  'dark-grey': 'bg-portal-dark-grey text-on-portal-dark-grey',
  success: 'bg-portal-success text-on-portal-success',
  info: 'bg-portal-info text-on-portal-info',
  warning: 'bg-portal-warning text-on-portal-warning',
  danger: 'bg-portal-danger text-on-portal-danger',
  transparent: '',
}

function backgroundClassName(backgroundColor: string): string {
  const className = backgroundColorClassNames[backgroundColor]
  if (className === undefined)
    throw new Error(`Unknown backgroundColor: ${backgroundColor}`)
  return className
}
```

@dcsaszar What do you think?

dcsaszar: Yes, this approach is also implied as a best practice by the docs https://tailwindcss.com/docs/detecting-classes-in-source-files?utm_source=chatgpt.com#dynamic-class-names, see the `colorVariants` example.

dcsaszar: > > Unfortunately this breaks the separation of concerns of widget config/rendering.
> 
> Can you elaborate? Which separation do you mean?

The Scrivito config keeps the actual enum values in the widget config modules.

So far, the component could stay “dumb” and simply render sanitized values without knowing anything about styling details. Changes in the config did not necessarily affect rendering logic.

Now, an explicit mapping from enum values to Tailwind class names is required so that all possible classes are statically detectable at build time. This introduces a new concern: a presentation mapping must exist as literal strings somewhere in the codebase.

If this mapping lives in the component, it couples domain configuration with presentation logic. Changes in the config will require corresponding changes in the component.

To untangle this, the mapping should be moved into a dedicated module, rather than being placed in the component.

How about `WidgetNamePresentation.ts`, so each concern has a single "axis of change" again?

apepper: I would like to stick to the best practice of Tailwind, regarding the object lookup.

Regarding "api surface": To me this is a bit arbitrary to try to reduce the coupling of a widget data class with the component layer. Both need to exists and to me it is accidental, that they previously didn't have a "mapping function". This becomes even more clearer, if you look at the e.g. `margin` of `CardWidget`. So far it defined an enum with the attributes `'mb-0', 'mb-1', 'mb-2', 'mb-3', 'mb-4', 'mb-5'`. All of them are "bootstrap style" margin bottom. But tailwind scales differently. So `margin` needs a mapping from bootstrap to tailwind, if we don't want to change the data layer (which I definitely don't want to do).

Regarding "shared function": The overall goal for widgets/objs is to be "self reliant", to allow "moving widgets" between projects. So a mapping from `mb-5` (bootstrap) to `mb-12` (tailwind) would need to be done in every widget/component. But for the `backgroundClassName` I think an exception would be acceptable, since the css class names (e.g. `bg-portal-primary`) are defined globally. That is why I created https://github.com/Scrivito/scrivito-portal-app/pull/1122.

src/Objs/defaultPageLayoutComponent.tsx
apepper: FYI: Indirectly addressed by #1122.

PR #1115, opened 2026-05-20
Tailwind: CSS: Convert jr-box-shadow to shadow-portal

apepper:
Future use: `shadow-portal`, e.g. for CardWidget.

### DIFF
--- a/src/assets/stylesheets/index.css
+++ b/src/assets/stylesheets/index.css
@@ -11,6 +11,8 @@
 
   --radius-portal: 8.5px;
 
+  --shadow-portal: 0 0.375rem 1.5rem 0 rgba(140, 152, 164, 0.25);
+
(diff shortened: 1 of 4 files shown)


agessler (approved):
👍

src/assets/stylesheets/index.css
```diff
@@ -11,6 +11,8 @@
 
   --radius-portal: 8.5px;
 
+  --shadow-portal: 0 0.375rem 1.5rem 0 rgba(140, 152, 164, 0.25);
```
dcsaszar, line 14: * [ ] FYI not introduced here: I just realized this color is a hardcoded function of blue. Maybe we should make it a function of the configured primary color.

apepper: I consider this more a nice to have.

PR #1116, opened 2026-05-20
Tailwind: CardWidget + ImageOrVideo

apepper:
I inlined `card-extended` into the only caller.

`.card` and `.card-body`  are also used by other components. For completeness-sake I added `.card-footer` also to `index.css`, even though it is only used by `CardWidgetComponent`.


src/Widgets/CardWidget/CardWidgetComponent.tsx
```diff
@@ -16,20 +16,30 @@ import { useRef } from 'react'
 
 provideComponent(CardWidget, ({ widget }) => {
   const cardBodyClassNames: string[] = ['card-body']
-  const padding = widget.get('padding')
-  cardBodyClassNames.push(padding ? padding : 'p-4')
+  const bootstrapPadding = widget.get('padding') || 'p-4'
```
dcsaszar: * [x] FYI I find it unnecessarily confusing to have the extra translation/renaming step to Bootstrap. So instead of `padding` -> `bootstrapPadding` -> `tailwindPaddingFor` -> `className`, how about `padding` -> `tailwindClassNameForPadding` -> `className`, i.e. treat the padding values as Portal App internal, without mentioning Bootstrap?

src/Widgets/CardWidget/CardWidgetComponent.tsx
```diff
@@ -16,20 +16,30 @@ import { useRef } from 'react'
...
+  if (backgroundColor === 'transparent') cardClassNames.push('shadow-none')
 
-  if (widget.get('cardExtended')) cardClassNames.push('card-extended')
+  if (widget.get('cardExtended')) {
+    cardClassNames.push(
+      'overflow-visible',
+      "after:rounded-portal after:pointer-events-none after:absolute after:top-0 after:bottom-0 after:-z-[1] after:block after:bg-inherit after:content-['']",
+      '[*:first-child_&]:after:right-0 [*:first-child_&]:after:left-[-2600px]',
+      '[*:last-child_&]:after:right-[-2600px] [*:last-child_&]:after:left-0',
+      '[*:only-child_&]:after:right-0 [*:only-child_&]:after:left-0',
```
dcsaszar: > I inlined card-extended into the only caller.

* [x] FYI This makes these complex styles hard to impossible to read. I would prefer keeping them extracted to a named class. But since it is single-use, to would go to the widget-local CSS.

src/assets/stylesheets/index.css
```diff
@@ -170,6 +170,53 @@
   }
 }
 
+/* Image / Video background
```
dcsaszar, line 174: * [ ] FYI `:has` and `:is` are smells. In order to make this more "tailwindy" and better maintainable, we need to clean up (maybe a follow-up). Instead of relying on DOM order and shape, we should prefer explicit class names with a clear function, so we can, for example, write `<section class="has-img-background has-muted-background zoom-on-hover">`.

src/Components/ImageOrVideo.tsx
```diff
@@ -65,6 +64,16 @@ export const ImageOrVideo = connect(function ImageOrVideo<T extends string>({
   if (className) classNames.push(className)
 
   if (background.contentType().startsWith('video/')) {
+    const playButtonClassNames = [
+      'absolute top-1/2 left-1/2 -translate-x-1/2 -translate-y-1/2',
+      'flex h-[50px] w-[50px] items-center justify-center',
+      'rounded-full border-none bg-black/40 shadow-[0_0_50px_5px_rgba(255,255,255,0.8)]',
+      'z-[1111] cursor-pointer transition-all duration-[400ms] ease-in-out',
+      'hover:scale-[1.3] hover:bg-black/60',
```
dcsaszar: * [x] FYI This is too complex to understand inline, and would be better extracted.

dcsaszar (approved):
Works but see comments for avoiding maintenance hell in the long run.

src/assets/stylesheets/index.css
apepper: Feel free to create a follow-up PR. I would first focus on migrating to tailwind all together for now.

PR #1114, opened 2026-05-19
Tailwind: ColumnWidget

apepper:
Goal: Render ColumnWidget with modern Tailwind.

PR #1121, opened 2026-05-22
Tailwind: Scrivito extensions

apepper:
Adresses https://github.com/Scrivito/scrivito-portal-app/pull/1114#discussion_r3287180245.

### DIFF
--- a/src/Components/ScrivitoExtensions/SocialCardsTab.css
+++ b/src/Components/ScrivitoExtensions/SocialCardsTab.css
@@ -1,12 +1,12 @@
-// -- forms -- //
-
-$red: #c04e4e;
-$yellow: #ff9f00;
-$yellowDarken: #cc7f00;
-$blue: #00addf;
-$blueDarken: #0085ac;
+/* -- forms -- */
 
 .social-cards-tab {
+  --sc-red: #c04e4e;
+  --sc-yellow: #ff9f00;
+  --sc-yellow-darken: #cc7f00;
+  --sc-blue: #00addf;
+  --sc-blue-darken: #0085ac;
+
   .content_property_input {
     background: #f3f3f3;
     border-radius: 3px;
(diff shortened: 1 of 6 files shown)


src/Components/ScrivitoExtensions/SocialCardsTab.css
```diff
@@ -27,44 +27,44 @@ $blueDarken: #0085ac;
     word-break: break-all;
   }
   .content_property_input:focus {
-    box-shadow: 0 0 0 2px #577fb6 inset !important;
+    box-shadow: 0 0 0 2px #577fb6 inset;
```
dcsaszar, line 30: * [x] FYI The selector is broken (already in main). It should be `:focus-within`.

src/Components/ScrivitoExtensions/SocialCardsTab.tsx
```diff
@@ -17,22 +17,22 @@ export function SocialCardsTab({ obj }: { obj: Obj }) {
   return (
     <div className={`social-cards-tab scrivito_${theme}`}>
       <div className="scrivito_detail_content">
-        <div className="row">
-          <div className="col-sm-6">
+        <div className="flex flex-wrap gap-6">
```
dcsaszar: * [x] This gap is used in the properties dialog. The gap in the sidebar is now wider than the original. `gap-2` is better for the sidebar.

src/Components/ScrivitoExtensions/SocialCardsTab.tsx
```diff
@@ -17,22 +17,22 @@ export function SocialCardsTab({ obj }: { obj: Obj }) {
...
+          <div className="w-full sm:w-[calc(50%-0.75rem)]">
```
dcsaszar: * [x] Not needed if we set up the row with `grid grid-cols-1 gap-6 sm:grid-cols-2` or `grid grid-cols-1 gap-2 sm:grid-cols-2`.
* [ ] Not sure how we should handle the different gap in dialog vs sidebar?

dcsaszar (changes requested):
See comments. Also
* [ ] The card previews are missing the dropshadow
* [ ] The font in general is heavier than in the original (and in the UI in general). Unclear which font setting causes this.

PR #1122, opened 2026-05-26
Tailwind: Extracted backgroundClassName to utils

apepper:
Also converted to tailwind best practice (see https://tailwindcss.com/docs/detecting-classes-in-source-files#dynamic-class-names)

Follow up of https://github.com/Scrivito/scrivito-portal-app/pull/1112#pullrequestreview-4343895121.


src/utils/theme/backgroundClassName.ts
dcsaszar, line 1: * [ ] FYI Idea: instead of adding to the unspecific utils folder, maybe we could treat design-system related helpers are a good candidate for a more specific folder. E.g. `src/theme`.

dcsaszar (approved):
LGTM

PR #1134, opened 2026-06-26
Allow content description and thumbnail

dcsaszar:
These will be used to visually enrich the "Add site from template" functionality.

[Concept (internal)](https://docs.google.com/document/d/1KoUBnifBIFLEBF6FxqoDSpjYffvAAgAF0vRQxophn5Y)


src/Objs/Homepage/HomepageEditingConfig.ts
```diff
@@ -121,10 +121,14 @@ provideEditingConfig(Homepage, {
   },
   propertiesGroups: (site) => [
     ...defaultPagePropertiesGroups,
+    {
+      title: 'Site profile',
```
markus-codewrangler: FYI "profile" vs. "settings": what's the difference? As a reader, I wouldn't be able to tell which prop belongs to which. can we find a more distinctive wording (I don't have any idea ATM, sorry)? or better keep all together in settings?

src/Objs/Homepage/HomepageEditingConfig.ts
```diff
@@ -17,6 +17,13 @@ provideEditingConfig(Homepage, {
   title: 'Homepage',
   attributes: {
     ...defaultPageEditingConfigAttributes,
+    contentDescription: {
```
markus-codewrangler, line 20: FYI please reference the API concept in the PR description

markus-codewrangler (approved):
LGTM.

Some FYIs for consideration.

src/Objs/Homepage/HomepageEditingConfig.ts
dcsaszar: I see the issues. The idea was to
* avoid bloating the site settings (idea: site settings have an effect on the actual site)
* keeping editing-only settings separate (which only affect site editing/creation)

How about:
1. Move description and thumbnail into a new group "Template appearance", or
2. Hide the new fields from editors, as they are only to be used in very special circumstances (creating a template) anyway? - OTOH they will be part of the created site, including the linked image, except if we refine the concept and don't copy them

markus-codewrangler: > Move description and thumbnail into a new group "Template appearance"

sounds good. maybe even "template" is sufficient as a name?

markus-codewrangler (approved):

PR #1139, opened 2026-07-03
Migrate ESLint config to flat config

dcsaszar:


eslint.config.mjs
```diff
@@ -1,50 +1,33 @@
...
-import path from 'node:path'
-import { fileURLToPath } from 'node:url'
-import js from '@eslint/js'
-import { FlatCompat } from '@eslint/eslintrc'
-
-const __filename = fileURLToPath(import.meta.url)
-const __dirname = path.dirname(__filename)
-const compat = new FlatCompat({
-  baseDirectory: __dirname,
-  recommendedConfig: js.configs.recommended,
```
markus-codewrangler, line 15: AFAICS this recommended [setting](https://typescript-eslint.io/getting-started#step-2-configuration) was dropped

eslint.config.mjs
```diff
@@ -1,50 +1,33 @@
...
   {
-    plugins: {
-      '@typescript-eslint': fixupPluginRules(typescriptEslint),
-      'jsx-a11y': fixupPluginRules(jsxA11Y),
-      'react-hooks': reactHooks,
-    },
-
     languageOptions: {
-      parser: tsParser,
+      parser: tsEslint.parser,
```
markus-codewrangler: already included in `tsEslint.configs.recommended` AFAICS

eslint.config.mjs
```diff
@@ -1,50 +1,33 @@
...
       parserOptions: {
         projectService: true,
         tsconfigRootDir: import.meta.dirname,
```
markus-codewrangler: FYI is this redundant?

package.json
markus-codewrangler, line 1: FYI it would have made the intention clearer, if the dropping of lines and adding packages in various files (for various reasons, I suspect) had been split into separate commits with explanatory messages. in the current shape the reviewer can only guess the motivation 😢

markus-codewrangler (changes requested):
looks good in general 👍
but let's check some details 🧐

package.json
dcsaszar: I agree. Unfortunately these dependencies can only be updated in one go, since they interdepend on each other in different ways. Not sure if explaining the inter-dependencies would help in this case.

I was happy the ecosystem gave me a chance to update at all. 🙈

eslint.config.mjs
```diff
@@ -1,50 +1,33 @@
...
-
-const __filename = fileURLToPath(import.meta.url)
-const __dirname = path.dirname(__filename)
-const compat = new FlatCompat({
-  baseDirectory: __dirname,
-  recommendedConfig: js.configs.recommended,
-  allConfig: js.configs.all,
-})
 
 export default [
```
markus-codewrangler: FYI this is the new recommended syntax AFAICS (I overlooked this in the first round):

```suggestion
export default defineConfig({
  extends: ...
```

package.json
markus-codewrangler: FTR I don't mind to have multiple explanatory commits which only work in sum – or add more details to the commit message of the single commit.

markus-codewrangler (approved):
👍

markus-codewrangler (approved):

PR #1144, opened 2026-07-17
Use etracker instance ID instead of secure code

dcsaszar:
See internal feature announcement: https://infopark.slack.com/archives/C04TN8LJA/p1784281288382679

For now, I made sure we only track the same list of domains as before.
I left relaxing this and rolling it out to all instances for later, b/c implications may need discussion.

The backend already migrated (linked) the two tracking accounts:
* `LVKmem` (Sales demo) -> `8e775f86c0194b4ca7b11bef3f458ee3`
* `L9bLhx` (Others, including tynacoon.com as the top-tracked no-multi-instance domain) -> `13b78a0a81072f996f5010bb59b48957` (now tynacoon.com only)

FYI the backend will happily ignore instance IDs without a linked tracking account.

CC @fmeies-jr

### DIFF
--- a/src/privateJrPlatform/jrPlatformConfigureEtracker.ts
+++ b/src/privateJrPlatform/jrPlatformConfigureEtracker.ts
@@ -11,10 +11,8 @@ export function jrPlatformConfigureEtracker() {
   if (typeof window === 'undefined') return
   if (isEditorLoggedIn()) return
 
-  const secureCode =
-    getJrPlatformInstanceId() === '8e775f86c0194b4ca7b11bef3f458ee3'
-      ? 'LVKmem'
-      : 'L9bLhx'
+  const jrPlatformInstanceId = getJrPlatformInstanceId()
+  if (!jrPlatformInstanceId) return
 
   const hostnameInstance = instanceFromHostname()
   if (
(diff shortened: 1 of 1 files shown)


src/privateJrPlatform/jrPlatformConfigureEtracker.ts
```diff
@@ -11,10 +11,8 @@ export function jrPlatformConfigureEtracker() {
   if (typeof window === 'undefined') return
   if (isEditorLoggedIn()) return
 
-  const secureCode =
-    getJrPlatformInstanceId() === '8e775f86c0194b4ca7b11bef3f458ee3'
-      ? 'LVKmem'
-      : 'L9bLhx'
```
markus-codewrangler, line 17: previously, all other instances fell back to `L9bLhx` AFAICS

> L9bLhx (Others, including tynacoon.com as the top-tracked no-multi-instance domain) -> 13b78a0a81072f996f5010bb59b48957 (now tynacoon.com only) 

FTR the description sounds like this is an intentional change – and I assume that you have checked if other instances are affected (and potentially need to be configured with their own etracker instance).

markus-codewrangler (approved):
LGTM

src/privateJrPlatform/jrPlatformConfigureEtracker.ts
dcsaszar: Yes, Frank and I have reviewed the data: this account is only managed by/visible to @apepper, so no other user will lose any tracking data visible to them. We decided to migrate the tracking account to tynacoon.com since it
* has the most tracking data already stored
* has the most internal users
* no other production (or production-like) single-instance domain is tracked in this tracking account

PR #1156, opened 2026-08-03
Move fixed-size img rules to Tailwind

dcsaszar:


src/assets/stylesheets/index.css
```diff
@@ -249,6 +249,27 @@
   }
 }
 
+/* Fixed-size images
+ ================================================== */
+
+@layer components {
+  .img-flag {
+    @apply -mt-[3px] mr-1.5 inline-block max-h-[15px] w-[17px] min-w-[17px];
```
markus-codewrangler, line 257: deviations intended?

- mr: 6px vs. 1.5
- min-w: only on right

src/assets/stylesheets/index.css
```diff
@@ -249,6 +249,27 @@
...
+  }
+
+  .profile-img {
+    @apply -m-[5px] inline-block size-[30px] rounded-[var(--radius-portal)] object-cover;
```
markus-codewrangler: deviations intended?

- left: max/min-width vs. right: size
- mr: 0 vs. -5px

markus-codewrangler (approved):
LGTM – assuming deviations are intended 😁

src/assets/stylesheets/index.css
dcsaszar: Yes, 
1. Tailwind prefers the rem measures and only needs px when there is no built-in value.
2. I added `min-w-` as a safeguard. It will have no effect in the final design but in in-between states it helps keeping the img visible in the layout.

markus-codewrangler: > It will have no effect in the final design

- [ ] FYI add `TODO` comment so this can be found easily for clean up later?

src/assets/stylesheets/index.css
dcsaszar: * [x] mr

Will double check if we still need min/max after the conversion of the nav is done otherwise.

src/assets/stylesheets/index.css
dcsaszar: As designers should also be able to tweak the CSS, I'd just keep it like this to prevent unrelated changes from breaking the flag images.

markus-codewrangler (approved):

PR #1160, opened 2026-08-14
Enable etracker for all instances

dcsaszar:
[Concept GDoc](https://docs.google.com/document/d/1n-CTWSNBskI0uP6-qsnr1kmx2Vr2jySnXu9dWefEEPY)

Removes the instance/hostname allowlist and the JR-Platform-only gate, so etracker now loads for all instances instead of just the pilot one.

FYI the etracker backend mapping for tynacoon.com (`L9bLhx` -> `13b78a0a81072f996f5010bb59b48957`) was set up by Frank on 2026-07-17.

### DIFF
--- a/src/config/configureEtracker.ts
+++ b/src/config/configureEtracker.ts
@@ -6,7 +6,7 @@ declare global {
   }
 }
 
-export function jrPlatformConfigureEtracker() {
+export function configureEtracker() {
   if (typeof window === 'undefined') return
   if (isEditorLoggedIn()) return
 
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -1,5 +1,5 @@
 import { configureErrorReporting } from './configureErrorReporting'
-import { jrPlatformConfigureEtracker } from '../privateJrPlatform/jrPlatformConfigureEtracker'
+import { configureEtracker } from './configureEtracker'
 import { configurePisaSalesQuestionnaireWidgets } from './configurePisaSalesQuestionnaireWidgets'
 import { configureHistory } from './history'
 import { configureObjClassForContentType } from './objClassForContentType'
@@ -16,8 +16,5 @@ export function configure() {
   configureErrorReporting()
   configureWindowScrivito()
   configurePisaSalesQuestionnaireWidgets()
-
-  if (import.meta.env.PRIVATE_JR_PLATFORM) {
-    jrPlatformConfigureEtracker()
-  }
+  configureEtracker()
 }
--- a/src/privateJrPlatform/jrPlatformConfigureEtracker.ts
+++ b/src/privateJrPlatform/jrPlatformConfigureEtracker.ts
@@ -1,5 +1,4 @@
-import { isEditorLoggedIn } from 'scrivito'
-import { getJrPlatformInstanceId, instanceFromHostname } from './multiTenancy'
+import { getInstanceId, isEditorLoggedIn } from 'scrivito'
 
 declare global {
   interface Window {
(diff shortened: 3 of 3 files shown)


src/privateJrPlatform/jrPlatformConfigureEtracker.ts
markus-codewrangler, line 1: > c7829f0: Enable etracker **for all JustRelate platform instances**

it's unclear to the reader how a commit which _stops_ calling `getJrPlatformInstanceId` is supposed to _enable it for all_ platform instances. To me it looks like it's enabled for _any_ instance now (not only platform instances) 🤔

src/Components/Etracker.tsx
```diff
@@ -0,0 +1,33 @@
...
+    script.setAttribute('data-block-cookies', 'true')
+    script.setAttribute('data-instance-id', getInstanceId())
+    script.setAttribute('data-page-changed-detection', 'url')
+    script.src = 'https://code.etracker.com/code/e.js'
+    script.async = true
+
+    document.head.appendChild(script)
+  }, [])
+
+  return null
```
markus-codewrangler: isn't a component which always returns `null` an indication that (e.g.) a custom hook would be a better choice?

src/config/configureEtracker.ts
```diff
@@ -1,28 +0,0 @@
-import { getInstanceId, isEditorLoggedIn } from 'scrivito'
-
-declare global {
-  interface Window {
-    _etr?: { protocol: string }
-  }
-}
-
-export function jrPlatformConfigureEtracker() {
-  if (typeof window === 'undefined') return
```
markus-codewrangler, line 11: so the window guard is no longer needed?

markus-codewrangler (changes requested):
Mostly good 👍 – but some open questions 🧐

src/privateJrPlatform/jrPlatformConfigureEtracker.ts
dcsaszar: The first commit only removes conditions inside `jrPlatformConfigureEtracker()`, but keeps the script in the `src/privateJrPlatform` namespace and code path.

The `if (import.meta.env.PRIVATE_JR_PLATFORM)` is removed in the 2nd commit.

src/config/configureEtracker.ts
dcsaszar: Yes, lifecycle hooks are not called during prerendering.

src/config/index.ts
```diff
@@ -16,8 +15,4 @@ export function configure() {
   configureErrorReporting()
   configureWindowScrivito()
   configurePisaSalesQuestionnaireWidgets()
-
-  if (import.meta.env.PRIVATE_JR_PLATFORM) {
-    jrPlatformConfigureEtracker()
```
markus-codewrangler, line 21: FYI it's unexpected that the "configureEtracker" call was moved to `App.tsx`:

- [x] wouldn't it be more consistent to keep all configure function calls in this module?
- [ ] alternatively: clarify how etracker is different, so that it should be called in a different module

markus-codewrangler (approved):

src/config/index.ts
dcsaszar: Fair call. The component was inspired by https://github.com/JustRelate/scrivito-example-app-js/blob/master/src/Components/Tracking.js, but that one depends on app consent state, while etracker integration manages consent itself.

markus-codewrangler: > but that one depends on app consent state, while etracker integration manages consent itself.

so we should keep the code where it was before this PR IMHO

dcsaszar: FYI force pushed to keep the history readable

markus-codewrangler (approved):
🚀

PR #1163, opened 2026-08-19
Send user email address to etracker

dcsaszar:
[Concept Gdoc](https://docs.google.com/document/d/1n-CTWSNBskI0uP6-qsnr1kmx2Vr2jySnXu9dWefEEPY/edit?tab=t.0#heading=h.a5ha7nj8rh3w) (see comments under "User Information")

A working integration can be checked by looking at `window.dataLayer`. There will be an item with the hashed email. The final etracker backend implementation is still WIP.

CC @fmeies-jr

### DIFF
--- a/src/config/configureEtracker.ts
+++ b/src/config/configureEtracker.ts
@@ -1,8 +1,10 @@
-import { getInstanceId, isEditorLoggedIn } from 'scrivito'
+import { currentUser, getInstanceId, isEditorLoggedIn, load } from 'scrivito'
 
 declare global {
   interface Window {
     _etr?: { protocol: string }
+    _etracker?: { setUserEmailAddress: (email: string) => void }
+    _etrackerOnReady?: Array<() => void>
   }
 }
 
@@ -25,4 +27,14 @@ export function configureEtracker() {
   script.async = true
 
   document.head.appendChild(script)
+
+  registerUserEmail()
+}
+
+function registerUserEmail() {
+  window._etrackerOnReady = window._etrackerOnReady ?? []
+  window._etrackerOnReady.push(async () => {
+    const email = await load(() => currentUser()?.email())
+    if (email) window._etracker?.setUserEmailAddress(email)
+  })
 }


src/config/configureEtracker.ts
```diff
@@ -1,8 +1,10 @@
-import { getInstanceId, isEditorLoggedIn } from 'scrivito'
+import { currentUser, getInstanceId, isEditorLoggedIn, load } from 'scrivito'
 
 declare global {
   interface Window {
     _etr?: { protocol: string }
+    _etracker?: { setUserEmailAddress: (email: string) => void }
+    _etrackerOnReady?: Array<() => void>
```
markus-codewrangler: FYI

- [ ] looking at your code below, I would expect this shape:

```suggestion
    _etrackerOnReady?: Array<() => Promise<void>>
```

- [ ] did you/someone clarify with the etracker team that async callbacks are supported on their end?
- [ ] it would be great if the etracker APIs used here could be transferred from the discussion thread into the main concept doc

markus-codewrangler (approved):
Nice 👍 

See comments for potential issues/clarifications.

src/config/configureEtracker.ts
dcsaszar: > did you/someone clarify with the etracker team that async callbacks are supported on their end?

An async callback is just a function returning a Promise instead of void. Since etracker supports "a function" this includes "a function returning something". See https://www.typescriptlang.org/docs/handbook/2/functions.html#return-type-void

Restricting the type to the subset of void async would be technically incorrect/questionable.

If you think this is a readability issue we could load the user email unconditionally before setting up the onReady callback:

```ts
async function registerUserEmail() {
  const email = await load(() => currentUser()?.email())
  if (!email) return

  window._etrackerOnReady = window._etrackerOnReady ?? []
  window._etrackerOnReady.push(() => window._etracker?.setUserEmailAddress(email))
}
```

But this has other readability issues:
* It apparently does unnecessary work in the case etracker is not activated (e.g. by blocking). This in turn is actually not true given other modules already load the user, but this is not obvious.
* The naming may raise doubts about the callback still being called if etracker is ready before the user email arrives, which has to be looked up in the docs

WDYT?

markus-codewrangler: I don't think rewriting the function as you proposed above makes it better 🙅‍♂️

> An async callback is just a function returning a Promise instead of void.

it's first of all a function which (can) return _later_.

depending on the implementations of `window._etrackerOnReady`, `window._etracker` and `window._etracker.setUserEmailAddress` this _could_ lead to race conditions or user emails never reaching the etracker backend IMHO.

this can/should not be improved on the callback side, but clarified with the etracker team – as I proposed above 🔄

dcsaszar: Maybe I was not clear with my comments. They are theoretical questions a **reader** could ask. No matter how the implementation, it will always work and have the identical effect. The discussion is not about which implementation works (they all do, so there is nothing that needs improvement), but what is **best readable** (let's say for a junior dev).

If we're not optimizing for readability, I completely misunderstood your comment. 🙈

markus-codewrangler: I think the current shape of your changes is good 👍 

but it would help to move the conversation forward if you answered my question (asking the 3rd time now):

>  did you/someone clarify with the etracker team that async callbacks are supported on their end?

fmeies-jr: @mbeboek-jr Can you please have a look?

mbeboek-jr: passing an async function into `_etrackerOnReady`is supported yes.
etracker internally just calls the functions without checking results, so an async one would just be fire-and-forget.

